PR #62, opened 2023-10-27
feat: Extism Support

bhelx:
Still have some work to do on this but this is working on my machine. We redesigned Extism so that most of the code can be compiled to Wasm. We call this the "kernel". This has allowed us to port Extism to new runtimes quickly. I was able to do this with about 3 hours work today. I'm pretty blown away that this works.

Screenshot 2023-10-27 at 3 38 29 PM


bhelx:
Closing and moving development to https://github.com/extism/chicory-sdk

PR #65, opened 2023-10-29
Add PoC for proxy API

thomasdarimont:
This example enables us to do this:
```java
public class ChicoryApiTest {

    @WasmModule("wasm/count_vowels.rs.wasm")
    public interface HelloWorld {

        @WasmFunction("count_vowels")
        int countVowels(String input);
    }

    @Test
    public void shouldCreateProxy() {

        HelloWorld hw = Chicory.proxy(HelloWorld.class);

        assertNotNull(hw);
        assertEquals(3, hw.countVowels("AAAx"));

        assertEquals(hw, hw);
        assertEquals(hw.hashCode(), hw.hashCode());
        assertNotNull(hw.toString());
    }
}
```


thomasdarimont:
Thanks for your thoughts! I reconsider the approach and try to come up with something more low-level.

andreaTP:
We are converging with an annotation processor and code generation :-)

It's coming up pretty well, I'll ask for your review as soon as it is ready @thomasdarimont !

PR #92, opened 2023-11-08
Disable test-gen when using -DskipTests

andreaTP:
This is a fix to make the `test-gen-plugin` compatible with Jitpack (and the automatic publication of artifacts).

Basically when passing `-DskipTests` the entire test infrastructure generation is disabled, this way Jitpack should be able to automatically publish preview releases.


andreaTP:
Superseeded by #93

PR #95, opened 2023-11-10
fix: Fixes mapHostFunctions algorithm

bhelx:
Adopts fix that's been hanging out in #62

We weren't noticing this was broken because all of our tests only use one import :)

### DIFF
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/Module.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/Module.java
@@ -202,9 +202,17 @@ public class Module {
         }
 
         var globalImportsOffset = 0;
+        var functionImportsOffset = 0;
         for (int i = 0; i < imports.length; i++) {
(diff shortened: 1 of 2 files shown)


runtime/src/main/java/com/dylibso/chicory/runtime/Module.java
```diff
@@ -200,9 +202,17 @@ public Instance instantiate(HostFunction[] hostFunctions) {
         }
 
         var globalImportsOffset = 0;
+        var functionImportsOffset = 0;
         for (int i = 0; i < imports.length; i++) {
-            if (imports[i].getDesc().getType() == ImportDescType.GlobalIdx) {
-                globalImportsOffset++;
+            switch (imports[i].getDesc().getType()) {
```
andreaTP, line 207: let's see how this goes, but I'll probably move this code to the `Instance` constructor next time I touch it.

andreaTP (approved):

PR #96, opened 2023-11-10
table parsing

andreaTP:
A little more logic for tables parsing.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -40,7 +40,7 @@ jobs:
         if: always() && matrix.os == 'ubuntu-latest' && matrix.version == '17'
         with:
           files: './**/surefire-reports/*.xml'
-  `   - name: Publish badge conditional
+      - name: Publish badge conditional
         id: publish
         run: echo "badge={{ (matrix.os == 'ubuntu-latest' && matrix.version == '17') && (github.event_name == 'workflow_run' && github.event.workflow_run.head_branch == 'main' || github.event_name != 'workflow_run' && github.ref == 'refs/heads/main') }}" >> $GITHUB_OUTPUT
       - name: Set badge color
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -42,7 +42,11 @@ jobs:
           files: './**/surefire-reports/*.xml'
       - name: Publish badge conditional
         id: publish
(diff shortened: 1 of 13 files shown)


runtime/pom.xml
```diff
@@ -54,15 +54,15 @@
             f64_cmp.wast,
             float_misc.wast,
             fac.wast,
-            data.wast,
```
andreaTP, line 57: this test was erroneously passing only because it was never instantiated.
Now the logic is fixed in `test-gen`.

wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java
```diff
@@ -308,13 +309,39 @@ private static ImportSection parseImportSection(
...
+                        imports[i] = new Import(moduleName, fieldName, funcDesc);
+                        break;
+                    }
                 case TableIdx:
-                    throw new ChicoryException("Don't support table type globals yet");
+                    {
+                        var min = (int) readVarUInt32(buffer);
+                        var next = (int) readVarUInt32(buffer);
+                        var max = 0;
+                        if (buffer.hasRemaining()) {
```
andreaTP, line 322: I'm not 100% this is the right way to handle the "optional" max value, I welcome any idea.

wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java
```diff
@@ -308,13 +309,39 @@ private static ImportSection parseImportSection(
...
+                            max = next;
+                            next = (int) readVarUInt32(buffer);
+                        }
+                        var limits = new Limits(min, max);
+                        var tableDescType = (next == 0) ? ValueType.FuncRef : ValueType.ExternRef;
```
andreaTP, line 327: let's see if this is correct or not ...

wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java
andreaTP: found the right encoding

wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java
andreaTP: found the definition

.github/workflows/ci.yaml
```diff
@@ -40,9 +40,16 @@ jobs:
         if: always() && matrix.os == 'ubuntu-latest' && matrix.version == '17'
         with:
           files: './**/surefire-reports/*.xml'
+      - name: Publish badge conditional
```
andreaTP, line 43: we can verify if this works correctly after merging the PR.

bhelx (approved):

PR #99, opened 2023-11-19
Make HostFunction's ctor accessible

AppleFlavored:
Looking at the README, it seems like `HostFunction`'s constructor should be public.


bhelx (approved):
Yes, thanks for taking the time to fix it!

I'm going to be pushing it out to maven central next week and I'm sure we'll find some more things like this.

PR #101, opened 2023-11-24
Release 0.0.2

bhelx:


andreaTP:
> I can skip publishing these submodules?

The maven plugins do not need to be published as they are self-contained in this project, but, as far as I understand it, it's failing just because it's the first to be published.

I have reviewed the changes and I do not get what's going wrong here, maybe @aalmiray has an idea 🙏

aalmiray:
These errors are printed out because https://github.com/dylibso/chicory/blob/main/wasm-support-plugin/pom.xml does not define `<description>` nor `<url>`.

Any modules that are not required for deployment may be configured with `<maven.deploy.skip>true</maven.deploy.skip>` in its `<properties>` section. Or you may use a `<profile>` to skip including the module, which is what ModiTect does to skip including the `integrationtests` module when the release profile is activated

https://github.com/moditect/moditect/blob/f32d851377b1f40627b8426a976f0de496068536/pom.xml#L42-L61

aalmiray:
FWIW every other pom file is missing `<description>` and `<url>`. Only the root pom defines an `<url>` element.

It's recommended that every pom defines the minimum set of XML elements required for publication to Maven Central (this is what JReleaser is checking via [Pomchecker](https://kordamp.org/pomchecker/pomchecker-maven-plugin/check-maven-central-mojo.html).

https://central.sonatype.org/publish/requirements/#correct-coordinates

andreaTP:
Thanks a lot @aalmiray for the quick response, much appreciated 👍 🙏

aalmiray:
I'd recommend running a release in dry run mode on your local environment first to iron out any details. This page also mentions how you may set the alternate deploy property as part of the build using a profile https://jreleaser.org/guide/latest/examples/maven/maven-central.html#_maven

aalmiray:
FWIW `jreleaser:prepare` is supposed to be used for creating files required by package managers. It is no way related to `release:prepare` as provided by the `maven-release-plugin`. 

More information can be found [in the guide](https://jreleaser.org/guide/latest/concepts/workflow.html#_prepare).

PR #109, opened 2023-12-05
ci: automate release using jreleaser action

bhelx:
https://github.com/marketplace/actions/jreleaser

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,59 @@
+name: release
+
+on:
+  workflow_dispatch:
+
+jobs:
+  release:
+    name: Release
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      # Configure build steps as you'd normally do
+
+      - name: Setup Java
+        uses: actions/setup-java@v3
+        with:
+          java-version: 11
+          distribution: 'zulu'
+          server-id: central
+          server-username: MAVEN_USERNAME
+          server-password: MAVEN_CENTRAL_TOKEN
+          gpg-private-key: ${{ secrets.GPG_PRIVATE_KEY }}
+          gpg-passphrase: MAVEN_GPG_PASSPHRASE
+
+      # Post JARs to Maven Central
+
+      - name: Release to Maven Central
+        env:
+          MAVEN_USERNAME: ${{ secrets.SONATYPE_USERNAME }}
+          MAVEN_CENTRAL_TOKEN: ${{ secrets.SONATYPE_PASSWORD }}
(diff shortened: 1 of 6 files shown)


.github/workflows/release.yaml
```diff
@@ -0,0 +1,59 @@
...
+        with:
+          fetch-depth: 0
+
+      # Configure build steps as you'd normally do
+
+      - name: Setup Java
+        uses: actions/setup-java@v3
+        with:
+          java-version: 11
+          distribution: 'zulu'
```
andreaTP, line 23: I would go for `temurin`

.github/workflows/release.yaml
```diff
@@ -0,0 +1,59 @@
...
+      # Post JARs to Maven Central
+
+      - name: Release to Maven Central
+        env:
+          MAVEN_USERNAME: ${{ secrets.SONATYPE_USERNAME }}
+          MAVEN_CENTRAL_TOKEN: ${{ secrets.SONATYPE_PASSWORD }}
+          MAVEN_GPG_PASSPHRASE: ${{ secrets.GPG_PASSPHRASE }}
+        run: |
+          export GPG_TTY=$(tty)
+          git config user.name "${{ github.event.head_commit.committer.name }}"
```
andreaTP, line 39: wdyt about using fixed "bot-credentials" ? Instead of recycling the last committer data?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,59 @@
+name: release
+
+on:
+  workflow_dispatch:
```
andreaTP, line 4: When releasing with `workflow_dispatch` I normally make it customizable:
- the version number to be released
- the branch to release from

e.g.: https://github.com/kiota-community/kiota-java-extra/blob/8b9f209e22ecf3490d0d4e18bad47bf4695bb52e/.github/workflows/release.yml#L6-L12

andreaTP (commented):
Releasing Java have one additional issue, which is bumping the version in the `pom.xml` files, this is one way of doing it:
https://github.com/kiota-community/kiota-java-extra/blob/8b9f209e22ecf3490d0d4e18bad47bf4695bb52e/.github/workflows/release.yml#L45-L52

but it would be great if we can also automate this detail ...

andreaTP:
@bhelx do you need some help to finish this one?
We have made some improvements and it will be ideal to easily release.

andreaTP:
@bhelx I finished applying the fix for my review, a couple of notes:

- you need to set an additional secret: `GH_TOKEN` that should be able to push to the `main` branch
- the release process is not atomic, e.g. we should avoid merging PRs during a release, otherwise we might need manual actions to recover

@lburgazzoli is waiting for a new release to finalize his Camel integration and distributing it ❤️

bhelx:
I set a personal access token at GH_TOKEN. Will expire in 90s days but i can revisit how best to do this. I was unsure which permissions to give it, it only seemed to care about private repo permissions.

PR #112, opened 2023-12-06
imports.wast tests infrastructure and global imports support

andreaTP:
Looking for feedback on this approach to testing imported functions/things.

I was thinking about generating the `HostFunctions`, but I'm unsure how this will play out ... probably defining them explicitly is going to be more flexible, thoughts?


bhelx:
> I was thinking about generating the HostFunctions, but I'm unsure how this will play out ... probably defining them explicitly is going to be more flexible, thoughts?

I think if we can get away with a static set of hand written host function then we should do that. I haven't looked too closely, but i think the imports with different behaviors different names so this should work.

andreaTP:
Thanks for the feedback @bhelx I'll continue in this direction 👍

andreaTP:
A few bits are a bit hacked around, feel free to ask for changes or improvements.

This PR features:
- Test Gen Maven Plugin support for the `imports.wast` spec
- Implement Global Imports

I think this is already big enough to review, we can follow up with Memory and Table imports in a follow-up PR.

andreaTP:
@bhelx heads up, this is ready for review.

bhelx (approved):
This is a pretty sprawling change so I think it's worth merging now and following up on small stuff and remaining specs. Thanks!

PR #113, opened 2023-12-13
Re-enable test generation for assert_invalid & assert_malformed

danielperano:
Re-enables test generation for `assert_invalid` & `assert_malformed` commands as per [issue 37](https://github.com/dylibso/chicory/issues/37).

Note that without this the main branch is passing 15780 tests (or 15781, don't recall exactly), and with this it's running 16971 & failing 1191 of them (the expected exceptions are not being thrown).

PR #115, opened 2023-12-13
Add contributing guidelines

andreaTP:
As noticed in #113 we are falling short on documentation about the current processes.
This should help a little, or, at least, will be a place to collect the information instead of repeating it on different channels.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,80 @@
+# Contributing guide
+
+**Want to contribute? Great!** 
+We try to make it easy, and all contributions, even the smaller ones, are more than welcome.
+This includes bug reports, fixes, documentation, examples... 
+But first, read this page (including the small print at the end).
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,80 @@
...
+  + [Coding Guidelines](#coding-guidelines)
+  + [Continuous Integration](#continuous-integration)
+  + [Tests and documentation are not optional](#tests-and-documentation-are-not-optional)
+  + [Current status](#current-status)
+* [The small print](#the-small-print)
+
+
+## Legal
+
+All original contributions to Apicurio projects are licensed under the
```
bhelx, line 21: I think we can change this to `Chicory` in this document

andreaTP: argh, now you know where I started from 😄

andreaTP: fixed

PR #116, opened 2023-12-13
Fixed a couple of issues with the I64_TRUNC_FXX_U opcodes.

danielperano:
Re-enabled a couple of the skipped conversion tests that were failing due to bugs in the unsigned float to i64 truncation implementations, and fixed those bugs.

### DIFF
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/Machine.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/Machine.java
@@ -1473,10 +1473,30 @@ public class Machine {
                                 throw new WASMRuntimeException("invalid conversion to integer");
                             }
 
-                            var tosL = (long) tos;
-
-                            if (tosL < 0 || (tosL == Long.MAX_VALUE)) {
+                            if (tos >= 2 * (float) Long.MAX_VALUE) {
                                 throw new WASMRuntimeException("integer overflow");
+                            }
+
+                            long tosL;
+                            if (tos < Long.MAX_VALUE) {
+                                tosL = (long) tos;
+                                if (tosL < 0) {
+                                    throw new WASMRuntimeException("integer overflow");
+                                }
+                            } else {
+                                // This works for getting the unsigned value because binary addition
+                                // yields the correct interpretation in both unsigned &
+                                // 2's-complement
+                                // no matter which the operands are considered to be.
+                                tosL = Long.MAX_VALUE + (long) (tos - (float) Long.MAX_VALUE) + 1;
+                                if (tosL >= 0) {
+                                    // Java's comparison operators assume signed integers. In the
+                                    // case
+                                    // that we're in the range of unsigned values where the sign bit
+                                    // is set, Java considers these values to be negative so we have
+                                    // to check for >= 0 to detect overflow.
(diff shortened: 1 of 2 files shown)


andreaTP (approved):
Nice one!
Looks like you are catching up with the codebase pretty quickly, please, let us know if something is not working well for you in the development experience (e.g. you should be able to run individual tests from the IDE etc. etc.).

Adding @thomasdarimont as a reviewer as I think he has spent some cycles on those conversions.

danielperano:
> Looks like you are catching up with the codebase pretty quickly, please, let us know if something is not working well for you in the development experience (e.g. you should be able to run individual tests from the IDE etc. etc.).

Thanks, @andreaTP - so far so good. The trickiest bit was getting used to Maven (briefly dabbled with Ant many years ago, then switched to Gradle and never looked back) & playing around with the test gen last night. Aside from a couple idiosyncrasies around test generation (need the right sequence of clean/verify sometimes), it's been pretty easy going - I've found it to be a clean and well-organized codebase.

runtime/src/main/java/com/dylibso/chicory/runtime/Machine.java
```diff
@@ -1489,12 +1509,26 @@ void eval(List<Instruction> code) throws ChicoryException {
                             if (Double.isNaN(tos)) {
                                 throw new WASMRuntimeException("invalid conversion to integer");
                             }
-                            var tosL = (long) tos;
-                            if (tos == (double) Long.MAX_VALUE) {
```
danielperano, line 1493: @thomasdarimont I'm interested to get your feedback on removing this check - this is the one bit I didn't understand, though I don't think we need it with these considerations of wrap-around added in.

My expectation is that there are probably faster & cleaner ways to handle this truncation with some bit-twiddling, but for the moment this gets the tests passing.

andreaTP:
Apparently Thomas is busy those days, I think that this is an improvement going in the right direction.
Also, the diff is pretty self-contained, all of the tests are already passing plus the 2 re-enabled here.

I'll take ownership to make sure we fix it on top of the latest `main` whenever we hear back from Thomas.

thomasdarimont:
Sorry guys for the lack of response. I'm currently dealing with some non-work related stuff that needs my attention. 

Please go ahead and continue at your own pace. I hope I can join you guys soon again.

runtime/src/main/java/com/dylibso/chicory/runtime/Machine.java
thomasdarimont: I had to fiddle quite a bit with the bitwise representations to get the tests pass. I'm fine with any more elegant or straight forward solution:)

bhelx:
> Sorry guys for the lack of response. I'm currently dealing with some non-work related stuff that needs my attention.

It's no problem! Let's not hold up any PRs unless it's dire. As soon as you get one approval you can merge. We can always circle back on our own timeline since we're all async, just leave a comment that the person can come check when they're ready. Because we don't have any prod users or anything, the risk of stagnating is lower than the risk of pushing potentially problematic code.

bhelx:
Everyone of you who have commit access is trustworthy enough to merge any changes on your own :)

danielperano:
Sounds good!

PR #118, opened 2023-12-14
refactor: Refactor Wasm Value to single primitive

bhelx:
An attempt to simplify the Value class. It's allocated and called a lot and was overly complex (totally my fault). I think storing everything in a single long primitive should make things faster and lower resource. Though there is probably some other low hanging fruit optimizations in here. I haven't tested it much but seems to shave about 4 seconds off the test suite so far.


andreaTP (approved):
LGTM, thanks for taking care of this!

PR #119, opened 2023-12-14
Fixed global import init

danielperano:
Re-enabled global tests that had been disabled due to lack of import support in tests. Found & fixed module instantiation bug that the tests discovered.

### DIFF
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/Module.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/Module.java
@@ -83,10 +83,17 @@ public class Module {
                     globals[i] = Value.f64(instr.getOperands()[0]);
                     break;
                 case GLOBAL_GET:
-                    // TODO this assumes that these are already initialized declared in order
-                    // should we make this more resilient? Should initialization happen later?
-                    globals[i] = globals[(int) instr.getOperands()[0]];
-                    break;
+                    {
+                        // TODO this assumes that these are already initialized declared in order
+                        // should we make this more resilient? Should initialization happen later?
+                        var globalImports = hostImports.getGlobals();
+                        var idx = (int) instr.getOperands()[0];
+                        globals[i] =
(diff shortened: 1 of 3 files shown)


runtime/src/main/java/com/dylibso/chicory/runtime/Module.java
```diff
@@ -83,10 +83,17 @@ public Instance instantiate(HostImports hostImports) {
...
-                    // should we make this more resilient? Should initialization happen later?
-                    globals[i] = globals[(int) instr.getOperands()[0]];
-                    break;
+                    {
+                        // TODO this assumes that these are already initialized declared in order
+                        // should we make this more resilient? Should initialization happen later?
+                        var globalImports = hostImports.getGlobals();
+                        var idx = (int) instr.getOperands()[0];
+                        globals[i] =
+                                idx < globalImports.length
```
andreaTP, line 92: I like this implementation, so far I have been following a slightly different pattern as it was "more used" in the rest of the codebase: https://github.com/dylibso/chicory/blob/361c145ebf83b283003ed7053f3cdb09607f9661/runtime/src/main/java/com/dylibso/chicory/runtime/Machine.java#L303

just a nitpick.

andreaTP (approved):
Awesome! Thanks a lot @danielperano looks like you are catching up quickly on the codebase 🥇

runtime/src/main/java/com/dylibso/chicory/runtime/Module.java
danielperano: In this case though, instance is not yet available since it's constructed at the end of this method.

danielperano: Did you want to discuss this further, or are you happy with merging?

andreaTP: good to go 👍  thanks!

PR #120, opened 2023-12-14
Fix the tests badge link

andreaTP:
The number was not getting updated, this should work.


andreaTP:
I knocked on the wood during the meeting @danielperano 🤣

PR #121, opened 2023-12-18
Minor Readme updates

andreaTP:


bhelx:
Strange, this didn't work for me before but now it does. Thanks!

PR #122, opened 2023-12-18
Finish the imports spec

andreaTP:
Add support for Table imports, I'm exposing a simplified API for defining the table to the user.

A possible improvement in the future will be to enable referencing imported functions by name instead of using the underlying index but this is enough to pass a few more tests.


andreaTP:
@bhelx this is ready for review, this finalizes the work on imports.

runtime/src/main/java/com/dylibso/chicory/runtime/Module.java
```diff
@@ -268,12 +267,16 @@ private HostImports mapHostImports(Import[] imports, HostImports hostImports) {
             switch (imprt.getDesc().getType()) {
                 case FuncIdx:
                     hostFuncNum++;
+                    break;
```
bhelx, line 270: oops 😆

bhelx:
Looks good to me. I'm assuming since it's green it's good to merge!

runtime/src/main/java/com/dylibso/chicory/runtime/Module.java
andreaTP: That precise feeling 😅

bhelx: For me this is the downside of writing in so many languages. I forget all these tiny little differences. I wish Java did more to protect you here. I like the enhanced switch of course, but we can't use that yet right?. Maybe some tooling could help?

andreaTP: > For me this is the downside of writing in so many languages. 

same here, I think that I wrote this bug by myself even if I'm using Java daily.

> we can't use that yet right?

we would need to switch to Java 17, which can be a reasonable choice as even [Quarkus is dropping Java 11 support](https://quarkus.io/blog/java-17/), I have mixed feelings about ...

PR #125, opened 2023-12-20
[chore] Separate sources from compiled wasm files

### LINKED ISSUE #56: Where to put test wasm modules
We have a folder with all the test wasm modules: https://github.com/dylibso/chicory/tree/main/runtime/src/test/resources/wasm
This has the source code (.wat, .c, .rs) as well as the binaries (.wasm) checked into the source. 
We have a script to compile and manage them: https://github.com/dylibso/chicory/blob/main/runtime/scripts/compile-tests.sh

These are both repeated in the `wasm` submodule project because they were historically 2 projects. We should look at one single / simple way to manage these.

andreaTP:
Fixes #56 
This PR aims to make the process of compiling `wasm` sources more "streamlined" inside this project.

More specifically:
 - define a container image that can compile `rust`, `c` and `wat` files
 - separate sources from compiled targets under the `resources` folder
 - add a script to regenerate from scratch the compiled targets
 
NOTE: I have not found the source for `code.wasm`


scripts/compile-resources.sh
```diff
@@ -0,0 +1,40 @@
+#! /bin/bash
+set -euxo pipefail
+
+CONTAINER_IMAGE="docker.io/andreatp/chicory-compilation-support"
```
andreaTP, line 4: @bhelx if you wanna rebuild the image and publish it under the `dylibso` organization feel free to go ahead.

bhelx: Thanks I will take a look!

andreaTP:
>  I think for the time being it's fine to manage the dockerfile locally like this but we'll consider publishing it.

Quick note, the image is available, is simply just published under my personal account.
E.g. you don't need to rebuild the container image locally.

PR #129, opened 2023-12-22
Fixed nonarithmetic NaN propagation

danielperano:
Fixed nonarithmetic NaN propagation  through Value box & fneg operations.

All `float_exprs.wast` tests pass with this.


andreaTP (approved):
Nice cleanup, thanks!

Issue #127, opened 2023-12-20
Load data segments in the imported memory


andreaTP:
This initialization is not correct and we should instead correctly load the data segments:
https://github.com/dylibso/chicory/blob/80b21fcf6b875f57d9bb1fe9ac340192e458716c/runtime/src/test/java/com/dylibso/chicory/imports/SpecV1ImportsHostFuncs.java#L120

PR #130, opened 2023-12-24
Refactor memory to improve API & fix failing tests

danielperano:
Misc. cleanup to internal & external API for the Memory class, targeting fixing all failing memory init tests.

### DIFF
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/Machine.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/Machine.java
@@ -1714,10 +1714,8 @@ public class Machine {
                             this.stack.push(Value.i64((long) tos));
                             break;
                         }
-                    case MEMORY_INIT_FC:
                     case MEMORY_INIT:
                         {
-                            System.out.println("Stack size: " + stack.size());
                             var segmentId = (int) operands[0];
                             var memidx = (int) operands[1];
                             if (memidx != 0)
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/Machine.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/Machine.java
@@ -320,42 +320,42 @@ public class Machine {
                     case I32_LOAD:
                         {
                             var ptr = (int) (operands[1] + this.stack.pop().asInt());
-                            var val = instance.getMemory().getI32(ptr);
+                            var val = instance.getMemory().readI32(ptr);
                             this.stack.push(val);
                             break;
                         }
                     case I64_LOAD:
                         {
                             var ptr = (int) (operands[1] + this.stack.pop().asInt());
-                            var val = instance.getMemory().getI64(ptr);
+                            var val = instance.getMemory().readI64(ptr);
                             this.stack.push(val);
                             break;
                         }
                     case F32_LOAD:
                         {
                             var ptr = (int) (operands[1] + this.stack.pop().asInt());
-                            var val = instance.getMemory().getF32(ptr);
+                            var val = instance.getMemory().readF32(ptr);
                             this.stack.push(val);
                             break;
                         }
                     case F64_LOAD:
                         {
                             var ptr = (int) (operands[1] + this.stack.pop().asInt());
-                            var val = instance.getMemory().getF64(ptr);
+                            var val = instance.getMemory().readF64(ptr);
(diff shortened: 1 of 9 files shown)


andreaTP:
Thanks for picking this up @danielperano it was on my todo but happy to see you taking over!
Please, while looking at this, I think we should debug the Data segments initialization of the memory which I think is not completely correct ATM.

runtime/src/main/java/com/dylibso/chicory/runtime/Machine.java
```diff
@@ -320,140 +320,140 @@ void eval(List<Instruction> code) throws ChicoryException {
                     case I32_LOAD:
                         {
                             var ptr = (int) (operands[1] + this.stack.pop().asInt());
-                            var val = instance.getMemory().getI32(ptr);
+                            var val = instance.getMemory().readI32(ptr);
```
bhelx, line 323: nice, i think this is a little clearer

runtime/src/main/java/com/dylibso/chicory/runtime/Memory.java
```diff
@@ -120,228 +116,191 @@ public void reinstantiate() {
...
     }
 
-    public void put(int offset, byte[] data) {
+    public void write(int addr, byte[] data) {
+        write(addr, data, 0, data.length);
+    }
+
+    public void write(int addr, byte[] data, int offset, int size) {
         try {
-            // System.out.println("mem-write@" + offset + " " + data);
```
bhelx, line 183: Thank you for fixing this! Couldn't find the right API for this.

danielperano: Yeah, `ByteBuffer` is tricky with that API inconsistency - I seem to recall running into it before, and it's one of the reasons I don't particularly like using `ByteBuffer` outside of I/O (though I think its use here is as good as we can do outside of using Panama's native memory API, which isn't even finalized until JDK 22).

wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java
```diff
@@ -643,22 +643,24 @@ private static DataSection parseDataSection(
         var dataSegments = new DataSegment[(int) dataSegmentCount];
 
         for (var i = 0; i < dataSegmentCount; i++) {
-            var memidx = readVarUInt32(buffer);
-            // if memory idx == 0, then this is an active datasegment
-            // if memory idx == 1, then this is a passive datasegment
-            // TODO won't this break once we support multi-memory?
-            if (memidx == 0) {
+            var mode = readVarUInt32(buffer);
```
andreaTP, line 646: Nice one!
This covers all the cases now:
https://webassembly.github.io/spec/core/binary/modules.html#data-section

🎉  do you want to change the explicit throw of the exception to an assertion?

andreaTP (approved):
A minor nitpick but LGTM it's a good step forward!

We still miss to move the memory initialization after instantiation and correctly apply it to  imported memory, but it can be done in a separate PR 👍

wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java
danielperano: For something like this I'd prefer the throw - if someone tries to parse a newer version of a binary (or an incorrectly version labeled binary, or a malformed binary) then it's conceivable that parsing might make it this far.

danielperano:
> A minor nitpick but LGTM it's a good step forward!
> 
> We still miss to move the memory initialization after instantiation and correctly apply it to imported memory, but it can be done in a separate PR 👍

I think that's a pretty tiny change - what did you have in mind for that? Just move the initialization call from the module constructor to `instantiate()`?

wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java
andreaTP: just a personal preference, not fussy on this 🤷

andreaTP (approved):
LGTM, feel free to merge.

> I think that's a pretty tiny change - what did you have in mind for that? Just move the initialization call from the module constructor to instantiate()?

Yes and making sure that initialization applies also to imported memories to fix #127 
Let's merge this one and build on top 👍

PR #133, opened 2023-12-29
Include the unwind spec

andreaTP:
Getting to pass the last unwind tests has been hard.

With this, all of the tests are passing, but I'm not 100% convinced about the implementation, especially because copying around the entire Stack is gonna be expensive.

At this point, I have spent quite some time on the control flow, and I might be overlooking something.
Having fresh eyes would be great, I'm available to pair in case.

### DIFF
--- a/runtime/pom.xml
+++ b/runtime/pom.xml
@@ -85,7 +85,8 @@
             nop.wast,
             float_exprs.wast,
             global.wast,
-            imports.wast</orderedWastToProcess>
+            imports.wast,
+            unwind.wast</orderedWastToProcess>
           <excludedTests>
             <!--
             Assertion failures -->
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/MStack.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/MStack.java
@@ -15,13 +15,17 @@ public class MStack {
         this.stack = new Stack<>();
     }
 
-    private StackFrame unwindFrame;
+    private Stack<Value> unwindFrame;
 
-    public void setRestoreFrame(StackFrame frame) {
-        this.unwindFrame = frame;
+    public void setUnwindFrame(Stack<Value> stack) {
+        this.unwindFrame = stack;
     }
 
-    public StackFrame getRestoreFrame() {
+    public void resetUnwindFrame() {
+        this.unwindFrame = null;
+    }
+
+    public Stack<Value> getUnwindFrame() {
         return this.unwindFrame;
     }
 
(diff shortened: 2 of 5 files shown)


danielperano (commented):
I can think of a couple of different ways to save/restore the operand stack that should be faster.

1. Instead of doing push/pop in a loop, use data structures that give direct access to the underlying storage arrays and `System.arraycopy` the parameter/return values. I think this is still suboptimal though.
2. Replace the current stack implementation with one that is array based and allows creating new "substacks" that don't allow accessing values outside of the current function's operand stack. This would look something like having a class `OperandStack { valueArray, minLimit }` where `minLimit` denotes the boundary between the current call frame's operand stack and the previous call frame's stack. Entering a function then just implies creating a new `OperandStack` backed by the same storage with an appropriate minLimit. The backing storage can either be pre-allocated at some max operand stack depth or can be wrapped (by creating a `ResizableArray`) so that storage can be re-allocated without tracking all of the `OperandStack` instances that share the same storage. Exiting a function just implies silently dropping the call frame's `OperandStack` and adjusting the stack index of the operand stack in the return frame to include any return values. This results in a copy-free mechanism where each function invocation creates at most one object for the duration of the call.

andreaTP:
Thanks for the input @danielperano , your second proposal seems appealing, I'll give it a try to see how it shapes up 👍

danielperano:
If this works out well, at some point in the future I'm curious if we could improve things further by replacing the stack representation as a `Value[]` with a `long[] + Object[]` (the latter for references). Then we could store all numeric operands as unwrapped 64-bit words, and reference operands as an index from the operands stack into the reference stack. That would alleviate a lot of short-lived allocations (typically not terribly important for a desktop VM, but may be much more optimal on other VMs and may improvement arithmetic speed since it avoids boxing operations) and might be better for cache locality/density.

bhelx:
The recommended way i've seen to deal with this is to store the block indices on the stack itself and use those as markers. Is there a reason we can't do that? I think something like @danielperano second idea sounds similar.

Overall I'm okay with moving forward with what we have if the specs are green. It's definitely something we can come back and look at refactoring and simplifying. It'll be easier to do with passing tests.

danielperano:
If we want to be minimal now I'd at least move the stack manipulation operations into their own methods so we have known and well-encapsulated reference points when we refactor the stack representation later.

andreaTP:
All great comments, thanks a lot for chiming in!
I think I have a concrete plan or two to try out, I'll time box the attempts and will share the result shortly.

danielperano (approved):

danielperano:
Looks good!

bhelx (approved):
> Anyhow, this is passing the specs, it's not optimal but looks like a reasonable first approximation (at least it's not super computational heavy).

That's okay. It unlocks more specs and we can go back and refactor this!

Issue #134, opened 2023-12-29
Wast2JsonWrapper#resolveOrInstallWast2Json does not work with wast2json 1.0.34+


starksm64:
I'm building the chicory project on an m1 max osx system and the downloaded wast2json binary fails to load because the x86_64 arch binary fails to load dependent libs like openssl which have only been installed as arm64 arch binaries. Instead of trying to install a separate set of x86_64 libs using brew under Rosetta, I just rebuilt the wabt repo natively.

The Wast2JsonWrapper#resolveOrInstallWast2Json fails to locate the system version of the wast2json because the lastest wast2json command expects a filename or an option like --version or --help:

```bash
(base) starksm@Scotts-Mac-Studio runtime % wast2json
wast2json: expected filename argument.
Try '--help' for more information.
(base) starksm@Scotts-Mac-Studio runtime % wast2json --version
1.0.34 (git~1.0.34-22-g9fdd0242)
```

I have a simple 1 line PR to add the --version option to the system command check.

andreaTP:
Closed by #135

andreaTP:
@starksm64 would you mind sharing the steps you followed to get everything working on an M1?
It would be extremely beneficial for other users 🙏

PR #135, opened 2023-12-29
Pass the --version option to the wast2json command check

starksm64:
wast2json no longer returns a 0 exit status with no options

Addresses #134


andreaTP:
Thanks!

PR #140, opened 2024-01-08
Fix flaky testgen

### LINKED ISSUE #139: Flaky code generation
At times the test code generator fails with errors even in CI, we should be able to narrow down the issues and have a more stable build.

andreaTP:
Fix #139

### DIFF
--- a/test-gen-plugin/src/main/java/com/dylibso/chicory/maven/JavaTestGen.java
+++ b/test-gen-plugin/src/main/java/com/dylibso/chicory/maven/JavaTestGen.java
@@ -8,15 +8,12 @@ import com.dylibso.chicory.maven.wast.Command;
 import com.dylibso.chicory.maven.wast.CommandType;
 import com.dylibso.chicory.maven.wast.WasmValue;
 import com.dylibso.chicory.maven.wast.Wast;
-import com.fasterxml.jackson.databind.ObjectMapper;
-import com.github.javaparser.JavaParser;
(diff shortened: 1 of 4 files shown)


andreaTP:
Tested on quite a few CI runs, this should be stable now:
https://github.com/andreaTP/chicory/actions/runs/7451456239

test-gen-plugin/src/main/java/com/dylibso/chicory/maven/JavaTestGen.java
```diff
@@ -115,12 +102,7 @@ public CompilationUnit generate(
         var testClass = cu.addClass(testName);
         if (ordered) {
             testClass.addSingleMemberAnnotation(
-                    "TestMethodOrder",
-                    new ClassExpr(
-                            JAVA_PARSER
-                                    .parseType("MethodOrderer.OrderAnnotation")
-                                    .getResult()
-                                    .get()));
```
andreaTP, line 123: For the records, the JavaParser issue was in this type lookup.
The rest of the fixes are to tackle other failures that I have observed.

danielperano (approved):

test-gen-plugin/src/main/java/com/dylibso/chicory/maven/JavaTestGen.java
bhelx: Nice catch!

PR #141, opened 2024-01-09
Finish Element Section parsing

andreaTP:
... just the boring part ... but this will enable more work around tables and prevent the Parser to fail on valid "Elements" sections.

The reference specification is available here:
https://webassembly.github.io/spec/core/binary/modules.html#element-section


andreaTP:
I really don't know if the names of the classes are good enough, I tentatively followed the sentence:

> that represent the elems component of a module

But it looks odd ...

bhelx (approved):
The names are good for me. Elements is a little bit of a weird concept so this is to be expected i think.

danielperano (approved):

danielperano:
LGTM!

PR #147, opened 2024-01-11
Table.get and table.set specs

andreaTP:


wasm/src/main/java/com/dylibso/chicory/wasm/types/Value.java
```diff
@@ -9,11 +9,13 @@
...
 public class Value {
 
-    public static final Value TRUE;
+    public static final Value TRUE = Value.i32(1);
 
-    public static final Value FALSE;
+    public static final Value FALSE = Value.i32(0);
 
-    public static final Value REF_NULL;
+    public static final int REF_NULL_VALUE = 0;
```
andreaTP, line 16: I'm not convinced by this, let me iterate once more, back to draft.

andreaTP:
I think it clicked now, ready for review.

bhelx (approved):

PR #154, opened 2024-01-15
Version-aware wast2json resolve

danielperano:
On some Linux distros, the wabt version in the repos is rather old (in my case, Linux Mint has 1.0.27). This causes issues when the system version is older than the tests require.

This PR does a basic semantic version number check between the system-installed `wast2json` (if present) and the version required by the tests before delegating to the system install.

### DIFF
--- a/test-gen-plugin/src/main/java/com/dylibso/chicory/maven/Wast2JsonWrapper.java
+++ b/test-gen-plugin/src/main/java/com/dylibso/chicory/maven/Wast2JsonWrapper.java
@@ -233,14 +233,21 @@ public class Wast2JsonWrapper {
     }
 
     /**
-     * Compares the available wast2json version with the minimum required version
-     * using semantic version numbers. Assumes that the version is a 3 part semantic
-     * version number.
+     * Compares the available wast2json version with the minimum required version. It
+     * first checks for an exact match on the version strings, and if they do not
+     * exactly match it tries to parse and compare them using 3 part semantic version numbers.
      * @param actual the version of wast2json already installed
-     * @param required the minimum version of wast2json that's required
+     * @param required the version of wast2json that's required
      * @return true if the actual version meets the requirements, false otherwise
      */
     private boolean versionMatches(String actual, String required) {
+        actual = actual.trim();
+        required = required.trim();
+
+        if(actual.equals(required)){
+            return true;
+        }
+
         var actualComponents = splitVersion(actual);
         var requiredComponents = splitVersion(required);
 
@@ -261,6 +268,6 @@ public class Wast2JsonWrapper {
     }
 
     private String[] splitVersion(String v) {
-        return v.trim().split("\\.");
+        return v.split("\\.");
     }
 }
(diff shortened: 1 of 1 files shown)


bhelx (approved):
Seems like a good idea, thanks!

test-gen-plugin/src/main/java/com/dylibso/chicory/maven/Wast2JsonWrapper.java
```diff
@@ -228,4 +231,36 @@ public String wabtArchitectureName() {
...
+     * version number.
+     * @param actual the version of wast2json already installed
+     * @param required the minimum version of wast2json that's required
+     * @return true if the actual version meets the requirements, false otherwise
+     */
+    private boolean versionMatches(String actual, String required) {
+        var actualComponents = splitVersion(actual);
+        var requiredComponents = splitVersion(required);
+
+        if (actualComponents.length != 3 || requiredComponents.length != 3) {
```
andreaTP, line 247: Have you tried to compile locally wabt and check the version number?
I'm afraid of causing more trouble if, in the future, we'd need to compile locally for any reason and obtain a version with something like `+<commit sha>` ... but maybe it's already fine.

danielperano: That's a good point - pushing a small modification that tries to do an exact match before falling back to semantic version checking. If either fail, it defaults to downloading the required version.

andreaTP: Make sense to me, do you mind adding a couple of log messages so that is clear what its going on?

andreaTP (commented):
A couple of small notes, but lgtm after that 👍 thanks!

test-gen-plugin/src/main/java/com/dylibso/chicory/maven/Wast2JsonWrapper.java
```diff
@@ -152,26 +149,32 @@ private String resolveLatestVersion() {
...
         ProcessBuilder pb = new ProcessBuilder(WAST2JSON, "--version");
         pb.directory(new File("."));
-        pb.inheritIO();
         Process ps = null;
+        String systemVersion = null;
         try {
             ps = pb.start();
             ps.waitFor(1, TimeUnit.SECONDS);
+            if (ps.exitValue() != 0) {
+                System.err.println(ps.getErrorStream().toString());
```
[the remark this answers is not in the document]
danielperano: I don't think so - I think we should just treat the system version as unusable and continue with a downloaded one.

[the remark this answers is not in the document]
andreaTP: Oh, ok, this makes sense!
Can you switch to logging instead of `System.err`?

andreaTP (approved):
LGTM, thanks for bearing with me!

PR #155, opened 2024-01-15
fix: Fixes parsing of _start function

bhelx:
Extracted from the wasi3 branch https://github.com/dylibso/chicory/compare/main...wasi3#diff-dee877c864dc5f9b538d7a4fc0c15f9ed45276f095e548172bb3780ea6b55764

There seemed to be a bug in the logic that parses the _start function. getExport("_start") was giving me the first function in the function list and not the actual _start function.

BTW I created this bug by naively writing that parsing code. I think it was working because in our tests of _start, _start was the only function. But when there are other functions before it, it fails.


andreaTP (approved):
Lgtm! Thanks!

PR #156, opened 2024-01-15
fix: Fixes memory fill out of bounds error

bhelx:
I believe the purpose of this code was to do a bounds check before writing the memory, but `getMaximumSize()` returns the unit of pages. What i think was meant was to find the last memory address which would be something like `memory.buffer.size()`.

However, I think it's better to let the bytebuffer handle the bounds checking here. So removed the check and captured the indexoob error like we do with the other memory access methods.

### DIFF
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/Machine.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/Machine.java
@@ -481,9 +481,6 @@ public class Machine {
                             var val = stack.pop().asByte();
                             var offset = stack.pop().asInt();
                             var end = (size + offset);
-                            if (end > instance.getMemory().getMaximumSize()) {
-                                throw new WASMRuntimeException("out of bounds memory access");
-                            }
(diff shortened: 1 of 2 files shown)


runtime/src/main/java/com/dylibso/chicory/runtime/Machine.java
```diff
@@ -481,9 +481,6 @@ void eval(List<Instruction> code) throws ChicoryException {
                             var val = stack.pop().asByte();
                             var offset = stack.pop().asInt();
                             var end = (size + offset);
-                            if (end > instance.getMemory().getMaximumSize()) {
```
bhelx, line 484: Perhaps `getMaximumSize` should be renamed to reflect it's unit. Something like `getMaximumNumberOfPages`

bhelx:
There is something a little wacky here. fill is maybe not throwing the exception.

andreaTP (approved):
Nice, thanks! LGTM!

PR #157, opened 2024-01-16
fix: memory.fill should only have 1 immediate value

bhelx:
I believe memory.fill should only have 1 immediate value. I'm making sure this isn't some kind of core module incompatible thing with 2.0 or something. But this rust module I just compiled had only 1 immediate.


andreaTP (approved):
I think you are right, thanks for fixing it.
ref: https://webassembly.github.io/spec/core/binary/instructions.html#memory-instructions

PR #158, opened 2024-01-16
fix: comment out logger

bhelx:
This log statement makes everything a lot slower because it's on the hot path. Runs the tests about 30% faster on my machine. I assume that Java has to compute that string everytime too. but the conditional to not log is probably happening after the function is called.

Open to different suggestions about how to deal with this. But i feel like since it's mostly used by us i don't mind uncommenting it when i need it.


andreaTP (approved):
LGTM, this logging is just for us working on the core.

PR #159, opened 2024-01-16
[WIP] Imports layering

andreaTP:
I think, it's easier to look at the code and discuss a proposal 🙂

### DIFF
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/HostImports.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/HostImports.java
@@ -4,8 +4,6 @@ import java.util.Arrays;
 import java.util.stream.Collectors;
 
 public class HostImports {
-
-    private String name;
     private final HostFunction[] functions;
     private final HostGlobal[] globals;
     private final HostMemory[] memories;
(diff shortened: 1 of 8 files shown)


runtime/src/main/java/com/dylibso/chicory/runtime/HostImports.java
```diff
@@ -91,4 +107,42 @@ public FromHost[] getIndex() {
...
+                    var matching =
+                            finalFunctions.stream()
+                                    .filter(
+                                            fun ->
+                                                    fun.getModuleName().equals(f.getModuleName())
+                                                            && fun.getFieldName()
+                                                                    .equals(f.getFieldName()))
+                                    .findFirst();
+                    if (matching.isPresent()) {
+                        if (!matching.get().override()) {
```
danielperano, line 131: I think we should switch this so that the controlling `override()` flag is on the function being added rather than the original one. As it stands now, this would mean that every host function would have to explicitly allow being overridden, and this would further mean that the decision to override a specific function or not would be removed from the end developer, but I think the end developers should be able to override any imported function if they choose.

I'd also suggest that we throw if the override is not applied rather than silently using the previously present one.

danielperano: I went over this a bit too quickly before and missed some important bits - need to set aside a few minutes to trace through it in detail.

andreaTP: 🙏 please consider that this is just "sketched up"

andreaTP:
Closing as it's too old now ...

PR #160, opened 2024-01-16
chore: refactor POMs

lburgazzoli:
- centralize version properties
- add plugin management to improve reproducible builds
- replace outdate maven-project with maven-core


andreaTP (approved):
Nice cleanup! LGTM!

PR #162, opened 2024-01-16
Logging without overhead on the hot path

andreaTP:


andreaTP:
Looks to be impacting a little performance anyhow, probably best to keep it commented out ATM.

PR #164, opened 2024-01-16
WasiP1 Support (only a subset for now)

bhelx:
I've still got some work to do here. But opening up now. API looks like this.

```java
var fakeStdin = new ByteArrayInputStream("Benjamin".getBytes());
var wasiOpts = WasiOptions.builder().setStdout(System.out).setStdin(fakeStdin);
var wasi = new WasiP1(wasiOpts);
var imports = new HostImports(wasi.toHostFunctions());
var instance = Module.build(new File("src/test/resources/compiled/greet-wasi.rs.wasm"))
                        .instantiate(imports);
var run = instance.getExport("_start");
run.apply();
```

Outputs can be set as PrintStreams and stdin can be passed an InputStream. You can also inherit from the parent process with: `WasiOptions.build().inheritSystem()`.


runtime/src/main/java/com/dylibso/chicory/runtime/WasiOptions.java
```diff
@@ -0,0 +1,50 @@
+package com.dylibso.chicory.runtime;
+
+import java.io.InputStream;
+import java.io.PrintStream;
+
+public class WasiOptions {
```
andreaTP, line 6: an idiomatic Java builder pattern will look like this (from the user's perspective):

```java
WasiOptionsBuilder.withStdin(mystdin).withStdout(mystdout).build();
```

andreaTP (commented):
A couple of comments on the implementation.
The more I think about this the more I'm inclined to have a separate `wasi` module, wdyt?

andreaTP:
Feel free to ping me for assistance if you have any trouble with Maven :+1: happy to help!

bhelx:
> Feel free to ping me for assistance if you have any trouble with Maven 👍 happy to help!

be careful what you wish for 😈

runtime/src/main/java/com/dylibso/chicory/runtime/WasiOptions.java
bhelx: Is the pattern internally to just have a static class that looks exactly the same as the regular class? seems pretty wacky. https://www.digitalocean.com/community/tutorials/builder-design-pattern-in-java

Is there a better way?

runtime/src/main/java/com/dylibso/chicory/runtime/WasiP1.java
```diff
@@ -0,0 +1,201 @@
...
+    public WasiP1(WasiOptions opts) {
+        this.options = opts;
+    }
+
+    public HostFunction[] toHostFunctions() {
+        var fidx = 8; // 46
+        var functions = new HostFunction[fidx + 1];
+
+        functions[fidx--] =
+                new HostFunction(
```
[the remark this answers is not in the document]
bhelx: I'm gonna pick this up later. Since it's only two methods i feel like it's not worth doing this just yet.

runtime/src/main/java/com/dylibso/chicory/runtime/WasiOptions.java
andreaTP: It's the same thing but with mutable fields ... no magic...

bhelx:
I moved it to it's own package. Eventually we should move this to it's own module. Maybe have it's own unit tests and stuff. it's a bit too much for me at the moment. I want to just use this to more easily test the runtime.

compilation-support-image/compile-rust-wasi.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+while IFS= read -r line || [ -n "$line" ]
+do
+  echo "$line" >> tmp.rs
+done < "/dev/stdin"
+
+rustc tmp.rs --target=wasm32-wasi -C opt-level=0 -C debuginfo=0 -o tmp.wasm > /dev/null 2> /dev/null
```
bhelx, line 8: I couldn't actually get this to work on my machine. The docker image is x86, so i made my own. but for some reason it's failing to create these tmp files. It gets to this point and then says that there is no file. It must be failing to run line 5.

runtime/src/main/java/com/dylibso/chicory/runtime/WasiOptions.java
bhelx: I'm confused, i still need both a builder class and a normal class with the same instance variables and setters right?

runtime/src/main/java/com/dylibso/chicory/runtime/wasi/WasiP1.java
```diff
@@ -0,0 +1,756 @@
...
+        // maybe we want a more explicit way of doing this though
+        this(new WasiOptions());
+    }
+
+    public WasiP1(WasiOptions opts) {
+        this.options = opts;
+        LOGGER.log(
+                System.Logger.Level.WARNING,
+                "Use of WASIP1 is experimental and will only work for the simplest of use cases.");
+    }
```
bhelx, line 32: I think this is sufficient to warn people for now. Not gonna document it just yet either.

runtime/src/main/java/com/dylibso/chicory/runtime/wasi/WasiP1.java
```diff
@@ -0,0 +1,756 @@
...
+
+    public HostFunction[] toHostFunctions() {
+        var fidx = 55;
+        var functions = new HostFunction[fidx + 1];
+
+        functions[fidx--] =
```
bhelx, line 38: I stubbed out as many WASI p1 functions as i could find. Oftentimes they are imported and never called. But if they are, it should trap here with the name of the wasi function that is not implemented.

The only real implementations at the moment are fd_read and fd_write and only for file descriptors 0,1, and 2

runtime/src/main/java/com/dylibso/chicory/runtime/wasi/WasiP1.java
```diff
@@ -0,0 +1,756 @@
...
+                                    break;
+                                default:
+                                    throw new WASMRuntimeException("We don't yet support fd " + fd);
+                            }
+                            if (stream == null) {
+                                throw new WASMRuntimeException(
+                                        "The program attempted an fd_read from"
+                                                + fd
+                                                + "but this was not granted.");
+                            }
```
bhelx, line 307: In the future we'll probably keep a map of open files keyed by their fd and do a lookup here. that will require some of the directory mapping stuff which i want to avoid.

runtime/src/main/java/com/dylibso/chicory/runtime/WasiOptions.java
andreaTP: > instance variables and setters

you need only the constructor and the getters

compilation-support-image/compile-rust-wasi.sh
andreaTP: That's wired, thanks for reporting!
Unfortunately, I don't have an ARM machine to test, so I'll need someone's help to debug this one.

andreaTP (changes requested):
LGTM after merging #166 in this branch, and let's not forget to take a closer look at the compilation issue.

runtime/src/main/java/com/dylibso/chicory/runtime/WasiOptions.java
andreaTP: resolved in #166

andreaTP:
Another follow-up task is gonna be documentation, updating the readme etc., not required in this PR as it's already big enough.

andreaTP (approved):
LGTM, we would need to iterate on the compilation/scripts etc. but that's fine and doesn't hurt.

bhelx:
> Another follow-up task is gonna be documentation, updating the readme etc., not required in this PR as it's already big enough.

Agreed. I think let's hold off until we're more confident though. I'd prefer for this to be hidden for the time being. I just want to use it to get some more modules in here for testing purposes.

compilation-support-image/compile-rust-wasi.sh
bhelx: well, docker images are ISA specific. But i may be able to detect if this is running in mac and force it to emulate x86. Will try some things out don't worry about it for now.

PR #165, opened 2024-01-16
Included all memory specs in tests.

danielperano:
Fixed 1 minor issue in memory - exception thrown did not contain the "out of bounds memory access" message expected.


andreaTP (approved):
Awesome!

PR #166, opened 2024-01-17
Wasi3 review

andreaTP:
I think I have tackled all the outstanding points, great work @bhelx !

I have attempted to re-compile everything, unfortunately, I'm unable to reproduce the output of `greet-wasi.rs` and `hello-wasi.rs` (seems like the exported `_start` is not generated), but we can file an issue and get back to it later on.


bhelx (approved):
Looks good, thanks for picking this up :)

.github/workflows/ci.yaml
```diff
@@ -4,7 +4,6 @@ on:
   push:
     branches: [ main ]
   pull_request:
-    branches: [ main ]
```
bhelx, line 7: Is this meant to be removed?

andreaTP: yes, it doesn't hurt, it will run the CI on any PR even not targeting `main`

bhelx: oh i see, makes sense

PR #169, opened 2024-01-17
table_grow.wast spec passing

andreaTP:
Getting `table_grow.wast` to pass, fixed a few bugs around to get there.
This is the reference:
https://webassembly.github.io/spec/core/exec/instructions.html#xref-syntax-instructions-syntax-instr-table-mathsf-table-grow-x

### DIFF
--- a/wasm/src/main/java/com/dylibso/chicory/wasm/types/ValueType.java
+++ b/wasm/src/main/java/com/dylibso/chicory/wasm/types/ValueType.java
@@ -9,7 +9,7 @@ public enum ValueType {
     I64(0x7e),
     I32(0x7f),
     V128(0x7b),
-    FuncRef(0x7a),
+    FuncRef(0x70),
     ExternRef(0x6f);
 
     private final long id;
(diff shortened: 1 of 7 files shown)


danielperano (approved):

wasm/src/main/java/com/dylibso/chicory/wasm/types/ValueType.java
```diff
@@ -9,7 +9,7 @@ public enum ValueType {
     I64(0x7e),
     I32(0x7f),
     V128(0x7b),
-    FuncRef(0x7a),
+    FuncRef(0x70),
```
bhelx, line 12: Oops, did i do this 😆 

nice catch!

bhelx (approved):

wasm/src/main/java/com/dylibso/chicory/wasm/types/ValueType.java
andreaTP: Or me :sweat:

PR #173, opened 2024-01-18
[chore] add a script to easily list missing specs

andreaTP:
It's a bit hacky and not sure about portability ...
Not sure if we should have it in the main repo, I can even keep It local, but it helps when eyeballing to find what to tackle next.


bhelx (approved):
Nice thinking :)

PR #175, opened 2024-01-19
fix the wasi sub-module version

andreaTP:


andreaTP:
@bhelx looks like I cannot directly push to `main`, sorry for the noise with subsequent PRs.

PR #179, opened 2024-01-22
Include call_indirect.wast spec & fix issues.

danielperano:
Added support for runtime type checking of indirect function call signatures.


andreaTP (approved):
LGTM! Thanks! Does it need to be rebased?

danielperano:
Yep! Didn't see that merge come in. I'll rebase.

andreaTP (approved):
Thanks for the cleanup!

Issue #151, opened 2024-01-13
Logging + Android


dgmltn:
Android does not implement JDK Platform Logging / `System.Logger`, even when targeting Java >= 9 (I'm trying with JAVA 17). It is compatible with `java.util.logging.Logger`, though (see: https://source.android.com/docs/core/tests/debug/understanding-logging ). 

But it would be easier from my perspective if I could inject an instantiation of a logger interface on app startup.

andreaTP:
👋 Hi @dgmltn !
First of all thanks for your interest and for taking the time to file this issue 🙏 

I'm not an Android expert, would it be possible to polyfill the JDK Platform logging only for Android builds?
If not, we still have options, and implementing a custom SPI for logging with a couple of backends should be pretty straightforward.

This issue, although, triggers the question of Android support which we are not tracking at the moment.
If we wanna guarantee the support we should setup [a linting step](https://github.com/microsoft/kiota-java/blob/9ae7ae2ed2eafeadda88947e9485b34606452621/.github/workflows/api-level-lint.yml#L26) but this has a lot of implications toward the Java features we are going to be allowed to use etc.

bhelx:
I thank you to @dgmltn!  

The original goal for me was to support as many platforms as possible. I ignorantly thought, as long as we had no dependencies, the way to do that would be to support as many JDKs versions as possible. But I guess this shows that there are some subtle differences that we need to watch out for.

With that in mind I think it seems kind of silly to hold back Android support just because of how we choose to do logging. Given there are over 3 billion active Android devices out there, I'd prefer to more explicitly support Android even if that means we lose Java 8 support. We could move to more of "wait and see" on Java 8 support.

What do you think @dgmltn ? Curious how we can explicitly support Android and also, as @andreaTP suggested, make sure our CI is protecting us from breaking it.

dgmltn:
Thanks for the responses, @bhelx @andreaTP ! Yeah, I agree that an automated test would help prevent from accidentally breaking Android. I'm not sure there's an "easy" way to maintain Android compatibility aside from trial and error.

But for the logging issue, I thought of 2 solutions:
1. Use a homegrown logger `interface`, that the user could inject.
2. Use SLF4J. I've seen other libraries use this (that also target Android), but I'm not familiar with it. Here's the library I'd use to get that to work with my app: ( https://github.com/nomis/slf4j-android ).

Either way (or some other solution too?) I'd be willing to test out a branch build with a POC Android app before it goes into `main`.

[...]

PR #182, opened 2024-01-23
feat: provide a pluggable logging abstraction and some refactoring

lburgazzoli:
This PR was meant to help fixing https://github.com/dylibso/chicory/issues/151 but it turned out to be a little bit more invasive than expected so a number of other things have also been amended.

The main change is the introduction of a module builder:

```java
var module = Module.builder("my-module.wasm")
    .withLogger(new MyLogger())
    .withType(ModuleType.TEXT)
    .build()
```

This would allow to _easily_ add further configuration options later on.

Also the builder is in charge to read the actual module bits (from `File`, `Path`, `byte[]`, classpath, ect) instead of delegating it to the `Parser` class.

Let me know if it goes to far.


bhelx:
Thanks you! I think changing the public API will add a little more scrutiny but I think it's the correct direction. We probably want to apply a builder pattern on some of these public classes because there will increasingly be more options for constructing and validating them and this is the most flexible API choice.

Will leave it to @andreaTP to say if the pattern fits our other builder patterns.

andreaTP:
I like the direction, it is a good enhancement over the current API, but I want to review this PR carefully as it's growing on the big side.
@lburgazzoli let me know when you think you are close to the end 👍

lburgazzoli:
no additional changes expected in this pr to avoid making it a bigger monster:)

will open new prs as follow up

andreaTP:
Sounds good! Can you have a look at the CI? Seems like something accidentally broke 😢

lburgazzoli:
> Sounds good! Can you have a look at the CI? Seems like something accidentally broke 😢

ok, ci is happy now

bhelx (approved):
Looks good to me, but i'll give @andreaTP the last word on this one since it has implications on the public interface and the DX.

runtime/src/test/java/com/dylibso/chicory/runtime/ModuleTest.java
```diff
@@ -273,7 +248,7 @@ public void shouldRunKitchenSink() {
     // // check with: wasmtime src/test/resources/wasm/memories.wat.wasm --invoke
     // run 100
     // var instance =
-    // Module.build("src/test/resources/wasm/memories.wat.wasm").instantiate();
+    // Module.builder(("src/test/resources/wasm/memories.wat.wasm").instantiate();
```
andreaTP, line 251:
```suggestion
    // Module.builder("wasm/memories.wat.wasm").instantiate();
```

runtime/src/test/java/com/dylibso/chicory/runtime/ModuleTest.java
```diff
@@ -155,36 +137,33 @@ public void shouldWorkWithStartFunction() {
...
         var module =
-                Module.build(new File("src/test/resources/compiled/start.wat.wasm"))
+                Module.builder("compiled/start.wat.wasm")
+                        .build()
                         .instantiate(new HostImports(funcs));
         var start = module.getExport("_start");
         start.apply();
-        assertTrue(printer.times() > 0);
+
+        assertThat(count).hasValueGreaterThan(0);
```
andreaTP, line 146: I would like to avoid adding `assertj` as a dependency only for this use case, not against in general but I believe that this is not compelling enough just yet.
```suggestion
        assertTrue(count > 0);
```

andreaTP (changes requested):
Really nice work @lburgazzoli much appreciated!

A few nitpicks but looks almost ready to merge 👍

bhelx:
Apologies about all the code changing underneath it. We've been merging a lot of small changes the past couple days.

lburgazzoli:
> Apologies about all the code changing underneath it. We've been merging a lot of small changes the past couple days.

no worries, can you double check to verify I've not made any mistake while rebasing and fixing conflicts ?

andreaTP (approved):
LGTM, great work!

bhelx:
Reminder: we should follow up with a small readme section about how to use this when we get the chance. Not urgent though!

bhelx:
For the Module builder pattern change i think this small change takes care of it: https://github.com/dylibso/chicory/pull/191

PR #184, opened 2024-01-23
Discontinue usage of BigInteger

dmlloyd:
Usage of BigInteger is expensive and not necessary. Also correct some inconsistencies in numerical conversions and use intrinsics for unsigned comparison.

### DIFF
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/Machine.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/Machine.java
@@ -16,10 +16,6 @@ public class Machine {
 
     private static final System.Logger LOGGER = System.getLogger(Machine.class.getName());
 
-    public static final double TWO_POW_63_D = 0x1.0p63; /* 2^63 */
-
-    public static final float TWO_POW_64_PLUS_1_F = 1.8446743E19F; /* 2^64 + 1*/
-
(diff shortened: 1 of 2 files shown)


bhelx:
Thank you @dmlloyd !

runtime/src/main/java/com/dylibso/chicory/runtime/Machine.java
```diff
@@ -16,10 +16,6 @@ public class Machine {
 
     private static final System.Logger LOGGER = System.getLogger(Machine.class.getName());
 
-    public static final double TWO_POW_63_D = 0x1.0p63; /* 2^63 */
-
-    public static final float TWO_POW_64_PLUS_1_F = 1.8446743E19F; /* 2^64 + 1*/
```
bhelx, line 21: These were unused right? I think @thomasdarimont added them. Maybe we refactored out and forgot to remove? Some linting on unused code might be helpful here.

dmlloyd: `TWO_POW_63_D` was used in the old code. `TWO_POW_64_PLUS_1_F` was unused.

andreaTP:
Nice one! 🎉 Thanks!

PR #185, opened 2024-01-23
Make `WasmFunctionHandle` accept an `Instance`

dmlloyd:
This will allow support for multiple memories and other features in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,10 +150,10 @@ Let's write that host function:
 
 ```java
 var func = new HostFunction(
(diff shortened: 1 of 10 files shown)


bhelx:
Looks good, would you mind also changing the readme? https://github.com/dylibso/chicory/blob/main/README.md?plain=1#L152

There might be a piece of docs or two to change

bhelx:
Thanks! trying to keep our docs minimal so our API can be as flexible as possible.

andreaTP:
🙏 Thanks! Appreciated!

PR #186, opened 2024-01-23
Bridge between `MethodHandle` and `WasmFunctionHandle`

dmlloyd:
This allows WASM functions to be called from Java, and Java functions to be called by WASM, with automatic argument boxing and unboxing. This is just an initial sketch; we would want more validation, flexibility, boxing and unboxing strategies, etc.

Comments welcome!

### DIFF
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/WasmFunctionHandle.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/WasmFunctionHandle.java
@@ -1,6 +1,10 @@
 package com.dylibso.chicory.runtime;
 
 import com.dylibso.chicory.wasm.types.Value;
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandleProxies;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodType;
 
 /**
  * Represents a Java function that can be called from Wasm.
@@ -8,4 +12,66 @@ import com.dylibso.chicory.wasm.types.Value;
 @FunctionalInterface
 public interface WasmFunctionHandle {
     Value[] apply(Instance instance, Value... args);
+
+    /**
+     * Create a {@code WasmFunctionHandle} which calls the given method handle with possible argument and return
+     * value type conversions.
+     *
+     * @param mh the method handle (must not be {@code null})
+     * @return the wrapper function handle (not {@code null})
+     * @throws IllegalArgumentException if the method handle cannot be converted to a {@code WasmFunctionHandle}
+     */
+    static WasmFunctionHandle forMethodHandle(MethodHandle mh) {
+        return WasmFunctionHandles.wasmFunctionHandle(mh);
+    }
+
+    /**
+     * Get a method handle for this {@code WasmFunctionHandle} with the given type, converting arguments and return
+     * types as necessary.
+     *
+     * @param type the type of the desired method handle (must not be {@code null})
+     * @return the method handle (not {@code null})
+     * @throws IllegalArgumentException if this {@code WasmFunctionHandle} cannot be converted to a method handle
+     *  of the given type
+     */
+    default MethodHandle asMethodHandle(MethodType type) {
+        return WasmFunctionHandles.methodHandle(this, type);
+    }
+
+    /**
+     * Get a {@code WasmFunctionHandle} as a functional interface. Calling the functional interface method will
+     * call this {@code WasmFunctionHandle}.
+     * <p>
+     * The functional interface method must accept a {@link Instance} as its first argument.
+     *
+     * @param interfaceType the interface type (must not be {@code null})
+     * @param methodType the functional interface method (must not be {@code null}) (TODO: we could maybe detect it)
+     * @return the functional interface instance
+     * @param <T> the functional interface type
+     * @throws IllegalArgumentException if this {@code WasmFunctionHandle} cannot be converted to a method handle
+     *  of the given type
+     */
+    default <T> T asFunctionalInterface(Class<T> interfaceType, MethodType methodType) {
+        return MethodHandleProxies.asInterfaceInstance(interfaceType, asMethodHandle(methodType));
+    }
+
+    static void main(String[] args) throws Throwable {
+        WasmFunctionHandle addUp =
+                WasmFunctionHandle.forMethodHandle(
+                        MethodHandles.lookup()
+                                .findStatic(
+                                        WasmFunctionHandle.class,
+                                        "addUp",
+                                        MethodType.methodType(int.class, int.class, int.class)));
+        MethodHandle mh =
+                addUp.asMethodHandle(
+                        MethodType.methodType(int.class, Instance.class, int.class, int.class));
+        int res = (int) mh.invokeExact((Instance) null, 12, 50);
+        System.out.println(res);
+    }
+
+    static int addUp(int a, int b) {
+        System.out.println("I'm being called from WASM");
+        return a + b;
+    }
 }
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/WasmFunctionHandles.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/WasmFunctionHandles.java
@@ -1,0 +1,244 @@
+package com.dylibso.chicory.runtime;
+
+import static java.lang.invoke.MethodHandles.lookup;
+
+import com.dylibso.chicory.wasm.types.Value;
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandleProxies;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodType;
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * Implementation helpers for {@link WasmFunctionHandle}.
+ */
+final class WasmFunctionHandles {
+    private WasmFunctionHandles() {}
+
+    static WasmFunctionHandle wasmFunctionHandle(MethodHandle mh) {
+        Objects.requireNonNull(mh, "mh");
+        MethodType type = mh.type();
+        // the first arg may or may not be an `Instance`
+        int cnt = type.parameterCount();
+        if (cnt > 0) {
+            Class<?> argType = type.parameterType(0);
+            if (argType != Instance.class) {
+                // drop the instance argument
+                mh = MethodHandles.dropArguments(mh, 0, Instance.class);
+                type = mh.type();
+                cnt = type.parameterCount();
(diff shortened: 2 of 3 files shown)


andreaTP:
Hi @dmlloyd 👋 and thanks for taking the time to put this together!

Looks a really desirable improvement, but I'm worried this is not going to fly when compiled with `native-image`, or am I wrong?
For context, we discussed something similar [here](https://github.com/dylibso/chicory/pull/65).

I'm also interested in hearing what's your opinion on tackling this UX challenge with a code generator?

dmlloyd:
Since this is only using (modern) reflection (presently), it should interoperate with native images just fine (assuming one does the appropriate reflection registrations). Additionally, even if it used code generation, it is still not too hard to make it work with native images (it's largely a question of build-time-initializing your code generators; see Quarkus for a prime example).

Method handles in general make pretty good low-level interfaces for bridging between code. A mechanism like this could allow interpreted and (future) compiled WASM methods to interoperate, for example.

andreaTP:
Thanks for getting back!

> Since this is only using (modern) reflection (presently), it should interoperate with native images just fine (assuming one does the appropriate reflection registrations).

So far, we have been strict about disallowing any usage of reflection, I think we want to keep it this way at least for the Low Level API exposed by the `wasm` and `runtime` modules.
When tinkering about a High Level API more possibilities are open, I'm more inclined toward fully static code generation, but I don't want to close the door to this approach.

A possible middle ground would be to have this bridge in a separate Maven sub-module that can optionally be included on top of the Low Level API (i.e. similar to what we did for Wasi), after publishing it we can track adoption by using Maven Central Statistics and re-evaluate the tradeoff with more data. wdyt?

> A mechanism like this could allow interpreted and (future) compiled WASM methods to interoperate, for example.

I don't have elements to support or deny, I believe that this discussion needs to happen when we have more stable communication about the upcoming Component Model Specification.

Interested in hearing more opinions, cc. @bhelx @danielperano @lburgazzoli

dmlloyd:
> Thanks for getting back!
> 
> > Since this is only using (modern) reflection (presently), it should interoperate with native images just fine (assuming one does the appropriate reflection registrations).
> 
> So far, we have been strict about disallowing any usage of reflection, I think we want to keep it this way at least for the Low Level API exposed by the `wasm` and `runtime` modules. 

Why? What does this buy you?

> When tinkering about a High Level API more possibilities are open, I'm more inclined toward fully static code generation, but I don't want to close the door to this approach.
> 
> A possible middle ground would be to have this bridge in a separate Maven sub-module that can optionally be included on top of the Low Level API (i.e. similar to what we did for Wasi), after publishing it we can track adoption by using Maven Central Statistics and re-evaluate the tradeoff with more data. wdyt?

As someone who has authored and published dozens of open-source libraries, I think this is probably an overly conservative vantage point. I think that being (for example) the fastest, or most JVM-idiomatic implementation is much more practically useful for most people than being the most conservative about things like reflection and JVM support. That's just my opinion though, and my opinion reflects my own use cases, and thus I acknowledge that there is considerable bias in those use cases.

andreaTP:
> Why? What does this buy you?

Maintainability, straightforward code and out-of-the-box experience on any Java target coming from any language that runs on the JVM.

>  I think that being (for example) the fastest, or most JVM-idiomatic implementation

For reference: those are [the goals](https://github.com/dylibso/chicory?tab=readme-ov-file#goals) we agreed on so far, we can challenge them and iterate if necessary.

> my opinion reflects my own use cases, and thus I acknowledge that there is considerable bias in those use cases.

Agreed, I probably come from a very different background/experiences and I have spent quite enough time of my life fixing those things, hence I'm on the fence here.
All in all, I'm convinced that we can end up with a clean implementation and design of the Low Level Interpreter with no need to rescue to reflection.

bhelx:
I think it's an interesting idea! I'm a bit on the fence as well though. Mostly because I'm having trouble mentally parsing what the code is doing. Admittedly I'm not as fluent in Java as you all, but I think if we're to put in some indirection it needs to buy us something and preferably something external like speed, or safety, or a big developer experience improvement (for the end user). I know it might seem like we're being overly conservative, I think that's a fair criticism, but I also think in this early stage of development it makes sense to keep complexity out as long as possible.

danielperano:
@dmlloyd freakin' wow! This is amazing - ever since I first saw the current API I've wanted something like this, though this is even better than I would have come up with (love the functional interface support).

> All in all, I'm convinced that we can end up with a clean implementation and design of the Low Level Interpreter with no need to rescue to reflection.

Agreed, though I have two counterpoints I'd offer for this case: (a) I wouldn't personally count method handles as reflection, and (b) I don't think this interferes with keeping reflection out of the Low Level Interpreter (LLI).
- (a) might be a bit semantic. I think of reflection as discovering (& generally accessing) information about type, methods, and fields of an object at runtime. Method handles are going the opposite direction - given known information about the name & type of a method, you're obtaining and invoking what is effectively little more than a function pointer with type information.
- (b) For now I think it's fine to merge into core as-is. We already know the API is likely to change extensively, so I don't think it's too intrusive to merge as-is and then later move into its own module if we find a good reason to not include code involving method handles in the core interpreter module.

> ... I think if we're to put in some indirection it needs to buy us something and preferably something external like speed, or safety, or a big developer experience improvement (for the end user).

This is the big "gotcha" on pretty much all of the above by the time we get to AOT compilation. This (or something very much like this, anyway) should be able to produce a direct `MethodHandle` to AOT'd WASM exports and invoke them *without static compilation* with `invokeExact` - which means exceedingly small overhead compared to directly compiling against pre-emitted bytecode (which isn't possible in any project that loads and uses WASM dynamically at runtime). Because method handle creation is type checked, invoking them is safe because any mismatches between the WASM function signature and your expectation of it will be caught by the JVM's MethodHandle API.

For now, it's a definite +1 on developer experience. I think that the current `Value[] apply(Value... args)` is the best way to handle things within the interpreter (and as a low-level interface to the interpreter), but this style of interface is so dreadful to use from higher-level code that the first thing you've really got to practically do is write a wrapper API that lets you declare and invoke the functions you want without manually packing and unpacking arguments and return types. Directly calling these is far too error-prone, and it's horrifically tedious, as I discovered with [Chipmunk](https://github.com/MyWorldLLC/Chipmunk/) (which now supports bi-directional interop mechanisms very similar in spirit to the above - Chipmunk can bind directly to host Java code at runtime, and Chipmunk objects can be proxied into instances of Java interfaces). It's too much to ask end users of Chicory to come up with their own way of doing this if they don't want to interact with the low-level `apply()` API (and honestly, I don't think anyone is going to want to except to create a binding layer to fulfill a similar need to this one).

In my opinion, the sooner we have something like this the better. It's easy to keep encapsulated in its own layer above the low-level API, and we can move it into an optional module at any time if we really feel the need to. I'm quite familiar with this type of use of the MethodHandle API, and I don't think this is going to be much of an ongoing maintenance issue - MethodHandle code is a royal pain to initially write (because it's so strict about safety and correctness), but it's fairly easy to debug (since it's fail-fast at creation time) and after it's stabilized it pretty much just sits there until something significant changes on either the caller or callee end.

bhelx:
> For now, it's a definite +1 on developer experience. I think that the current Value[] apply(Value... args) is the best way to handle things within the interpreter (and as a low-level interface to the interpreter), but this style of interface is so dreadful to use from higher-level code that the first thing you've really got to practically do is write a wrapper API that lets you declare and invoke the functions you want without manually packing and unpacking arguments and return types. 

I think this is the most compelling case to me, but I'm still unsure and need to spend some time thinking about it. We should probably discuss in the meeting tomorrow to help educate me. I'd particularly like to learn how this might keep developers from making mistakes. I'm not super concerned about the low-levelness of the API right now. This first core API is going to be low level and it needs to be the most flexible API we offer. Once we iterate on it and get some confidence I think it will be worth creating a layer that uses it for a higher-level experience.

andreaTP:
Thanks a lot for sharing your opinions @danielperano , super interesting!
My thoughts are aligned with @bhelx but happy to discuss it further.

> (b) I don't think this interferes with keeping reflection out of the Low Level Interpreter (LLI)

I fully agree here, but I'd consider this a fundamental aspect.
If we move on with this either we have it sorted out in the first PR or expect me to split things up right after.

On the other point I have to differ:

> (a) I wouldn't personally count method handles as reflection

I consider this reflection at its core, the functionality offered is, basically the same as `DynamicProxy`.
I agree that the code is elegant, but the complexity(and the underlying mechanism) is evident to me.

dmlloyd:
> Thanks a lot for sharing your opinions @danielperano , super interesting! My thoughts are aligned with @bhelx but happy to discuss it further.
> 
> > (b) I don't think this interferes with keeping reflection out of the Low Level Interpreter (LLI)
> 
> I fully agree here, but I'd consider this a fundamental aspect. If we move on with this either we have it sorted out in the first PR or expect me to split things up right after.

I've left this as a draft specifically because it is really not ready yet; it's just an early sketch to show some things that are possible. In particular, the functional interface factory was almost an afterthought; I just added it in because it shows how this could be done.

The main idea of this PR is to suggest that `MethodHandle` is possibly the best representation of a callable function available to any JVM language, due to being able to represent the type of the thing being called (nearly) exactly without any boxing (see the Zulip discussion around boxing of return types for the exception to this rule).

When I wrote the Java bytecode interpreter in [qbicc](https://github.com/qbicc/qbicc), I had a very similar interface representing a callable function which I implemented by the interpreter as well as having implementations for host functions. What first pushed me in this direction was the high maintenance burden of maintaining hundreds and hundreds of lambdas or classes for all of the needed host functions. Being able to put methods on a class was significantly easier. From there it didn't take me long to realize that using `MethodHandle` would allow me to completely skip all intermediate boxing steps.

> On the other point I have to differ:
> 
> > (a) I wouldn't personally count method handles as reflection
> 
> I consider this reflection at its core, the functionality offered is, basically the same as `DynamicProxy`. I agree that the code is elegant, but the complexity is evident to me.

`MethodHandle` is integral to the Java language - in fact, you're using one (or more), along with a corresponding dynamic proxy, every time you use a lambda or a method reference, or concatenate two strings, among other things. A quick and hacky grep through the class files of the `runtime` module's `target/classes` directory for constant pool references for `MethodHandle` suggests there may be 300 or so usages of them already, just in that module.

bhelx:
We chatted about this in the team meeting yesterday. I think Andrea and I were still on the fence, but Daniel made some strong arguments about needing this for developer experience. We all agreed that we eventually need this layer but still feel like it's a little too early and want to iterate on the API a little more. So let's just keep this going in draft for now and revisit once we feel like the API is stabilizing and we have a few people who've stress tested it.

PR #188, opened 2024-01-24
Enforce no star imports

andreaTP:
Follow up from this: https://github.com/dylibso/chicory/pull/182#discussion_r1464807326

As we see more contributors coming to the repo it's not unusual to see big diffs for small changes due to the different IDE settings.
This PR prohibits the usage of star imports(which is anyhow a good practice) and enforces it through Spotless.

cc. @lburgazzoli


danielperano (approved):

bhelx:
Makes sense. I was surprised you guys allowed me to do star imports this long :)

andreaTP:
> I was surprised you guys allowed me to do star imports this long :)

hahaha

PR #190, opened 2024-01-24
Move Machine.eval() switch bodies into static methods

enebo:
Looking at JITWatch it seems Machine.eval() is too big to make it out of the Java bytecode interpreter.  I reduced the size by moving the bodies of nearly all of the switch statements into static methods.

After performing this I can see eval being compiled by C2 and several static methods representing asm opcode impls also C2 compiling.

A super scientific benchmark result was made from running a loop of fib calls from the example in the README:  before(13.5s), after (4.6s).

The other suggestion of moving this to a table of lambdas is partially shortened with this PR since all these impls have been parameterized into methods. Converting them to lambdas should be less work than starting from before this PR.  The only difficulty is deciding how to order them (I did not do these in a particular order so someone may be moving bodies around).

### DIFF
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/Machine.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/Machine.java
@@ -24,15 +24,22 @@ public class Machine {
 
     public Machine(Instance instance) {
         this.instance = instance;
-        this.stack = new MStack();
+        stack = new MStack();
         this.callStack = new Stack<>();
     }
 
     public Value[] call(int funcId, Value[] args, boolean popResults) throws ChicoryException {
-        return call(funcId, args, null, popResults);
+        return call(stack, instance, callStack, funcId, args, null, popResults);
     }
 
-    public Value[] call(int funcId, Value[] args, FunctionType callType, boolean popResults)
+    public static Value[] call(
+            MStack stack,
+            Instance instance,
+            Stack<StackFrame> callStack,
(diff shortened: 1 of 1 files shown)


bhelx:
> The only difficulty is deciding how to order them (I did not do these in a particular order so someone may be moving bodies around).

For the most part i've been trying to order by op code. I'm not sure if we've kept that consistent or not.

bhelx:
This is great! Adds a little bit of indirection but still very clear where the code is 😃

danielperano (approved):

enebo:
@bhelx We do this in JRuby's lexer (hand-written) and it makes jumping to what you want quicker.  If this moves to a table of lambdas this will remove that benefit but it may be faster yet.  That obviously will remove that benefit but if it ends up quicker yet then it is worth it.

enebo:
DOH.  I will fix conflicts.  2 commits came in between my last merge and opening this.

andreaTP (approved):
LGTM, still really readable and nice, we have been merging a few things and we caused some conflicts, feel free to ping as soon as you resolve them so that we can merge this without additional pain on your side!


For @bhelx :

> For the most part i've been trying to order by op code. I'm not sure if we've kept that consistent or not.

We have not 😢 . We should probably review the current order to match, e.g. the specification, we have been a bit too undisciplined.

enebo:
@andreaTP I just pushed now!

PR #191, opened 2024-01-24
doc: Update README with Module builder pattern change

bhelx:


README.md
```diff
@@ -55,7 +55,7 @@ import com.dylibso.chicory.runtime.Instance;
 import java.io.File;
 
 // point this to your path on disk
-Module module = Module.build(new File("./factorial.wasm"));
+Module module = Module.builder(new File("./factorial.wasm")).build();
```
bhelx, line 58: Maybe we want to recommend this pattern?:

```suggestion
Module module = Module.builder("./factorial.wasm").build();
```

andreaTP: I think we should keep the `File` here, as `factorial.wasm` needs to be available on the class path to be able to load it with your suggestion.

andreaTP (approved):
That's the only occurence?
WoW! We have been extremely good at avoiding over-documentation 🙂 

Less importantly:
we probably need also to update this section: https://github.com/dylibso/chicory?tab=readme-ov-file#modules
or just drop it 😇

bhelx:
>  WoW! We have been extremely good at avoiding over-documentation 🙂

yeah, admittedly i've love more documentation. But the API is gonna be changing so rapidly it will be really tricky to keep up to date right now.

PR #192, opened 2024-01-24
Use newer "getter" idiom

dmlloyd:
Modern Java is moving towards the idiom of using `xxx()` rather than `getXxx()` for getters. Setters are not migrating as aggressively, so leave them for now.


andreaTP (approved):
I like a lot how this reads! Thanks a lot for taking the time and making the changes.

We need to update the `Readme` accordingly, but if it's too much, we can do it in a follow-up task.

PR #193, opened 2024-01-24
Do not use `java.util.Stack`

dmlloyd:
It uses synchronization, which is rarely desirable. Use `ArrayDeque` instead.


andreaTP (approved):
Lgtm thanks for the improvement!
This looks like something that would improve performance, have you done any, even rough, benchmarks?

dmlloyd:
No, I haven't benchmarked this particular change in any place I've made it since I convinced myself it is always worthwhile, over 15 years ago :)

andreaTP:
Haha I fully trust you on this one!
Thanks for the change!
Lgtm

andreaTP (approved):

bhelx:
> It uses synchronization

I didn't know that. we definitely don't need synchronization here

PR #195, opened 2024-01-25
Make `Machine` non-public

dmlloyd:
The `Machine` class is an implementation detail. Callers cannot meaningfully interact with it, so it should be non-public.


andreaTP (approved):
Agree, thanks for making the change!

PR #196, opened 2024-01-25
Make `CustomSection` immutable, introduce hierarchy

dmlloyd:
The `CustomSection` class is mutable, including its name. Since this class is added to a `Map` based on its name, allowing the `name` property to be mutable can cause subtle bugs. Additionally, allowing section contents to be mutable after parsing can cause inconsistencies. Finally, formalize the custom section hierarchy somewhat to avoid redundant memory usage and to release the byte array when it is no longer needed.

### DIFF
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/Module.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/Module.java
@@ -22,7 +22,7 @@ import com.dylibso.chicory.wasm.types.Global;
 import com.dylibso.chicory.wasm.types.Import;
 import com.dylibso.chicory.wasm.types.ImportDescType;
 import com.dylibso.chicory.wasm.types.Instruction;
-import com.dylibso.chicory.wasm.types.NameSection;
+import com.dylibso.chicory.wasm.types.NameCustomSection;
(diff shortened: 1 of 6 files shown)


wasm/src/main/java/com/dylibso/chicory/wasm/types/NameSection.java
```diff
@@ -6,18 +6,30 @@
 import java.util.Collections;
 import java.util.List;
 
+/**
+ * The "name" custom section.
+ */
 public class NameSection extends CustomSection {
```
andreaTP, line 12: Since we are at it:
```suggestion
public class NameCustomSection extends CustomSection {
```

andreaTP (approved):
A nitpick but makes sense to me, thanks again @dmlloyd 👍

wasm/src/main/java/com/dylibso/chicory/wasm/types/NameSection.java
dmlloyd: Fixed!

dmlloyd:
It may seem nitpicky, but it's all leading up to something :)

andreaTP (approved):

PR #197, opened 2024-01-25
Passing start.wast

### LINKED ISSUE #127: Load data segments in the imported memory

andreaTP:
Moved the initialization of the memory to fix #127 as well since I was at it.

### DIFF
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/Instance.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/Instance.java
@@ -1,6 +1,10 @@
 package com.dylibso.chicory.runtime;
 
+import static com.dylibso.chicory.runtime.Module.START_FUNCTION_NAME;
+
 import com.dylibso.chicory.runtime.exceptions.WASMMachineException;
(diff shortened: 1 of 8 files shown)


runtime/src/main/java/com/dylibso/chicory/runtime/Instance.java
```diff
@@ -47,17 +52,32 @@ public Instance(
...
         this.functionTypes = functionTypes.clone();
         this.imports = imports;
         this.machine = new Machine(this);
         this.tables = tables.clone();
         this.elements = elements.clone();
+
+        initialize();
+    }
+
+    private void initialize() {
```
andreaTP, line 67: @danielperano this should enable you to implement the `GLOBAL_GET` constant expression needed by `data.wast`

bhelx (approved):

PR #198, opened 2024-01-25
Remove size property from sections

dmlloyd:
The size property is unused in section objects. Requiring it to be specified does not pose any inconvenience for reading, but it would make it very difficult to construct sections for writing.

### DIFF
--- a/wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java
+++ b/wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java
@@ -180,49 +180,49 @@ public final class Parser {
                         }
                     case SectionId.TYPE:
                         {
-                            var typeSection = parseTypeSection(buffer, sectionId, sectionSize);
+                            var typeSection = parseTypeSection(buffer, sectionId);
                             listener.onSection(typeSection);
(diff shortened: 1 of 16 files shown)


andreaTP:
I was thinking about this today after your prev PR.
If it doesn't hurt we can keep the additional argument with the hope that this library will, one day, become a useful foundation for reading and writing wasm Modules on the JVM.
That said, if you have a good argument like perf or similar I'm open to learning more.

dmlloyd:
We need to remove it if we want to be able to support writing. The reason is that you will not know the size until all elements have been added, and their indices are known... i.e. it's a late computation which can change if and when we support adding elements to sections.

andreaTP (approved):
Oh, ok, makes sense! In this case lgtm!

andreaTP:
I leave the last word here to @bhelx just in case he left the sizes around for a reason.

bhelx:
Yes, this makes sense. We need to know the size mostly for skipping sections, but it doesn't need to be stored on the Section. Also, I'd love to start on a wasm encoder. It's gonna be key to building other tooling and I just put it off for the time being.

PR #199, opened 2024-01-26
Fixes to `Table` model class

dmlloyd:
Make the limits immutable. Use the maximum limit constant `0x1_0000_0000` when the upper limit is unspecified, and simplify comparisons accordingly. There is a class which represents limits; change `Table` to reuse it.

### DIFF
--- a/wasm/src/main/java/com/dylibso/chicory/wasm/types/Limits.java
+++ b/wasm/src/main/java/com/dylibso/chicory/wasm/types/Limits.java
@@ -1,19 +1,31 @@
 package com.dylibso.chicory.wasm.types;
 
 public class Limits {
-    private final int min;
-    private final int max;
+    public static final long LIMIT_MAX = 0x1_0000_0000L;
 
-    public Limits(int min, int max) {
-        this.min = min;
-        this.max = max;
+    private static final Limits UNBOUNDED = new Limits(0);
+
+    private final long min;
+    private final long max;
+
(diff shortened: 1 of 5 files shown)


wasm/src/main/java/com/dylibso/chicory/wasm/types/Limits.java
```diff
@@ -1,19 +1,31 @@
 package com.dylibso.chicory.wasm.types;
 
 public class Limits {
-    private final int min;
-    private final int max;
+    public static final long LIMIT_MAX = 0x1_0000_0000L;
```
andreaTP, line 4: I was wondering where this number comes from and I found [this thread](https://github.com/WebAssembly/spec/issues/607). Do you have a better link?
According to the [follow up PR](https://github.com/WebAssembly/spec/pull/873/files#diff-73846d0b7899a94fe050d3c0ba447da607bebedf80f8b5ed31cb4d164756f048R997-R1014) they are defining different limits for different things.
Should we specialize `Limits` for each usage?

Interested in hearing your reasoning around this 🙏

dmlloyd: I think specializing limits might be a good idea, but I don't know if specializing it by type is necessarily the right thing, if we think of "limits" as a data structure.

We could possibly pursue that in a follow-up PR.

PR #200, opened 2024-01-26
Add a utility method to create a new empty module

dmlloyd:


dmlloyd:
This one has a dependency on a future PR that I didn't notice, sorry for the noise.

PR #201, opened 2024-01-26
Make section IDs fixed by section

dmlloyd:
The sections know their ID, so having it be variable only introduces opportunities for problems. Always use the fixed ID instead.


andreaTP (approved):
LGTM, thanks for the cleanup again!

andreaTP:
The CI is acting up, I'll try to restart it as soon as it becomes possible.

PR #202, opened 2024-01-26
Control mutability of sections

dmlloyd:
Control mutability of sections. For now they are unmodifiable, but prepare for limited mutability (adding only) to support writing WASM files from the in-memory model.

### DIFF
--- a/wasm/src/main/java/com/dylibso/chicory/wasm/types/CodeSection.java
+++ b/wasm/src/main/java/com/dylibso/chicory/wasm/types/CodeSection.java
@@ -1,14 +1,25 @@
 package com.dylibso.chicory.wasm.types;
 
+import java.util.ArrayList;
+import java.util.List;
+
 public class CodeSection extends Section {
-    private FunctionBody[] functionBodies;
+    private final ArrayList<FunctionBody> functionBodies;
 
     public CodeSection(FunctionBody[] functionBodies) {
         super(SectionId.CODE);
-        this.functionBodies = functionBodies;
+        this.functionBodies = new ArrayList<>(List.of(functionBodies));
     }
 
     public FunctionBody[] functionBodies() {
-        return functionBodies;
+        return functionBodies.toArray(FunctionBody[]::new);
+    }
+
+    public int functionBodyCount() {
+        return functionBodies.size();
+    }
+
+    public FunctionBody getFunctionBody(int idx) {
+        return functionBodies.get(idx);
     }
 }
--- a/wasm/src/main/java/com/dylibso/chicory/wasm/types/ImportSection.java
+++ b/wasm/src/main/java/com/dylibso/chicory/wasm/types/ImportSection.java
@@ -1,14 +1,26 @@
 package com.dylibso.chicory.wasm.types;
 
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Stream;
+
 public class ImportSection extends Section {
-    private Import[] imports;
+    private final ArrayList<Import> imports;
 
     public ImportSection(Import[] imports) {
         super(SectionId.IMPORT);
-        this.imports = imports;
+        this.imports = new ArrayList<>(List.of(imports));
     }
 
-    public Import[] imports() {
-        return imports;
+    public int importCount() {
+        return imports.size();
+    }
+
+    public Import getImport(int idx) {
+        return imports.get(idx);
+    }
+
+    public Stream<Import> stream() {
+        return imports.stream();
     }
 }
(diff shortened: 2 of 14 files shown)


wasm/src/main/java/com/dylibso/chicory/wasm/types/CodeSection.java
```diff
@@ -1,14 +1,25 @@
 package com.dylibso.chicory.wasm.types;
 
+import java.util.ArrayList;
+import java.util.List;
+
 public class CodeSection extends Section {
-    private FunctionBody[] functionBodies;
+    private final ArrayList<FunctionBody> functionBodies;
```
andreaTP, line 7: are we going to pay a runtime penalty for not using a primitive array and doing the unboxing in the accessor function?

andreaTP (commented):
Thanks for those changes!
If this is not time-critical for you(e.g. blocking further development) I'll review this one early next week @dmlloyd , let me know 👍

wasm/src/main/java/com/dylibso/chicory/wasm/types/CodeSection.java
dmlloyd: No, it should be about equal. It also saves us from having to grow the array to add more items (once that support is added, which should be in the next PR).

wasm/src/main/java/com/dylibso/chicory/wasm/types/ImportSection.java
```diff
@@ -1,14 +1,26 @@
...
         super(SectionId.IMPORT);
-        this.imports = imports;
+        this.imports = new ArrayList<>(List.of(imports));
+    }
+
+    public int importCount() {
+        return imports.size();
+    }
+
+    public Import import_(int idx) {
```
andreaTP, line 19: something like `importNr` ?
The trailing `_` looks like generated code 😅

andreaTP (approved):
A minor nitpick but LGTM overall, thanks for wrapping those things up and give more consistency to the API!

wasm/src/main/java/com/dylibso/chicory/wasm/types/ImportSection.java
dmlloyd: It's always tricky when you have a conflict with a Java keyword. I've used the trailing `_` convention for a long time, but I'd be open to something else, as long as it would be something that would make sense for any name (i.e. not situation-specific - in this case the method happens to take an `int`-index argument but choosing a suffix on this basis means that a solution for this situation is not applicable to other situations).

andreaTP: Naming is hard 😅 I don't think we will ever find a "general" solution that doesn't look like generated code.

Since this code is handwritten I would be willing to keep underscores as a last resource, what do you think about `importIdx`? Similar signatures are already used in the codebase and it will be nice to streamline more.

I don't want to drag this PR long for this, nitpicky detail, but it would be great to find a good naming, even if it's situation-specific, in my opinion. wdyt?

dmlloyd: Should we then rename all of the other section getters to `xxxIdx` as well?

dmlloyd: I don't like it, honestly; it doesn't read well. To read properly it would have to be something like `importForIndex` or something like that, but now it's a little strange because `import` is a verb so it sounds like something is supposed to happen when you call it. And, it's a bit wordy. What about using `getXxx` uniformly when there is an integer index, if `import_` is unacceptable?

andreaTP: Make sense, I don't have super strong opinions here, I let someone else chip in, lgtm either way, thanks for trying things out! :+1:

andreaTP (approved):

wasm/src/main/java/com/dylibso/chicory/wasm/types/ImportSection.java
andreaTP: I don't like to keep this PR floating around for such a small detail, @bhelx and @danielperano can chip into this conversation later on and I'll take the action of fixing on top when necessary.
@dmlloyd sorry for making you wait on this one, if you can resolve the conflict it's ready to merge for me.

dmlloyd:
Resolved conflicts and renamed the indexed getters per discussion.

andreaTP (approved):
LGTM, thanks for making the changes!

PR #203, opened 2024-01-26
Add a utility method to create a new empty module

dmlloyd:


dmlloyd:
Looks like I've prematurely pushed this one again...

andreaTP:
No worries, let me convert this to draft.

dmlloyd:
This is ready now. But, I'm not sure it's really important anymore given the changes to how sections are parsed.

andreaTP (approved):
It's fine by me, you know that I would lock down the mutability a bit more in the future, but this is ok with the current tradeoffs and understanding.

dmlloyd:
I'm going to close this one. I thought it might be useful for parsing, but in the end it turns out it's better to create the sections as we go than it is to pre-create them and populate them after the fact. I suspect that it will be the same for other use cases as well.

PR #204, opened 2024-01-26
Allow custom parsers to be plugged in

dmlloyd:


andreaTP (approved):
Simple, elegant and straight to the point, love this change! Thanks again!

PR #206, opened 2024-01-29
fix numberOfValuesToReturn

mariofusco:


andreaTP:
Discussed sync and there are a few more bits to be aligned, but on a good track! 🎉

runtime/pom.xml
```diff
@@ -112,7 +114,9 @@
             SpecV1AddressTest.test203, SpecV1AddressTest.test211, SpecV1AddressTest.test212,
             SpecV1AddressTest.test213, SpecV1AddressTest.test214, SpecV1AddressTest.test217,
             SpecV1AddressTest.test233, SpecV1AddressTest.test234, SpecV1AddressTest.test236,
-            SpecV1AddressTest.test252, SpecV1AddressTest.test253, SpecV1AddressTest.test255
+            SpecV1AddressTest.test252, SpecV1AddressTest.test253, SpecV1AddressTest.test255,
+            SpecV1SwitchTest.test16, SpecV1SwitchTest.test17, SpecV1SwitchTest.test20,
```
andreaTP, line 118: Mario will look into those remaining tests in an upcoming PR.

andreaTP (approved):
We discussed this PR in detail sync, I agree with the changes that are improving in terms of correctness of the control flow.

andreaTP:
Thanks @mariofusco for this fix! 🎉 🎉 🎉

PR #207, opened 2024-01-30
Add controlled mutability to section definitions

dmlloyd:
Allow section definitions to be programmatically created and populated. Deletion and random insertion are not supported, because section indices are expected to remain stable and are referenced from other sections.


dmlloyd:
> Thanks for sending this one!
> 
> You are going in the direction of being able to populate and write Wasm Modules in Java and I really like it 👍 Are you planning on using this module in `qbicc` or is it still a stretch?

After reviewing Chicory pretty thoroughly, my tentative plan is to donate most of my WASM-related things to this project (with various changes to match up with the design of Chicory - some small and some significant) and drop all of it from qbicc, just consuming Chicory instead.

I can go into more detail on this week's call if you like.

> I feel a bit uncomfortable having methods and constructors that are not used nor tested in any other part of the codebase. I see a few ways how we can proceed, and I'm open to alternative options:
> 
> * add a simple test to cover, at least, the basic functionality and invoking those new methods

I think I can probably come up with something. Maybe it won't be 100% (yet) but at least the basics.

> * spike an integration test that will demonstrate how you envision those things working together at the end of the process(e.g. create a module and write it), it can be a bit hacky at first and we can refine with iterations

What I'd like to do eventually (at a minimum) is what I did in qbicc, which is to load a model from a file, then write the model back out to another file, and then load it in again and make sure it still works and/or compare to the original. Maybe across the whole test suite? I'm not 100% sure how to approach this yet though.

> * show, in a test, how those things can be used all together for some practical purpose, but I struggle to come up with something "simple enough"

Right, it's all leading up to being able to write WASM modules to disk, but there are a lot of pieces involved in doing that. Locally I'm maybe 8 commits ahead and I'm still not quite there, though I'm getting close. I'm trying to do this as incrementally as possible to avoid review whiplash and to make sure each change is in line with the overall philosophy of the project, making changes as needed. So far things have been relatively simple...

andreaTP:
> After reviewing Chicory pretty thoroughly, my tentative plan is to donate most of my WASM-related things to this project (with various changes to match up with the design of Chicory - some small and some significant) and drop all of it from qbicc, just consuming Chicory instead.

Awesome ❤️ we already discussed this and are open to going with you in this direction!
On my side, I'll try to be responsive and make this happen.

> I can go into more detail on this week's call if you like.

This would be much appreciated 🙏  having a better context/picture will make things easier.

> though I'm getting close

👍 let me encourage you to open a PR, in draft, to show the progress, it would be really helpful to connect all the dots.

> I'm trying to do this as incrementally as possible to avoid review whiplash and to make sure each change is in line with the overall philosophy of the project, making changes as needed. So far things have been relatively simple...

This is great, thanks a lot for sticking with those principles!

dmlloyd:
> 👍 let me encourage you to open a PR, in draft, to show the progress, it would be really helpful to connect all the dots.

I've opened #208 to show my current direction (and what's more or less working).

dmlloyd:
Pushed up a slight change; now, the new mutable section constructors are used during parsing, and the array-based constructors are dropped. All of the existing parsing tests now test this functionality.

andreaTP (approved):
Good thinking I'm ok with this change 👍 
A possible improvement will be to have a Builder pattern and immutable final objects.

andreaTP (changes requested):
@dmlloyd I propose to build on top of #211 wdyt?
I'm pleased by the resulting public API.

dmlloyd:
I'll put comments on that PR about the specific implementation, but this is my impression.

This model doesn't really have a life of its own. You can't do anything with it, other than create a runtime model from it (or, eventually, write it to disk). And the API is already more or less a builder. By adding another layer of builder, now you have a builder to build a builder. This adds a bunch of copying cost and API overhead, but it doesn't really give any specific benefit in return that I can see.

That said, at the end of the day it's not my decision and if you think that there is a benefit to adding the builder layer that I don't realize, then I would say that the choice is yours and I can adapt either way.

andreaTP:
Just to make it clear, I don't think that the sections are "Builders", also they are evolving in a direction that allows mutability which is undesirable for the `runtime`.

But I might be biased as well, let's hear more opinions cc. @lburgazzoli @bhelx @danielperano @enebo

dmlloyd:
I don't see why the runtime would want to refer directly to these sections though. Especially once there is a nontrivial component model, the information in the model structures would certainly be copied into new runtime structures. For example, rather than a type and code section plus function names section plus function imports (and eventually debug section), you'd only have a list of functions which have a type and name on them. Rather than the data and data count sections, you'd want to have initialized memories and passive segments (active and declarative segments would be dropped after loading).

If you require the model in order to manipulate the run time programmatically, you will be forced to construct a model as if you were going to save it to disk, which is likely not the best abstraction for talking to a runtime which already has capabilities that the model does not have (for example, host functions). I just think it's a mismatch.

dmlloyd:
If #211 is based on this PR, could we merge this PR and then consider #211 separately?

andreaTP:
This PR effectively makes the Sections mutable, which I'm worried about merging without, at least, a second pair of eyes.
With #211 on top, I think that this aspect is under control.

Let me try to get another reviewer around quickly.

andreaTP:
If no one else will chip in, just agreed to sort this out by tomorrow with @lburgazzoli 's help.

andreaTP:
@dmlloyd I wholeheartedly appreciate the work you are doing in this repo and the direction you are bringing it to, I'm really grateful!
In this case we have a slightly different view and I don't feel 100% comfortable in deciding the direction by myself without seeking additional advice, for the greater good of the project.
I think that both this and/or #211 are going in a direction that enables you to move on.
I'll do my best to sort this out tomorrow.

> sections were already mutable

I believe this is a loose end of the current implementation(e.g. not done completely intentionally), since we are at it I'd love to bring the public API to a better place.

andreaTP:
Thanks again @dmlloyd ! And sorry for dragging this.

PR #208, opened 2024-01-30
WIP - bringing over code/concepts from qbicc

dmlloyd:
## WIP only

At the request of @andreaTP, I'm opening this PR showing the work I'm doing to bring the qbicc WASM code over to this project. All of this is subject to change, or subject to being dropped completely, if it doesn't work out. Feedback is also welcome (with the caveat that I'm not necessarily committed to anything found in this branch). I'm not married to any of this code; it's all open for debate and discussion.

Anyone interested should definitely review only one commit at a time, for sanity's sake.

### Goals

My use case goals with this code (and my future eforts), both when I wrote it in qbicc and as I port it here, and go beyond, are:

* Reading and writing models - I should be able to load and store a model with no functional changes, and (ideally) minimal structural changes (note that in qbicc, I couldn't avoid structural changes due to certain design choices that with this PR I can revisit)
* Programmatic code generation - The opcodes API I developed for qbicc makes it easy to programmatically generate even very complex WASM code; I have further improved these concepts here but this would be a radical change and so I'm still debating it
* Draft spec support - I already need support for threads/atomics, exceptions, and multi memory; this code provides some groundwork for all of these specs
* Integration testing - After generating some code, I would like to be able to load and run it (with any necessary support modules) for testing purposes, with some kind of seamless Java integration
* Supporting existing use cases - any additions should support the existing Chicory use cases (i.e. don't make something impossible that was previously possible, or hard that was previously easy)

My design goals are:

* Clean and usable API (both for users of the APIs and for internal consumption)
* Type safety, tight encapsulation - it should be hard to make mistakes; the abstractions should match the underlying specifications
* Future-proofing - while this code does already support several proposed specifications, it's also designed to make it easy to support more
* Memory efficiency - programs can get quite large (hundreds of thousands of ops expected), so ideally this will be handled as gracefully as possible


andreaTP:
Quick feedback that we are completely aligned wrt: the `design goals`.

I think we should discuss a little further the "Programmatic code generation" subject to be sure we are aligned and we go in the same direction.

The shared code makes it a good reference when reviewing other PRs, thanks a lot for sharing!

andreaTP:
We surely want to enable it! 😄

bhelx:
Just a quick note, I looked at the qbicc codebase when @evacchi first pointed it out to me. There are a lot of solid, well built pieces in the core wasm encoder and parser. It's been my hope that we team up and share the base level tooling for that layer. I think a lesson I've learned from the rust wasm community is there is a lot of value around sharing the parsing and encoding codebases: https://github.com/bytecodealliance/wasm-tools . So I'm very supportive of @dmlloyd 's ideas on this front.

PR #209, opened 2024-01-30
Make FunctionBody more final

enebo:
In my last PR I recognized I made localTypes as a final primitive array and thought it looked a bit weird that instructions were also not a final primitive array.  Unless there is a need to mutably update instructions this feels like it shows intent that this is a set in stone definition as an artifact of parsing.

I also moved the unused ast into Ast and renamed it.  The benefit of this new definition is it is uncoupled from FunctionBody and it obviously already needs to know what Instructions are.  This may just be dead code though?

Not sure if this is needed but I figured I may as well tie up what looked like a loose end to me.

### DIFF
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/Machine.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/Machine.java
@@ -107,10 +107,7 @@ class Machine {
     }
 
     static void eval(
-            MStack stack,
-            Instance instance,
-            ArrayDeque<StackFrame> callStack,
-            List<Instruction> code)
+            MStack stack, Instance instance, ArrayDeque<StackFrame> callStack, Instruction[] code)
             throws ChicoryException {
 
         try {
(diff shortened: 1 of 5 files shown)


dmlloyd:
Since arrays are mutable (and you therefore must copy them defensively to use them safely), does this improve immutability?

enebo:
@dmlloyd It is merely in reading intent.  The old code was an ArrayList and I can add elements to that without even doing an arraycopy.  I do want something fully immutable so that would be preferable.  If someone plan on changing this they should make a new FunctionBody right?  Do you expect people would be replacing the array while someone is reading this?  Perhaps that is the disconnect.

I am ok closing this though.  It is fairly clear you do not like the original PR either for similar reasons so it should probably be done differently.

dmlloyd:
Yeah the existing `FunctionBody` code is not ideal, I agree. Maybe this kind of change is OK for the short term, it just happens to conflict with one of my bigger upcoming changes which introduces a first-class instruction list type with validation and explicit mutability control. If you feel strongly that this is a worthwhile incremental improvement I can rebase and resolve the conflicts (I've been doing plenty of that lately). I'm probably a few weeks away from that commit being ready to merge so I would understand if you want to do that.

PR #210, opened 2024-01-30
Make FunctionBody fields immutable

enebo:
Both fields will end up using ImmutableCollections impls making it impossible to change the valueType and instruction list.


dmlloyd (approved):
I like this a lot better, thanks!

danielperano (approved):

andreaTP (approved):
LGTM, thanks @enebo !

PR #211, opened 2024-01-31
Section Builders

andreaTP:
Builds on top of #207 introduces the Builder pattern for the `Section`s and standardize the API across the codebase.

Forcing ppl to use the Builder pattern to construct the Sections makes the API more "future-proof" and, I think, improves the usability.


wasm/src/main/java/com/dylibso/chicory/wasm/types/CodeSection.java
```diff
@@ -1,14 +1,22 @@
...
+import java.util.Objects;
 
 public class CodeSection extends Section {
     private final ArrayList<FunctionBody> functionBodies;
 
-    public CodeSection(FunctionBody[] functionBodies) {
+    private CodeSection(ArrayList<FunctionBody> functionBodies) {
         super(SectionId.CODE);
-        this.functionBodies = new ArrayList<>(List.of(functionBodies));
+        this.functionBodies = functionBodies;
```
dmlloyd, line 11: If the builder is used after the `build` call, the lists of everything that was built with it will be modified. You'd need to make a copy here e.g. `List.copyOf(functionBodies)`.

wasm/src/main/java/com/dylibso/chicory/wasm/types/CustomSection.java
```diff
@@ -10,4 +10,10 @@ protected CustomSection() {
     }
 
     public abstract String name();
+
+    public interface Builder {
+        Builder withBytes(byte[] bytes);
```
dmlloyd, line 15: I think it would be more uniform if every section were parsed the same way. In this case the section builder takes bytes, but in other cases the parsing logic is stored in the `Parser`. What about having a common `Builder` method which parses that section, or making this uniform some other way?

wasm/src/main/java/com/dylibso/chicory/wasm/types/CodeSection.java
```diff
@@ -22,4 +30,28 @@ public int functionBodyCount() {
     public FunctionBody getFunctionBody(int idx) {
         return functionBodies.get(idx);
     }
+
+    public static final class Builder {
```
dmlloyd, line 34: It's somewhat surprising that the builders don't have a type hierarchy which mirrors their section type hierarchy. A common `Builder` supertype would be a good place to put a `parse` method, to make parsing more uniform.

wasm/src/main/java/com/dylibso/chicory/wasm/types/ExportSection.java
```diff
@@ -18,4 +26,27 @@ public int exportCount() {
...
+    public static final class Builder {
+        private final ArrayList<Export> exports;
+
+        private Builder() {
+            this.exports = new ArrayList<>();
+        }
+
+        private Builder(ExportSection exportSection) {
+            this.exports = new ArrayList<>();
+            this.exports.addAll(exportSection.exports);
```
dmlloyd, line 39:
```suggestion
            this.exports = new ArrayList<>(exportSection.exports);
```

dmlloyd (commented):
There are some technical things I would change, commented above. But see also [this comment](https://github.com/dylibso/chicory/pull/207#issuecomment-1919055729).

andreaTP:
All great comments @dmlloyd I'm willing to integrate everything you pointed out if we decide to move on with this.
Let's just wait for more feedback!

andreaTP:
Let's put this on hold as decided on #207

PR #213, opened 2024-01-31
Calculate the correct number of value to be dropped from the stack when jumping out of multiple blocks

mariofusco:


andreaTP (approved):
Much cleaner to read and to reason about, as it passes all of the control flow related specs it's good to merge.

Thanks again @mariofusco 🎉

PR #215, opened 2024-02-01
Name section updates

dmlloyd:
* Fix parsing so that it does not fail if function names do not come first
* Add support for all name types in the core and `extended-name-section` specifications
* Preserve index because it is significant (names can appear in any order)

### DIFF
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/StackFrame.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/StackFrame.java
@@ -70,7 +70,7 @@ public class StackFrame {
         var nameSec = instance.module().nameSection();
         var id = "[" + funcId + "]";
         if (nameSec != null) {
-            var funcName = nameSec.getFunctionName(funcId);
+            var funcName = nameSec.nameOfFunction(funcId);
             if (funcName != null) id = funcName + id;
         }
         return id + "\n\tpc=" + pc + " locals=" + Arrays.toString(locals);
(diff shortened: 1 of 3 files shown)


andreaTP:
Is it this spec, right?
https://webassembly.github.io/spec/core/appendix/custom.html#name-section

I checked in the [testsuite](https://github.com/WebAssembly/testsuite) but I cannot find any test stressing this aspect.

Would you mind adding a test that will, at least, cover the happy path and prevent us from breaking the functionality in the future?

andreaTP (changes requested):
Need a test to cover, at least, the happy path of the code introduced here.

dmlloyd:
The existing tests cover at least some of this: specifically function and global names.

We'll have to craft some kind of `wast` for the rest.

dmlloyd:
I've added test conditions to the existing test to verify that the name subsections which appear in the existing parser test file is fully parsed, and that the contents are correct where possible (validating the name of every function isn't feasible, and also isn't likely to remain stable across rebuilds of the test WASM file from Rust anyway AFAIK).

While it doesn't cover 100% of cases yet, all cases use the same two code paths so I would consider it unlikely that one kind of name would succeed where another fails. Additionally, this code is used in every existing test where a test file is loaded with a name section.

By any measure, the testing is more comprehensive now than before. But if there are additional test cases you'd like to see, please let me know.

andreaTP (approved):
Thanks for covering even the test part, LGTM!

PR #216, opened 2024-02-01
Change `Import`/`ImportDesc` to immutable typesafe hierarchy

dmlloyd:
At present, the specifics of imports are represented using the god-class `ImportDesc`, which can hold any kind of import in a type-unsafe manner. As the number of supported import types increase over time, this class will become increasingly unwieldy to use and validate, since some of the fields only apply to only some of the import types.

Drop `ImportDesc` in favor of a closed class hierarchy of possible import types, which are strongly typed and validated.

While we're here, also modify `MemoryImport` to use `MemoryLimits` instead of `Limits` to be consistent with other kinds of memory declarations.


andreaTP (approved):
Love this clean up! Thanks!

PR #218, opened 2024-02-01
Make `DataSegment` hierarchy (mostly) immutable

dmlloyd:
Make the base abstract and the leaves final. Use a list for holding the immutable instruction sequence.


andreaTP (approved):
Good to go, thanks for the improvement!

PR #220, opened 2024-02-01
Import/export improvements

dmlloyd:
* Make exports immutable
* Remove descriptor abstraction similarly to imports
* Unify `ExternalType` abstraction per spec
* Add external type for tags (exception spec)
* Add spec references for imports, exports, and external type


wasm/src/main/java/com/dylibso/chicory/wasm/types/ExternalType.java
```diff
@@ -0,0 +1,43 @@
...
+ * See <a href="https://webassembly.github.io/spec/core/syntax/types.html#external-types">External Types</a> for
+ * reference.
+ */
+public enum ExternalType {
+    // note: keep in order
+    FUNCTION(0x00),
+    TABLE(0x01),
+    MEMORY(0x02),
+    GLOBAL(0x03),
+    TAG(0x04),
```
andreaTP, line 17: What's a tag?
The spec mentions only 4 types:
https://webassembly.github.io/spec/core/syntax/types.html#external-types

func/table/mem/global
Do you have a reference?

andreaTP (changes requested):
I like the unification of ExternalTypes 👍 need a clarification re: "TAG"

wasm/src/main/java/com/dylibso/chicory/wasm/types/ExternalType.java
dmlloyd: Sure. https://github.com/WebAssembly/exception-handling/blob/main/proposals/exception-handling/Exceptions.md#external_kind

andreaTP: Thanks! Do you mind to add a comment with the link, otherwise LGTM

andreaTP (approved):
Lgtm, thanks!

PR #222, opened 2024-02-01
Make `Memory` and `MemoryLimits` final and immutable

dmlloyd:
* Add in documentation and spec references
* Add in `shared` flag support from threads spec
* Add `null`-checking

### DIFF
--- a/wasm/src/main/java/com/dylibso/chicory/wasm/types/MemoryLimits.java
+++ b/wasm/src/main/java/com/dylibso/chicory/wasm/types/MemoryLimits.java
@@ -1,11 +1,21 @@
 package com.dylibso.chicory.wasm.types;
 
-public class MemoryLimits {
+/**
+ * Limits for memory sizes, in pages.
+ * Memory limits also define whether the corresponding memory is <em>shared</em>.
+ * <p>
+ * See <a href="https://webassembly.github.io/spec/core/syntax/types.html#syntax-limits">Limits</a>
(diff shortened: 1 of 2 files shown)


wasm/src/main/java/com/dylibso/chicory/wasm/types/MemoryLimits.java
```diff
@@ -44,14 +81,27 @@ public static MemoryLimits defaultLimits() {
...
+     * {@return the maximum size, in pages}
+     */
     public int maximumPages() {
         return maximum;
     }
 
+    /**
+     * {@return <code>true</code> if the limits apply to a shared memory segment, or <code>false</code> otherwise}
+     */
+    public boolean shared() {
```
andreaTP, line 101: I'm failing to find a reference to this field in the spec:
https://webassembly.github.io/spec/core/search.html?q=shared
Do you have a link?

dmlloyd: Sure. https://github.com/WebAssembly/threads/blob/main/proposals/threads/Overview.md#spec-changes

andreaTP: Oh, this is coming from the threads proposal, I'm fine with it, but please add this link in a comment, otherwise LGTM.

dmlloyd: Done.

andreaTP (approved):
Lgtm, thanks!

PR #226, opened 2024-02-05
Demo the usage of Javy for wasm modules compiled from JS

andreaTP:
I'm testing the usage of alternative languages, here is Javascript with Javy.
Please note that the changes to "adapt" the extracted function arguments are the same that enabled [running Doom](https://github.com/dylibso/chicory/commit/a8c897558fdd73a99d43ca9bf1c0e54016fe5c03).

Unfortunately, I cannot find a place in the Spec where this behavior is defined.


runtime/src/main/java/com/dylibso/chicory/runtime/Machine.java
```diff
@@ -2368,7 +2368,7 @@ static Value[] extractArgsForParams(MStack stack, ValueType[] params) {
             var p = stack.pop();
             var t = params[i - 1];
             if (p.type() != t) {
-                throw new RuntimeException("Type error when extracting args.");
+                p = p.as(t);
```
bhelx, line 2371: Interesting. I've run into this before as well with my extism branch but I thought it had something to do with something i messed up in a host function. It seems wrong given that Wasm is supposed to be typesafe and there are explicit instructions for doing this

andreaTP: I noticed `I32` passed in place of `I64`, but we would need to minimize it to debug ...

andreaTP: Let me try to go through, the relevant spec is [this](https://webassembly.github.io/spec/core/bikeshed/index.html#invocation%E2%91%A1):

image

Where valid is defined [here](https://webassembly.github.io/spec/core/bikeshed/index.html#valid-val), **if** I read it correctly(and might not), [here](https://webassembly.github.io/spec/core/bikeshed/index.html#numeric-values--threfsyntax-instr-numericmathsfconstc) they define that a numeric type is valid against other numeric types.

But I would love someone else to have a look(moving to draft until we understand this detail).

danielperano: I read the spec differently - I read 
```
4.5.2.1. [Numeric Values](https://webassembly.github.io/spec/core/bikeshed/index.html#syntax-val) t.[const](https://webassembly.github.io/spec/core/bikeshed/index.html#syntax-instr-numeric) c

    The value is valid with [number type](https://webassembly.github.io/spec/core/bikeshed/index.html#syntax-numtype) t.
```
as "every value with type `t` is a valid `t`". I'm not familiar with any place where the spec permits automatic type widening (or narrowing).

danielperano: One of the things I find annoying about the wasm spec is the number of cases where it's formal to the point of being confusing - like the above case is certainly formal and precise, but (assuming I'm reading it correctly) is phrased in a way that's much easier to confuse than some alternate, albeit less formal/precise, ways of stating the same.

andreaTP: @danielperano and @bhelx I have asked to @mariofusco fo have a look into this, he was not finding anything strange, so we resorted to look at other implementations:

- [Wazero](https://github.com/tetratelabs/wazero/blob/36676928d22ab92c34299eb0dca7608c92c94b22/internal/wasm/gofunc.go#L109) takes the raw part of the data and adapts it to the required type (similar to what I proposed in this PR)
- [Wasm3](https://github.com/wasm3/wasm3/blob/139076a98b8321b67f850a844f558b5e91b5ac83/source/m3_env.c#L947) does an explicit cast

Reading again the spec, we should probably read it as:
any Numeric value is **valid** against any other Numeric value and the necessary widening/narrowing applies.

danielperano: Well, in that case we're outvoted so I think we should go with the majority opinion that the correct handling is a type conversion.

danielperano: Even if it were true that the wasm spec authors intended it to be as we were reading it, it would seem that converting instead of trapping has become the defacto standard among multiple runtimes so if any compilers depend on that behavior then in my opinion we're taking on a needless incompatibility risk by trapping instead of converting.

andreaTP: I'll add a stricter check to apply the conversion only to numeric types tomorrow.

bhelx: Sounds good. I think as long as we add some logic to make sure the cast is safe it shouldn't be a problem

andreaTP: I added the logic to check that it's a "Numeric" value.

andreaTP:
ready for final review 👍 thanks for everyone that starred with me at the spec because of this 😅

bhelx (approved):

mariofusco (approved):

danielperano (approved):

andreaTP:
🎉

PR #229, opened 2024-02-07
docs: update readme with 0.0.7 release reference

nilslice:
I believe 0.0.7 has been released out to package managers, so this just brings the readme installation instructions up to date


andreaTP (approved):
Thanks! We should update the Readme automatically on release ...

PR #230, opened 2024-02-07
create new control flow for each function + remove stack unwinding

mariofusco:


andreaTP (approved):
LGTM!

I'm really surprised by the removal of the unwinding functions, but if all the tests are passing means we are going in the right direction.
If needed we can add it back later on with a proper test for it.

PR #236, opened 2024-02-08
Move compilation support image to wasm-corpus

### LINKED ISSUE #171: Reproducible compiled wasm
When merging the WASI support we struggled with the compatibility and reproducibility of the compiled `wasm` files.

Problems:
- the container image is for X86 and seems to have issues on ARM
- the `wasi` compiled modules are generated without the `_start` export

bhelx:
Closes #171 

This centralizes all of our guest source code into a maven submodule `wasm-corpus`. These files get included as a testresource at the pom level in the submodules that need them. I've slightly changed the way we build these. There is a different interface and it now mounts the directory instead of passing bytes across the boundary. `run.sh` simply dockerizes calls to a `compile.sh` script which can run inside the docker guest or on your host machine so it simplifies things a bit.

```bash
cd wasm-corpus

# rebuild (or build image for first time)
./run.sh rebuild

# compile all the submodules
./run.sh 

# compile just one folder
./run.sh rust
```

If you want to run on your host machine you can run the subscript yourself. It has the same arguments (assumes you have `WASI_SDK_PATH` set for compiling c:

```bash
cd wasm-corpus/src/test/resources

# example
./compile.sh rust
```


andreaTP:
Skimmed on the phone and looks like a great improvement!
I'll look at it closely tomorrow morning.

wasm-corpus/run.sh
```diff
@@ -0,0 +1,12 @@
...
+
+# run with `rebuild` to build the image
+# run with no args to compile everything
+
+if [[ "$1" = "rebuild" ]]; then
+  docker build . --platform linux/amd64 -t chicory/wasm-corpus
+else
+  # Optionally takes the args `lang` (ex: wat, rust) and `file` (ex: br.wat)
+  # both default to all
+  docker run --platform linux/amd64 -v ./src/test/resources:/usr/code --rm chicory/wasm-corpus $1 $2
```
andreaTP, line 11: I have used this mechanism in the past and, on some systems, the file permissions are going to be messed up, still, we can fix this later on.

andreaTP (approved):
LGTM, I love the direction!

I made a note on the generated files permissions, but we can keep it as an improvement.
Another idea we can explore is to rely on the `wasm-corpus` artifact as a straight dependency for the `test` scope instead, but I'll need to evaluate the tradeoff.

Feel free to merge!

PR #237, opened 2024-02-09
Fix FunctionType#toString implementation

illarionov:
Fixed copy-paste typo in the `FunctionType#toString` implementation. Added a test


bhelx:
Thank you! Good catch

PR #239, opened 2024-02-09
Fuzzying

andreaTP:
Opening as wip to encourage early reviews.
This is an initial fuzz-testing infrastructure:

- produce a CLI of Chicory to be able to compare results
- generate WASM modules using `wasm-smith`
- invoke the exported functions on `wasmtime` (aka the Oracle), if `wasmtime` fails we skip the test
- Invoke the same exported functions on Chicory
- when the output from the 2 CLIs differs a new folder is created under `fuzz/src/test/resources/` with a prefix `crash-`
- there is a parametrized test running on all of the reproducers
- there is a configurable test to provide a seed and the types and run the same infrastructure (it seems like `wasm-smith` is producing different results on different machines)
- a GH action to run those tests nightly, it will open PRs on failure, similar to: https://github.com/andreaTP/chicory/pull/6

All in all, this implementation is the first bulk of work and there are some quick and dirty things to get it running.
We should iterate on it as we grow expertise in it.
Happy to hear what you think about it!


.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,55 @@
...
+        uses: bytecodealliance/actions/wasmtime/setup@v1
+        with:
+          version: "v17.0.0"
+      - name: Setup `wasm-tools`
+        uses: bytecodealliance/actions/wasm-tools/setup@v1
+        with:
+          version: "1.0.57"
+      - name: Fuzz Test
+        # TODO: change before merging!
+        if: github.repository_owner == 'andreaTP'
```
andreaTP, line 34: if you want to try it on your personal fork, you need to provide a `GH_TOKEN` and change this line.

bhelx:
Great start so far! First question I have is about the CLI. it's a very useful tool to have regardless, but do you know the overhead with starting up the jvm and jitting all the code, etc? Could make a big difference with fuzzing time. Would it be possible to compile it to native code?

fuzz/src/test/java/com/dylibso/chicory/fuzz/TestModule.java
```diff
@@ -0,0 +1,111 @@
...
+    WasmTimeWrapper wasmtime = new WasmTimeWrapper();
+    ChicoryCliWrapper chicoryCli = new ChicoryCliWrapper();
+
+    public List<String> paramsList(FunctionType type) {
+        return Arrays.stream(type.params()).map(p -> randomNumber()).collect(Collectors.toList());
+    }
+
+    public String randomNumber() {
+        // TODO: 2 digits integer seems not enough, but a starting point ...
+        return RandomStringUtils.randomNumeric(2);
```
bhelx, line 30: This is a good start and not worth worrying too much about right now, but i think generally you don't want a uniform random distribution. There are some values which are more likely to break things than others. You want to try those first. There are probably some libraries that can help us define these. It's probably worth asking an expert for guidance on this part.

bhelx: we may also need to "reduce" inputs after a failure. minimizing a broken example down to the smallest failure case. again i assume there are some libraries in java that can help with this.

bhelx: maybe a library like this could help https://github.com/pholser/junit-quickcheck

dmlloyd (commented):
I love the idea of fuzzing. I think it will be a great way to catch problems. I don't have a lot of commentary on the change itself because I'm not all that familiar with the testing infrastructure. But I like the overall approah.

andreaTP:
> First question I have is about the CLI.

Correct, I was lazy enough to only bundle everything up in a jar and make it executable 😅 .
Building a native binary is definitely possible but it requires a little more work, the biggest challenge is that cross-compilation is not supported and you need separate jobs to release for every possible target architecture ( [e.g.](https://github.com/andreaTP/kc-quarkus-launcher-poc/blob/main/.github/workflows/release-binaries.yaml) ).

I think we can discuss the native CLI subject as an improvement on top of this PR as it also brings the questions about frameworks etc. etc. (can we use Quarkus? should we strive to stay away from any framework? etc. etc.)

fuzz/src/test/java/com/dylibso/chicory/fuzz/TestModule.java
andreaTP: Awesome idea!
I'm not 100% sure how to properly integrate it, and I think we should gain some experience with the tools themselves but surely something we can explore at some point.

andreaTP:
I think this is ready for review, the CI failure is legit and shows an actual mismatched result.
In case this is too much for a single PR I can break it down to pieces, but I would love to build some experience with this setup before making additional changes.

bhelx:
I’m out until Wednesday. I can check it out then. But if someone else has bandwidth to review go for it!

bhelx (approved):
Okay looks like a great start! I think we can iterate on improvements. Nice to have all this setup!

PR #240, opened 2024-02-10
Fix functionTypes indexes

illarionov:
In recent commits indexing of the  [Module#functionTypes](https://github.com/dylibso/chicory/blob/0e523565a2d42164b03783af0c28710be1bb5edc/runtime/src/main/java/com/dylibso/chicory/runtime/Module.java#L157) array has changed several times from `funcIdx` to `importId` and back. This resulted in functions being imported and called incorrectly if the WASM module imports external memory in addition to functions.

This pull request fixes the import of host functions.

I also added a test covering this case (Wasm code that fails if the functions are not imported correctly).


bhelx:
Thanks again @illarionov! And we appreciate the test. Perhaps we haven't gotten to the relevant wasm testsuite tests for this yet, or we have a blind spot of some sort. I'll investigate when I get back next week.

andreaTP (approved):
Looks like a good bug fix!
Thanks! Really incredible that we missed it so far! And I see some cleanup too :+1:

PR #241, opened 2024-02-15
Demonstrate tinygo guest support

bhelx:
Working on a larger corpus of modules for running tests so I'm testing out different languages. Expect to see some more (including official google go toolchain) soon.


wasm-corpus/run.sh
```diff
@@ -8,5 +8,5 @@ if [[ "$1" = "rebuild" ]]; then
 else
   # Optionally takes the args `lang` (ex: wat, rust) and `file` (ex: br.wat)
   # both default to all
-  docker run --platform linux/amd64 -v ./src/test/resources:/usr/code --rm chicory/wasm-corpus $1 $2
+  docker run --platform linux/amd64 -v $(pwd)/src/test/resources:/usr/code --rm chicory/wasm-corpus $1 $2
```
andreaTP, line 11: Instead of `pwd` which depends on "from where" the script is invoked, I started relying on "where the script is placed" using this at the start:

https://github.com/dylibso/chicory/blob/ddb11ea73a99ffb5b290afe5a70145193191e78c/scripts/compile-resources.sh#L5

andreaTP (approved):
Looks great! I haven't realized we can run Go already! 🎉 

I left a nitpick on using relative paths, but we can fix it later.

Also, I'm wondering why this is changed, is it not reproducible across machines?
Screenshot 2024-02-15 at 09 35 20

wasm-corpus/run.sh
bhelx: Thanks i'll follow up with some improvements in the next PR!

PR #242, opened 2024-02-15
Include most of the remaining wast specs

andreaTP:
After this PR(and excluding SIMD) we have 2 outstanding wast files not included:

- `elem.wast` -> the test requires sharing a table across modules (getting an exported instance and injecting it later on)
- `linking.wast` -> need to have a closer look ...

### DIFF
--- a/wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java
+++ b/wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java
@@ -252,7 +252,11 @@ public final class Parser {
                         {
                             // "Skipping Section with ID due to configuration: " + sectionId
                             listener.onSection(new Section(sectionId));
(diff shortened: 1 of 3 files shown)


wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java
```diff
@@ -252,7 +252,11 @@ void parse(InputStream in, ParserListener listener) {
                         {
                             // "Skipping Section with ID due to configuration: " + sectionId
                             listener.onSection(new Section(sectionId));
-                            buffer.position((int) (buffer.position() + sectionSize));
+                            buffer.position(
```
andreaTP, line 255: FYI: @dmlloyd this was failing only in CI on Windows

PR #244, opened 2024-02-16
Rework types (option 3)

dmlloyd:
* Create constants for each type ID
* Use switch statements for finding types by ID to avoid boxing overheads
* Add query methods for type characteristics
* Replace `ElementType` with `ValueType`


wasm/src/main/java/com/dylibso/chicory/wasm/types/ValueType.java
```diff
@@ -1,45 +1,198 @@
...
     }
 
-    public static ValueType byId(long id) {
-        return byId.get(id);
+    /**
+     * {@return the <code>ValueType</code> for the given ID value}
+     *
+     * @throws IllegalArgumentException if the ID value does not correspond to a valid value type
+     */
+    public static ValueType forIdOrThrow(int id) {
```
andreaTP, line 147: I think that this can be the only default implementation, I would call it `forId` and make the current `forId` `private` or simply remove it, wdyt?

wasm/src/main/java/com/dylibso/chicory/wasm/types/ValueType.java
```diff
@@ -1,45 +1,198 @@
...
+                return null;
         }
     }
+
+    /**
+     * {@return the reference-typed <code>ValueType</code> for the given ID value}
+     *
+     * @throws IllegalArgumentException if the ID value does not correspond to a valid reference type
+     */
+    public static ValueType refTypeForIdOrThrow(int id) {
```
andreaTP, line 174: What about keeping only this as the default implementation for `refTypeForId` ?

andreaTP (changes requested):
I like the changes done in this direction, I would remove the `OrThrow` suffix and make the throwing methods the only and default implementation.
Since we are consolidating the API I think we can go down this path 🙂

dmlloyd:
Applied suggestions to not have a `null`-returning option for the get-by-ID methods.

andreaTP (approved):
This looks good to me, but I would like to have a second opinion before merging if it doesn't take too long.

cc. @bhelx @danielperano @mariofusco

dmlloyd:
Looks like there's one problem due to the non-null returning option, but the fix should be simple.

andreaTP:
(and CI should be green 😄 )

dmlloyd:
All fixed :+1:

mariofusco (approved):

PR #245, opened 2024-02-16
Rename `BlockType` to `BlockKind`

dmlloyd:
Blocks already have a "type", which is expressed as a function type (input and output type lists). So, disambiguate this terminology.


andreaTP (approved):
LGTM, thanks!

PR #247, opened 2024-02-19
Function types cleanup

dmlloyd:
* Make function types immutable and list-based, with caching of common types.
* Introduce equals/hashCode for function types.

### DIFF
--- a/wasm/src/main/java/com/dylibso/chicory/wasm/types/FunctionType.java
+++ b/wasm/src/main/java/com/dylibso/chicory/wasm/types/FunctionType.java
@@ -1,57 +1,132 @@
 package com.dylibso.chicory.wasm.types;
 
-import java.util.Arrays;
+import java.util.EnumMap;
+import java.util.List;
+import java.util.Objects;
 
-public class FunctionType {
-    private ValueType[] params;
-    private ValueType[] returns;
+public final class FunctionType {
+    private final List<ValueType> params;
+    private final List<ValueType> returns;
+    private final int hashCode;
 
-    public FunctionType(ValueType[] params, ValueType[] returns) {
+    private FunctionType(List<ValueType> params, List<ValueType> returns) {
         this.params = params;
         this.returns = returns;
+        hashCode = Objects.hash(params, returns);
     }
 
-    public ValueType[] params() {
+    public List<ValueType> params() {
         return params;
     }
 
-    public ValueType[] returns() {
+    public List<ValueType> returns() {
         return returns;
     }
 
     public boolean paramsMatch(FunctionType other) {
-        return Arrays.equals(params, other.params);
+        return params.equals(other.params);
     }
 
     public boolean returnsMatch(FunctionType other) {
-        return Arrays.equals(returns, other.returns);
+        return returns.equals(other.returns);
+    }
+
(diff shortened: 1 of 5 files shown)


wasm/src/main/java/com/dylibso/chicory/wasm/types/FunctionType.java
```diff
@@ -1,57 +1,132 @@
...
+
+    public boolean equals(FunctionType other) {
+        return hashCode == other.hashCode && paramsMatch(other) && returnsMatch(other);
+    }
+
+    public int hashCode() {
+        return hashCode;
+    }
+
+    private static final EnumMap<ValueType, FunctionType> returning;
```
andreaTP, line 46: IIRC `EnumMap` has some idiosyncrasies.

Any specific reason to keep those functions statically in a Map instead of using a mapping function?

```java
private static getAccepting(ValueType type) {
  return new FunctionType(List.of(), List.of(type));
}
```

andreaTP (changes requested):
A question regarding the usage of `EnumMap`, but the direction is good 👍

wasm/src/main/java/com/dylibso/chicory/wasm/types/FunctionType.java
dmlloyd: Using an enum map (privately) allows for constant-time lookup, so as a data structure, it's often a good choice (when it is applicable). It's basically equivalent (but more readable) compared to a `List<FunctionType>` indexed by `ValueType#ordinal`.

The reason to use a map instead of creating a new instance each time is because these common function types can occur many times in a module, so having them be cached is bound to provide a savings in terms of allocation overhead as well as memory usage.

dmlloyd: As a general programming pattern, this doesn't scale well to large numbers of enums (at least, in terms of readability) so I would avoid this approach *unless* there was a compelling reason to expose the static fields directly. In general, using a `Map` is idiomatic for caches. I would only avoid maps for caching if (for example) the key is a primitive that would require boxing on every lookup, or if as I said earlier the field should be directly accessible for some reason, which IMO is not the case here. So I think that using a map (particularly an `EnumMap`, which is specifically designed for this job) is more correct than other approaches for this case.

If the initialization code is what is unreadable to you, I could instead use a streamwise initialization e.g. `EnumMap<ValueType, FunctionType> returning = Stream.of(values()).collect(Collectors.toMap(...));` if that would help. That would eliminate the `static` blocks. I only went with the `static` approach because it seemed more readable but of course this is subjective.

andreaTP: I proposed a different implementation as, in this case, doesn't need to scale any further than the already defined 7 cases.
I do buy your argument, but it looks a little odd applied in such a small and well-defined context.

The initialization is more readable in its current form than using streams IMHO.

andreaTP (approved):
I keep thinking that it would be best to avoid the `EnumMap`s here, but it's a well encapsulated technical detail.
In the spirit of keeping the ball rolling and accepting that we have a different bias on this detail LGTM in whatever form you prefer. Thanks!

PR #249, opened 2024-02-19
Introduce `WasmInputStream` for reading WASM files

dmlloyd:
* Parse all sections streamwise
* Do not keep any portion of the file in memory during parsing other than what is needed to parse the current element
* Ensure that all sections are parsed safely even if there is extraneous data


andreaTP (commented):
I like the direction, and the fact that the tests are passing is already great!
But I would like to review this carefully as it's a sprawling change, I'll(likely) try to go through it tomorrow.

wasm/src/main/java/com/dylibso/chicory/wasm/io/BufferWasmInputStream.java
```diff
@@ -0,0 +1,88 @@
...
+
+import java.nio.BufferUnderflowException;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+
+/**
+ * A WASM input stream backed by a byte buffer.
+ */
+class BufferWasmInputStream extends WasmInputStream {
+    private final ByteBuffer buffer;
```
andreaTP, line 11: Have you evaluated the option of wrapping something more high level like a `ByteArrayInputStream` ?
It will make the implementation more straightforward as we won't need to deal with indexes and more

wasm/src/main/java/com/dylibso/chicory/wasm/io/InputStreamWasmInputStream.java
```diff
@@ -0,0 +1,142 @@
...
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.VarHandle;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+
+/**
+ * A WASM input stream based on {@link InputStream}.
+ */
+class InputStreamWasmInputStream extends WasmInputStream {
+    private static final VarHandle le32 =
```
andreaTP, line 15: What would be the alternative to the usage of `VarHandle` / `MethodHandles` ?

Those classes are not going to be available on some JVM targets(i.e. Android) and I refrain from using them in such a low-level building block.

If I understand correctly an alternative would be to use `ByteBuffer` primitives, such as:

```java
ByteBuffer.wrap(byteArray)
     .order(ByteOrder.BIG_ENDIAN)
     .asIntBuffer(); // .asLongBuffer()
```

Unless you have a really compelling argument for moving ahead with this implementation I kindly ask you to avoid any usage of `java.lang.invoke`, especially in the low-level `wasm` sub-module.

andreaTP (changes requested):
Thanks for the nice step forward @dmlloyd !

Added a couple of comments on the concrete implementations of the `WasmInputStream` classes.

wasm/src/main/java/com/dylibso/chicory/wasm/io/BufferWasmInputStream.java
dmlloyd: This implementation is for when you already have a `ByteBuffer`. Using `ByteBuffer` here does not dictate anything outside of this class (note that the class is non-public). The use case is when you want to (for example) memory-map a very WASM large file, or when you have a WASM file within a buffer already (this will happen when we start talking about output).

dmlloyd: Note that you can also wrap an `InputStream` using a separate API and implementation.

wasm/src/main/java/com/dylibso/chicory/wasm/io/InputStreamWasmInputStream.java
dmlloyd: According to what I could find online, Android does in fact support `java.lang.invoke`, at least to some extent (I was not able to determine exactly what).

In any event, when it comes to Android I think we need to do one of two things starting immediately:

* Start testing with Android to ensure that it really is supported, or else
* Stop saying we will support Android

Reading online I found that Android does in fact support `j.l.i`... and, they also support Java 17 starting with Android 14. So it seems like we may be doing some acrobatics for no good reason.

Now, that having been said: the reason for this specific implementation choice is that with only one static instance, there is no allocation pressure; this is the recommended approach for such things (it's the reason this API exists). Allocating a byte buffer each time increases allocation pressure and incurs some overhead due to the extra setup every time we want to read a value, or at *best*, every time we set up the stream. The alternative would be to do the math inline to combine the bytes, and hope that HotSpot will eventually recognize the pattern and optimize it for us, but I know of no evidence that it would do so.

wasm/src/main/java/com/dylibso/chicory/wasm/io/BufferWasmInputStream.java
andreaTP: I would love not to have to wrangle with low-level details of the `InputStream` and underlying `ByteBuffer`, do you have an estimate on how much less performant it will be to rely on a `ByteArrayInputStream`?

For what kind of load is this going to become an issue?

wasm/src/main/java/com/dylibso/chicory/wasm/io/BufferWasmInputStream.java
```diff
@@ -0,0 +1,88 @@
...
+
+    BufferWasmInputStream(final ByteBuffer buffer) {
+        super();
+        // make our position independent, and zero-based
+        this.buffer = buffer.slice();
```
andreaTP, line 16: Here we are creating a new ByteBuffer: https://docs.oracle.com/javase/8/docs/api/java/nio/ByteBuffer.html#slice--
getting independent on the index, but still sharing the content.
I'm not sure about this decision:
- if we wanna be independent of the provided `ByteBuffer` we should `duplicate` instead
- If we wanna share the original `ByteBuffer` not considering the current index is not the behavior I would expect as a user

wasm/src/main/java/com/dylibso/chicory/wasm/io/BufferWasmInputStream.java
```diff
@@ -0,0 +1,88 @@
...
+        }
+        long newOffs = buffer.position() + count;
+        if (newOffs > buffer.limit()) {
+            throw new WasmEOFException();
+        }
+        buffer.position((int) newOffs);
+    }
+
+    public int rawByteOpt() {
+        return buffer.hasRemaining() ? Byte.toUnsignedInt(buffer.get()) : -1;
```
andreaTP, line 36: Do we need the `hasRemaining`? Is it breaking on any test?

wasm/src/main/java/com/dylibso/chicory/wasm/io/BufferWasmInputStream.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    public int peekRawByteOpt() {
+        return buffer.hasRemaining() ? Byte.toUnsignedInt(buffer.get(buffer.position())) : -1;
```
andreaTP, line 40: Do we need the `hasRemaining`? Is it breaking on any test?

wasm/src/main/java/com/dylibso/chicory/wasm/io/BufferWasmInputStream.java
```diff
@@ -0,0 +1,88 @@
...
+
+    public void rawBytes(final ByteBuffer buf) {
+        int srcPos = buffer.position();
+        int srcLim = buffer.limit();
+        int dstPos = buf.position();
+        int dstLim = buf.limit();
+        int copySize = dstLim - dstPos;
+        if (copySize <= srcLim - srcPos) {
+            // reduce limit to size of copy
+            buffer.limit(srcPos + copySize);
```
andreaTP, line 61: I struggle to grasp this line, and the comment further confuses me. Would you mind expanding?

wasm/src/main/java/com/dylibso/chicory/wasm/io/InputStreamWasmInputStream.java
```diff
@@ -0,0 +1,139 @@
...
+    public void skip(long count) {
+        // todo: java 12+: is.skipNBytes(count);
+        if (count <= 0) {
+            return;
+        }
+
+        try {
+            long t;
+            while (count > 0) {
+                t = is.skip(count);
```
andreaTP, line 39: Here seems like we are forcing a stricter semantics than the underlying `skip` mechanism, do you mind expanding on the reasons? Would it be possible to just reuse the underlying `is.skip` method?

wasm/src/main/java/com/dylibso/chicory/wasm/io/InputStreamWasmInputStream.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    public int rawByteOpt() {
+        try {
+            int res = is.read();
+            if (res >= 0) {
+                position++;
+                return res & 0xff;
+            } else {
+                return -1;
```
andreaTP, line 61: Does `-1` have a special meaning?

wasm/src/main/java/com/dylibso/chicory/wasm/io/InputStreamWasmInputStream.java
```diff
@@ -0,0 +1,139 @@
...
+    public byte[] rawBytes(final int len) {
+        if (len == 0) {
+            return NO_BYTES;
+        }
+        // todo: java 12+: return is.readNBytes(len);
+        return readAll(new byte[len], 0, len);
+    }
+
+    public void rawBytes(final ByteBuffer buf) {
+        if (buf.hasArray()) {
```
andreaTP, line 90: I see around a bunch of optimizations on the usage of the `ByteBuffer`, would it make sense to refactor those in a separate utility class so that they apply to any usage of `ByteBuffer` in the codebase?
Looks a bit of complexity and it should probably be tested thoroughly to ease any debugging in the future.

wasm/src/main/java/com/dylibso/chicory/wasm/io/InputStreamWasmInputStream.java
```diff
@@ -0,0 +1,139 @@
...
+        for (int p = 0; p < len; p += res) {
+            try {
+                res = is.read(buf, p + off, len - p);
+            } catch (IOException e) {
+                throw new WasmIOException(e);
+            }
+            if (res == -1) {
+                throw new WasmEOFException();
+            }
+            position += res;
```
andreaTP, line 117: Would it be possible to avoid keeping the position ourselves and delegating to the underlying `InputStream`?
Any consideration to be done WRT concurrency?

wasm/src/main/java/com/dylibso/chicory/wasm/io/WasmInputStream.java
```diff
@@ -0,0 +1,641 @@
...
+    }
+
+    /**
+     * Read a ULEB value returning up to 64 significant bits, with range validation.
+     *
+     * @param bits the number of significant bits allowed
+     * @return the value
+     * @throws WasmIOException if an underlying I/O error occurs or the value exceeds the allowed range
+     */
+    private long uleb(int bits) throws WasmIOException {
```
andreaTP, line 482: We already have Leb implementations in `com.dylibso.chicory.wasm.Encoding`, I think we should consolidate on one.

wasm/src/main/java/com/dylibso/chicory/wasm/io/BufferWasmInputStream.java
dmlloyd: Of course. Please read the documentation of the base class method. If the end-of-file is reached, then -1 shall be returned, else the next byte in the stream is returned. Without `hasRemaining`, an exception will be thrown at EOF, which is contrary to the specification of the method that this is implementing.

wasm/src/main/java/com/dylibso/chicory/wasm/io/BufferWasmInputStream.java
dmlloyd: See above.

wasm/src/main/java/com/dylibso/chicory/wasm/io/BufferWasmInputStream.java
dmlloyd: The documentation for `ByteBuffer` and `ByteBuffer#put(ByteBuffer)` explains clearly the requirements for how position and limit works. Essentially the source buffer has to be limited in size in order to avoid overflowing the target buffer, but we also must restore the original limit afterwards.

If `ByteBuffer` is confusing you, you might find [this guide](https://docs.jboss.org/xnio/2.x-userguide/html/buffers.html) helpful which I wrote for XNIO many years ago. Just ignore the XNIO parts.

wasm/src/main/java/com/dylibso/chicory/wasm/io/InputStreamWasmInputStream.java
dmlloyd: No, it is not possible because the contracts are in conflict. The use case for this method is that you want to unconditionally skip some amount of bytes (for example, skipping a section that we do not want to read). If fewer bytes are skipped, then the stream will be in an indeterminate state, which is not useful to the user. So if we do not provide the stricter contract, then the user of this method will have to do so themselves. The job of an API is to provide a useful abstraction over a lower-level construct, and this is one example of how we do so.

Note that Java 12 and later has a `skipNBytes` API on `InputStream` which implements this same logic for this same reason; once we can move beyond Java 11, we can replace most of this implementation code with a call to that method.

wasm/src/main/java/com/dylibso/chicory/wasm/io/InputStreamWasmInputStream.java
dmlloyd: Yes. Please read the documentation of the superclass method; as with `InputStream`, this method returns a byte value between 0-255 inclusive, or otherwise returns -1 if the end of file is reached.

wasm/src/main/java/com/dylibso/chicory/wasm/io/InputStreamWasmInputStream.java
dmlloyd: I don't know of any other place or way to factor this out, nor any other use site where this would be used since it's specific to this class. Did you have something in mind? What optimizations are you referring to?

wasm/src/main/java/com/dylibso/chicory/wasm/io/InputStreamWasmInputStream.java
dmlloyd: `InputStream` does not track position. No input stream (JDK or `WasmImputStream`) is thread-safe, and thus they must never be used from more than one thread concurrently.

wasm/src/main/java/com/dylibso/chicory/wasm/io/WasmInputStream.java
dmlloyd: That class will no longer be used after this change. A follow-up PR will remove it.

wasm/src/main/java/com/dylibso/chicory/wasm/io/BufferWasmInputStream.java
andreaTP: > We use a slice because we need the byte content to extend from position zero

I'm failing to get the "why", I think that there might be use-cases where preserving the current position would be the expected behavior.

dmlloyd: The contract of this class is that reading extends from the start to the end (no random access is allowed), and that the `position` of the stream reports the number of bytes consumed by the stream. So, we have to ensure that the buffer position increases monotonically (hence having our own copy); after that, using the buffer position as the stream position follows logically. But in order to do so, the position must start at zero, hence the slice.

wasm/src/main/java/com/dylibso/chicory/wasm/io/BufferWasmInputStream.java
andreaTP: The more I go through those details the more I would love to compare an implementation based on a higher level API such as `ByteArrayInputStream` that already implements those details.

wasm/src/main/java/com/dylibso/chicory/wasm/io/InputStreamWasmInputStream.java
andreaTP: Fair, thanks for explaining!

wasm/src/main/java/com/dylibso/chicory/wasm/io/WasmInputStream.java
andreaTP: Any challenge including the change in this PR in the spirit of keeping things more "atomic"?

wasm/src/main/java/com/dylibso/chicory/wasm/io/BufferWasmInputStream.java
dmlloyd: You cannot use ByteArrayInputStream to wrap a ByteBuffer, so I fail to see how it is possible to use that class for this case.

wasm/src/main/java/com/dylibso/chicory/wasm/io/InputStreamWasmInputStream.java
dmlloyd: Oh, yes, I see. The slow path is *very* slow though, so we'd want to avoid it if possible.

I don't think there's a direct equivalent in `Memory` but we could always introduce a utility class in a later follow-up if needed.

wasm/src/main/java/com/dylibso/chicory/wasm/io/BufferWasmInputStream.java
andreaTP: > it's not really low level in any way either
> It's quite straightforward and the implementation is pretty minimal.

I completely understand your reasons for saying so, but, I do believe that this is not completely objective, let me elaborate.

I would stay away from the complexities of having to deal directly with the `ByteBuffer` API when we can avoid so, by accepting the PR in its current form we take on the challenges of fixing bugs in this implementation.

If there are no big performance or other implications I would love to see an implementation using a higher-level Java API/construct and doing the minimal job(none to little) required to conform to the interface.

wasm/src/main/java/com/dylibso/chicory/wasm/io/WasmInputStream.java
dmlloyd: Sure, I can move that change.

wasm/src/main/java/com/dylibso/chicory/wasm/io/InputStreamWasmInputStream.java
andreaTP: Commented above.

wasm/src/main/java/com/dylibso/chicory/wasm/io/BufferWasmInputStream.java
andreaTP: above

wasm/src/main/java/com/dylibso/chicory/wasm/io/BufferWasmInputStream.java
andreaTP: > The contract of this class is that reading extends from the start to the end

Do you have a reference? I'm failing to find this detail other than in the implementation.

I see a possible use-case where you are reading something that "contains" WASM.
In this case, the user might want to consume part of the Stream, and directly pass the buffer to be consumed starting from the current position.
This doesn't break monotonicity but doesn't need to start at 0.

wasm/src/main/java/com/dylibso/chicory/wasm/io/BufferWasmInputStream.java
andreaTP: Solid explanation, works for me, thanks for elaborating 👍

wasm/src/main/java/com/dylibso/chicory/wasm/io/WasmInputStream.java
dmlloyd: Looks like I have to do some commit log surgery to get the whole thing. One moment.

dmlloyd: Now it should be OK.

wasm/src/main/java/com/dylibso/chicory/wasm/io/InputStreamWasmInputStream.java
andreaTP: Let's keep this in mind in the future, thanks for going through it 👍

wasm/src/main/java/com/dylibso/chicory/wasm/io/InputStreamWasmInputStream.java
andreaTP: Thanks, appreciated 👍

wasm/src/main/java/com/dylibso/chicory/wasm/io/BufferWasmInputStream.java
dmlloyd: There is no other alternative. If a WASM module is in a `ByteBuffer` (for example, due to coming from a very large memory-mapped file), then unless we copy the entire thing to the heap (potentially gigabytes?), we must read the data where it is. It's not directly about performance because this is the most direct implementation for this use case. And, there is no higher-level construct to model this concept (until Java FFM in Java 21/22/23+ which introduces `MemorySegment` - which would probably be considered even more low-level, though it is also a better fit for this task).

I understand wanting to keep the code simple. But, "keeping it simple" to me means "avoid using subtle/tricky code practices without documentation", "factor out complexity", "hide necessary complexity behind usable APIs", and things like that; it does not mean "avoid core JDK APIs that I don't understand well". As Java software engineers, understanding the core JDK APIs is an incumbent and prerequisite part of our job responsibilities. The behavioral contract of `ByteBuffer`'s methods are very clear and well-documented, and the way that this class utilizes `ByteBuffer` should also be quite clear here. There is little subtlety in this implementation; it presents a stream view over a buffer slice in the most direct way possible.

wasm/src/main/java/com/dylibso/chicory/wasm/io/BufferWasmInputStream.java
dmlloyd: Implicit in the definition of "stream" is that it progresses monotonically from start to end. I'm not aware of any "stream" type class in any commonly used Java API that does not behave this way.

The use case you illustrate is supported directly by this API. If the user passes the buffer to be consumed, the stream will read from that buffer starting at the buffer's position and extending to the buffer's limit. The stream's reported position will start at zero. The user's buffer position and limit is not affected, and the user can independently change their buffer position and limit without affecting the stream position.

wasm/src/main/java/com/dylibso/chicory/wasm/io/BufferWasmInputStream.java
dmlloyd: No, it will give an error on read-only buffers and direct (off-heap) buffers, the latter of which being part of the use case of this implementation.

wasm/src/main/java/com/dylibso/chicory/wasm/io/BufferWasmInputStream.java
andreaTP: I have no interest in continuing the discussion after this passive-aggressive comment.

dmlloyd:
If there are no further comments, I'd like to move ahead with this.

andreaTP:
https://github.com/dylibso/chicory/blob/main/CONTRIBUTING.md#code-reviews

dmlloyd:
I'm aware, thanks. Are you saying you are not interested in having this change, or my subsequent changes which depend on it, and that your citing of this policy is indicative of your decision to block this change? If so, it would probably be better to say so directly, because then I can get on with developing what I need under my own fork without having to be delayed by upstreaming. If not, then let's continue the discussion properly.

dmlloyd:
@andreaTP, I'll add a third option: non-response. If you are no longer interested in reviewing this change, I will dismiss your review and allow someone else to have a look instead.

bhelx:
Sorry I'm just getting around to looking at this PR. I'm supportive of where this is going, but I think we should hold off on it. There are a lot more changes than I thought would be necessary. I'm having trouble understanding it and keeping up. I'd like to keep work focused on the milestones: Improving our test posture and specification coverage. I worry all these changes are making it harder to understand what is happening in the codebase. I understand some of the arguments for how these are improvements, but the cost of not understanding them is too high this early in the project.

dmlloyd:
If you look at `WasmInputStream` as an encapsulated interface, it should be very straightforward: methods to read each exact value type without casting, smart handling of sections, lookahead (peek) is handled for you. The interface itself is simple and as spec-correct as possible, and that is the most important idea.

To help understand how this simplifies things, start by looking at the code where sections are parsed. In each case, the code is simplified or at worst made more readable by eliminating casts (or replacing them with explicit transformations in a few select places), and reading things by their WASM type as given in the spec. Having a spec-direct and simple API also serves to make things more correct (for example, using limited-range *LEB decoding to detect errors on invalid input which previously were allowed, and correcting our still-incorrect UTF-8 validation behind a simple API).

In terms of encapsulation, this PR moves code complexity from a central hub class (`Parser`) to a leaf class which encapsulates the idea of reading WASM data, and removes direct reliance on `ByteBuffer`, which has already been labeled as user-unfriendly in this PR comment section. Its symmetrical counterpart, `WasmOutputStream` which is already written, will do the same for output.

The immediate next step after this patch is to move each section's individual parsing code to that section's class, further reducing the conceptual complexity of `Parser` (which today does many distinct jobs related to parsing). This patch is also ready to go.

An additional and critical benefit of this code is of course that currently, the entire WASM program must be held in memory, even if it is quite large, especially in the case where you start with an `InputStream` (where the entire stream must be copied). By encapsulating input behind a stream, we can still support in-memory buffers (including memory-mapped files), and also plain `InputStream`. Since there is a layer of encapsulation, we can also support reading from a `MemorySegment` in Java 22+ by adding one implementation class. Retaining `ByteBuffer` as an input type is not tenable, and it is definitely not tenable as an output type (since buffers must be pre-sized).

PR #250, opened 2024-02-19
First validations for malformed bytecode

mariofusco:


runtime/src/test/java/com/dylibso/chicory/testing/TestModule.java
```diff
@@ -17,6 +18,14 @@ public static TestModule of(File file) {
         return new TestModule(file);
     }
 
+    public static TestModule of(File file, ModuleType moduleType) {
+        if (moduleType == ModuleType.TEXT) {
```
andreaTP, line 22: This change is going to affect only the execution of the tests in this codebase, we should probably revisit this piece of code: https://github.com/dylibso/chicory/blob/5dadeaa21241f3f0d1ada659fb9670490f1a7e30/runtime/src/main/java/com/dylibso/chicory/runtime/Module.java#L493

test-gen-plugin/src/main/java/com/dylibso/chicory/maven/JavaTestGen.java
```diff
@@ -437,13 +448,25 @@ private List<Expression> generateAssertThrows(Command cmd, String wasmFile) {
...
-            return List.of(assertThrows);
+            method.getBody().get().addStatement(exceptionMessageMatch(cmd.text()));
+        }
+
+        if (excluded) {
+            method.addAnnotation(
+                    new SingleMemberAnnotationExpr(
+                            new Name("Disabled"),
+                            new StringLiteralExpr("Validation test excluded")));
+        } else if (cmd.moduleType() != null && cmd.moduleType().toLowerCase().contains("text")) {
```
andreaTP, line 464: nitpick
```suggestion
        } else if (cmd.moduleType() != null && cmd.moduleType().toLowerCase().equals("text")) {
```

wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java
```diff
@@ -399,6 +401,19 @@ private static ImportSection parseImportSection(ByteBuffer buffer) {
         return importSection;
     }
 
+    private static MutabilityType getMutabilityType(ByteBuffer buffer) {
+        MutabilityType globalMut = null;
+        try {
+            globalMut = MutabilityType.forId((int) readVarUInt32(buffer));
```
andreaTP, line 407: For consistency with the "style" started [here](https://github.com/dylibso/chicory/pull/244/files#diff-b617d5b6a9607327caad59874d23af108d8b5dac7d8225f4726db391075c7dd6R141) I would move the throwing of the exception inside the `MutabilityType.forId` method.

cc. @dmlloyd

runtime/pom.xml
```diff
@@ -145,6 +145,7 @@
             SpecV1AddressTest.test233, SpecV1AddressTest.test234, SpecV1AddressTest.test236,
             SpecV1AddressTest.test252, SpecV1AddressTest.test253, SpecV1AddressTest.test255
           </excludedTests>
+          <excludedValidationTests>binary,binary-leb128,custom</excludedValidationTests>
```
andreaTP, line 148: Since we are excluding at the file level is it possible to have the filename listed here?
e.g. `binary.wast, custom.wast,`

andreaTP (changes requested):
A few minor comments, but I love where this is going!
Thanks!

wasm/src/main/java/com/dylibso/chicory/wasm/ParserUtil.java
```diff
@@ -0,0 +1,27 @@
...
+        return true;
+    }
+
+    private static boolean isValidIdentifierChar(int ch) {
+        return Character.isUnicodeIdentifierPart(ch)
+                || ch == '-'
+                || ch == '>'
+                || ch == ' '
+                || ch == '.'
+                || ch == '⌣';
```
andreaTP, line 25: Haven't noticed this before, looks like a very specific character, do you have a reference to the spec or is it just empirical?

wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java
dmlloyd: In #249 I've replaced this parsing code with `WasmInputStream#mut`.

test-gen-plugin/src/main/java/com/dylibso/chicory/maven/JavaTestGen.java
dmlloyd, line 464: `toLowerCase` should always accept a locale (e.g. `Locale.ROOT`); otherwise, another way to do this would be `equalsIgnoreCase`.

wasm/src/main/java/com/dylibso/chicory/wasm/ParserUtil.java
mariofusco: I know it's very specific and pretty weird. If I don't do this all tests generated from `custom.wast` fail. I'm pretty sure there should a more a general way to check the validity of a character, but so for I haven't find any.

wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java
mariofusco: If I do so I should actually duplicate where the exception is thrown, both here and in `MutabilityType.forId` because in in some cases also the invocation of `readVarUInt32(buffer)` throws an Exception and in this way I'm dealing with both situation here.

andreaTP: > in some cases also the invocation of `readVarUInt32(buffer)` throws an Exception

Interesting, do you know in which circumstances?

wasm/src/main/java/com/dylibso/chicory/wasm/ParserUtil.java
andreaTP: I went back to the specification and import names are defined as [Names](https://webassembly.github.io/spec/core/syntax/values.html#names) and all the Unicode characters are allowed.
I hope we can come up with a more principled check here.

runtime/pom.xml
mariofusco: Done.

wasm/src/main/java/com/dylibso/chicory/wasm/ParserUtil.java
mariofusco: Thanks for pointer, following it I found a more reasonable way to test allowed chars.

andreaTP: Nice! Thanks! 👍

wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java
```diff
@@ -339,8 +344,8 @@ private static ImportSection parseImportSection(ByteBuffer buffer) {
 
         // Parse individual imports in the import section
         for (int i = 0; i < importCount; i++) {
-            String moduleName = readName(buffer);
-            String importName = readName(buffer);
+            String moduleName = readName(buffer, false);
```
andreaTP, line 347: same as the default, we can use the single parameter version, here and the next.

wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java
```diff
@@ -167,17 +168,18 @@ void parse(InputStream in, ParserListener listener) {
                     "unexpected token: unsupported version, found: " + version + " expected: " + 1);
         }
 
+        var firstTime = true;
```
andreaTP, line 171: I think we should leave a comment for our future selves on this "trick", but I'm ok with it if there is no better way ...

wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java
```diff
@@ -931,9 +934,25 @@ public static long readFloat32(ByteBuffer buffer) {
...
+        }
+        var name = new String(bytes, StandardCharsets.UTF_8);
+        if (!isValidIdentifier(name) && !allow) {
+            throw new MalformedException("malformed UTF-8 encoding");
+        }
+        return name;
+    }
+
+    private static boolean isValidIdentifier(String string) {
+        return string.chars().allMatch(ch -> ch < 0x80 || Character.isUnicodeIdentifierPart(ch));
```
dmlloyd, line 958: The specification says only "[Names](https://webassembly.github.io/spec/core/syntax/values.html#syntax-name) are encoded as a [vector](https://webassembly.github.io/spec/core/binary/conventions.html#binary-vec) of bytes containing the [Unicode](https://www.unicode.org/versions/latest/) (Section 3.9) UTF-8 encoding of the name’s character sequence." Note that they do not specify any characters as being valid or invalid; therefore, it's not correct to restrict the valid character space of names in any way AFAICT.

mariofusco: I see your point and indeed it's very likely that this is not 100% compliant with the spec and that we're leaving out some valid characters. That said @andreaTP and me have already spent a quite relevant amount of time on this and we couldn't find a better solution. I believe that there isn't much value added in spending more time on it at this point. We are aware that we are probably leaving out some valid characters. Nevertheless my suggestion is to merge this pull request as it is, move forward, and come back to this only when we will have a test evidence that this validation isn't working as it should in all cases.

dmlloyd: My point is that name validation is completely unnecessary, because all Unicode strings are allowed by spec (even the empty name). Adding additional restrictions changes the implementation from being strictly correct to being strictly incorrect. It's not a matter of degrees. Furthermore as I stated above, #249 removes this code from `Parser` in any event.

mariofusco: > My point is that name validation is completely unnecessary, because all Unicode strings are allowed by spec.

This is untrue. If I try to remove this validation I have a total of 528 test failures in `SpecV1Utf8ImportModuleTest`, `SpecV1Utf8ImportFieldTest` and `SpecV1Utf8CustomSectionIdTest`, all expecting a `MalformedException` to be throw but without the validation isn't.

dmlloyd: According to [https://webassembly.github.io/spec/core/syntax/values.html#names](the Names section of the spec), the name codepoints are expected to be within the range of Unicode; perhaps there is a code point outside of that range? Perhaps it's not the character but the encoding? According to [the binary format spec for Names](https://webassembly.github.io/spec/core/binary/values.html#names) it seems to imply that the shortest encoding is required, so maybe the expected failures are related to using longer-than-minimal encoding?

It'd be really good to understand why the test fail rather than patch it over IMO.

mariofusco: If you're in doubt try this pull request giving a run to one of the tests I mentioned. For instance if you execute `SpecV1Utf8ImportModuleTest.test0` you will see that at some point the import section reads a name with a single character having codepoint 65533 that is invalid and correctly recognized as such by that method.

dmlloyd: Ah so I think my suspicion is correct. The code point isn't invalid, its encoding is.

In UTF-8, each character can be encoded in potentially many ways. The leading bits of each UTF-8 character determine how many bits there are in the final character value. However, it is possible to encode a Unicode code point into a UTF-8 sequence that is longer than necessary. It is also possible to provide byte sequences that are invalid. That appears to be what is tested by this test. The JDK will attempt to accept these values in various ways, so in order to property validate, we need a completely different approach to UTF-8 decoding which validates each byte more strictly than what the JDK is doing.

mariofusco: Actually @andreaTP already attempted to do [here](https://github.com/mariofusco/chicory/pull/1/files#diff-d5bf04fbcefd84882667b907d29a2b88ceaa9ff6fb176408f3b5dc6f5f24883f) something similar to what (I believe) you're suggesting, but we didn't manage to make it 100% work and anyway it seemed too much complicated at least for our current needs. My suggestion is to avoid wasting more time on this for now (I guess we have more important things to do), merge what we have and eventually come back to this when and if necessary.

PR #251, opened 2024-02-22
Ref types and type-safe compact MStack

dmlloyd:
* Create box types for function and external reference types, with common `Ref` base class
* Create temporary compat external value for compatibility with integer external refs from test scripts (drop this after test script interpreting updates to use test external ref objects)
* Type-safe, memory-compact and GC-friendly MStack implementation with documentation
* Update `SELECT_*` to use `drop` operation
* Tests

Amortized memory usage of new impl is:

* zero objects per stack entry
* 4/8 bytes (array entry) per stack entry (depends on platform bits) (could be reduced to a single byte, perhaps?)
* 4 bytes (array entry) per I32/F32
* 8 bytes (array entries) per I64/F64
* 16 bytes (array entries) per V128
* 4/8 bytes (array entry) per *Ref (depends on platform bits)

Amortized memory usage of pervious `Value`-based impl is:

* one object per I32/I64/F32/F64: ~12/16 bytes (object fields) + header + padding
* 4/8 bytes (array entry) per entry (any type, depends on platform bits)
* V128 not supported at all, refs only supported as integers

TODO:

* Fix external ref and function ref handling so that they consistently use box objects in all contexts, drop integer representation
* Establish import/export-friendly scheme for function refs (discussion?)
* Investigate stack underflow problem in `StackFrame#dropValuesOutOfBlock` (see comment)
* Add mark-stack in to `MStack` to replace `stackSizeBeforeBlock` stack, which uses autoboxing profusely

Future:

* Function arguments and local variables could live on the `MStack` instead of a `Value[]`-per-call, which would not only eliminate excessive boxing of values but also eliminate copying of values between caller and callee (caller pushes arguments on to the stack; callee establishes a "true" stack frame starting at the first argument and also reserves space for local variables, all of which is popped on return using `stack.drop()` to preserve any number of return value(s))


danielperano (approved):

danielperano:
Not sure this is worth pursuing now (or if it's even worthwhile longer term), but one thing I've done in the past to elide extra array bounds checks when using an array as stack storage is to optimistically attempt to write to an index, catch the exception if the array is too small, and then grow the array and repeat the operation. The downsides are that this is only possibly faster if the throws are quite rare, and the code is not nearly so clean.

If the method that grows the array gets inlined, I also wonder if the JVM will even do bounds checks in all locations following an initial check or if it will optimize them out (come to think of it, in most scenarios I'd guess the bounds checks incur no real runtime penalty unless the code that's running keeps the CPU cache hot enough to avoid pausing for cache <-> RAM transfers, which I assume is fairly rare).

runtime/src/main/java/com/dylibso/chicory/runtime/MStack.java
```diff
@@ -1,42 +1,797 @@
...
+                throw new IllegalStateException();
+        }
+    }
+
+    /**
+     * Push a zero value of the given type on to the stack.
+     *
+     * @param type the value type (must not be {@code null})
+     */
+    public void pushZero(ValueType type) {
```
mariofusco, line 123: Probably I'm missing this, but I don't see where this method is used, neither I fully understand why we need a special method to push a zero on the stack instead of relying on other existing methods.

runtime/src/main/java/com/dylibso/chicory/runtime/StackFrame.java
```diff
@@ -104,9 +104,9 @@ public void dropValuesOutOfBlock(MStack stack) {
             }
 
             int expectedStackSize = stackSizeBeforeBlock.pop();
-            while (stack.size() > expectedStackSize) {
-                stack.pop();
-            }
+            // todo: the expected stack size is often larger than the actual stack; investigate &
```
mariofusco, line 107: This todo is a bit worrying and the current implementation is probably just hiding a bug somewhere else. I'd probably investigate this further before merging this.

runtime/src/main/java/com/dylibso/chicory/runtime/MStack.java
```diff
@@ -1,42 +1,797 @@
...
+            System.arraycopy(
+                    refs, refDepth - skipRefs, refs, refDepth - dropRefs - skipRefs, skipRefs);
+        }
+        // clear refs for GC
+        Arrays.fill(refs, refDepth - dropRefs, refDepth, null);
+        // finally, shrink the stacks
+        depth -= dropCount;
+        wordDepth -= dropWords;
+        refDepth -= dropRefs;
     }
```
bhelx, line 339: This is going to take some time for me to wrap my head around what is happening in this class. I understand that this level of control should make some performance improvements, but we're sacrificing the simplicity and safety of just using a standard type like Stack or ArrayDeque. It's hard to do a cost benefit-analysis on this kind of change, but it feels a bit premature before we have any users or the AOT compilation started.

runtime/src/main/java/com/dylibso/chicory/runtime/StackFrame.java
dmlloyd: This patch didn't cause this issue, but it did uncover it. I changed the old pop loop, which pops until the expected size, to pop to the previous depth exactly (see the comment about removing `Math.min`), and it cause failures. So I added in the `Math.min` to replicate the previous behavior for later investigation.

runtime/src/main/java/com/dylibso/chicory/runtime/MStack.java
dmlloyd: The gist if it is to copy the "keep" values to the lower array position, which is calculated arithmetically above by counting the number of substack slots. Since we have separate array substacks for the type ID, integral words, and ref types, we have to do three copies to copy down each "keep" region, and we also have to fill in the abandoned spot in the refs substack to avoid holding strong references to ref values after they are discarded.

runtime/src/main/java/com/dylibso/chicory/runtime/MStack.java
dmlloyd: This method would be used if/when stack frames are introduced: a zero value of each local type has to be pushed to reserve space for them. Since we start with a list of types, we can do it quickly this way e.g. `localTypes.forEach(stack::pushZero)`.

runtime/src/main/java/com/dylibso/chicory/runtime/StackFrame.java
dmlloyd: I guess the original problem is subtle. The intent of the loop above is "while the stack is bigger than the expected size, pop", but it also passes quietly if the stack size is *smaller* than the expected size. This subtle behavior is what went undetected, and it's almost certainly incorrect.

dmlloyd:
> Not sure this is worth pursuing now (or if it's even worthwhile longer term), but one thing I've done in the past to elide extra array bounds checks when using an array as stack storage is to optimistically attempt to write to an index, catch the exception if the array is too small, and then grow the array and repeat the operation. The downsides are that this is only possibly faster if the throws are quite rare, and the code is not nearly so clean.

Yeah for simple pushes this would be a workable technique, but for multi-stage pushes it might be slightly uglier. It's possible that the ugliness (if any) could be encapsulated behind (inlinable?) utility methods though (one for each substack). It's a good idea.

runtime/src/main/java/com/dylibso/chicory/runtime/MStack.java
danielperano: > It's hard to do a cost benefit-analysis on this kind of change, but it feels a bit premature before we have any users or the AOT compilation started.

I'm of the opinion that this is better done sooner rather than later - this puts a lot less pressure on the GC, and probably also helps a fair amount with CPU cache locality and memory bandwidth overhead. It also has fairly high risk of impacting other foundational classes (such as `Machine`, as happened in this PR). I've generally found that in interpreters and compilers, foundational changes are easier to make sooner rather than later.

I agree that there are plenty of optimizations that are best left for later, but I'm in favor of picking off low-hanging fruit (especially if it has large and generally beneficial impact on the codebase) sooner rather than later. My concern with only writing code that pays little to no attention to performance is that if users start picking it up now it might get a reputation of being "slow," and that can be hard to shake.

For AOT, I think this is either irrelevant or minimal impact - it may be useful to use `MStack` for some type tracking tasks during bytecode generation, but I'm not expecting impact beyond that.

dmlloyd:
Implemented the suggested array-growth optimization.

danielperano (approved):

danielperano:
> Implemented the suggested array-growth optimization.

Looks great - came out a lot cleaner than my implementations typically have!

runtime/src/main/java/com/dylibso/chicory/runtime/MStack.java
andreaTP: > a bit premature before we have any users

I'm aligned, we agreed on completing full support for the WASM specification before we dig into invasive performance improvements.
Currently, we are barely starting to crack validation.

> I'm in favor of picking off low-hanging fruit 

LOC is not always a good indication of the complexity of something, but, at least, we need to review and go through 1700+ lines to incorporate this change:
Screenshot 2024-02-28 at 09 40 44
calling it a "low-hanging fruit" sounds a bit of a stretch to me.

dmlloyd: With respect, I don't think it is likely that the full WASM spec is supportable under the current architecture without incurring substantial technical debt. Speaking as someone who has worked on *hundreds* of OSS projects, it is absolutely and unequivocally *always* best to design for the future. For example, I can say, without reservation, that in the future having box objects for every value is going to be a significant problem that will be much harder to solve later than it is to solve now.

This patch is as much about having a clean future-ready API which supports all of the spec types as it is about performance.

This patch is also about simplicity. Yes, you are correct that the change adds a lot of lines, because the old API usages are replaced with the new API. `cloc` tells me that the `MStack` implementation itself is 613 lines of code (of which 10% is dedicated to just having a nice `toString` method) plus 345 lines of comments. The tests add 369 lines (tests are good, right?). The remaining LOC count is reworking and simplifying the existing code to use the new implementation efficiently. It's disingenuous to characterize the patch as complex based on +1722 additions.

What you get for 613 lines is a clean and efficient stack implementation which supports every existent WASM type with minimal memory overhead. Simplicity isn't about avoiding complexity. It's about *managing* complexity by encapsulating it behind simple interfaces, and that's exactly what this patch strives to do. Try not to think about how complex the implementation itself is, but how simple the API it presents is. With this API you can push and pop each primitive type, and even vector types, without additional box objects in a way that is idiomatic and intuitive. It also paves the way for call stack frames, which is something that will surely be much harder to implement as time goes on.

A system with clean APIs and strong encapsulation around arbitrary complexity is, by any reasonable calculation, simpler overall than a system which avoids any kind of complexity but also lacks encapsulation. It's never too early to define strong encapsulation, and the cost of not doing so is often very high.

bhelx:
I took some time at lunch today to re-read the code and try to understand the changes to MStack, as well as all of the comments. All arguments about simplicity and what that means aside, I still don't really understand the code. I understand the intention of the code, but I don't have confidence I can easily fix it if it's broken. Whereas today, if there is a problem with the stack (or any of the other components) I can easily understand and fix it. Given we've (Dylibso) taken on the responsibility of maintaining this project, I can't merge this right now and would rather wait to mainline these kind of optimizations until our community grows some more.

I understand any arguments about this not being a sustainable way to develop a project like this, but I'm not in a hurry to change it and I think doing this has outsized costs on the community and participation which I take as the highest priority at this early phase of the project. @andreaTP and I both agreed to this approach when we started. I respect that there is reasonable disagreement on this way of running things, but it's not something that I consider to be open to debate at this time. Of course in the the future we will need to make tradeoffs to this, but we don't yet have a good framework for doing so. Apologies for not making this clearer earlier. I will try to write some more in our contributing guide about this. But, in this case I just underestimated how complex these changes would be.

danielperano:
> All arguments about simplicity and what that means aside, I still don't really understand the code. I understand the intention of the code, but I don't have confidence I can easily fix it if it's broken. Whereas today, if there is a problem with the stack (or any of the other components) I can easily understand and fix it.

With the exception of the `drop()` method, I personally find this easier to understand than the prior implementation when it comes to types and type conversions because the type information of the expected value is encoded within the stack operation calls rather than separately. With the boxed representation each stack op is actually two ops: the stack op itself and (un)boxing the value correctly. This merges them into one operation which to my eye makes the opcode implementations cleaner and easier to read - something that's especially valuable for some of the non-trivial opcodes. I find pushing/popping primitives simpler and easier to immediately understand than pushing and popping values that must be manually (un)boxed. Further, this still supports the boxed form, so if there are other areas where working with boxed forms is easier we're not actually losing anything.

The one downside I see to this PR is that `drop()` is pretty involved. To some degree that's inherent to the difficulties posed by having split-stack storage like this, but I'd guess there might be some refactorings that could be done that would significantly enhance its readability. I share @dmlloyd's concern that delaying this could result in accruing tech debt down the road that's significantly harder to eliminate then than it is now, and I think that for the most part this PR serves the purposes of simple and clean code at least as well as the value boxes do, if not better.

Is there a specific set of areas of the code that make you reluctant to merge this that if addressed accordingly could change your mind? (I agree with attempting to refactor `drop()` to be more immediately understandable, for example.)

bhelx:
> Is there a specific set of areas of the code that make you reluctant to merge this that if addressed accordingly could change your mind? (I agree with attempting to refactor drop() to be more immediately understandable, for example.)

Yeah the area i am concerned about is the MStack.java changes. Before, we simply relied on java's `ArrayDeque` and after we have a lot of custom code to maintain. David's a great programmer and I'm sure all this code is high quality, but I'm not really confident I can maintain it at this moment in time. I understand there are plenty of times where you want trade off maintenance burden for performance, but we've explicitly been very careful about only making low hanging fruit performance changes or ones that are necessary.

I'd like to put a halt on all the performance based changes unless it's deleting code or just a small surface change. When everything is green and we have some more users, testing infra, performance tracking, etc, I'll feel more confident about making these tradeoff decisions for performance.

danielperano:
I see where you're coming from. If we're not going to merge this now, I'd at least like to open an issue and link this PR as a reminder to revisit it down the road - I highly value the possibility of eliminating boxed numerics, and I want to make sure this doesn't fall by the wayside by being buried in the sea of PRs that I hope we'll be receiving. ;)

andreaTP:
> I'd at least like to open an issue and link this PR as a reminder to revisit it down the road

Definitely works for me, thanks for understanding!

PR #265, opened 2024-03-25
Fix test module links in README

electrum:


nilslice (approved):
Thank you! Will let @andreaTP or @bhelx make the final call on merge just in case there are other changes they'd like to make.

andreaTP (approved):
Thanks a lot! Lgtm

PR #266, opened 2024-03-26
Use test scope for JUnit dependencies

electrum:


electrum:
I see now that the scope is defined in the `dependencyManagement` section of the parent POM, so this change isn't strictly necessary. Personally, I find it helpful from a readability standpoint to have the scope at the usage site, but this is a matter of preference, so feel free to close this if you disagree.

andreaTP:
Hi @electrum and thanks for your interest in the project!
Have you attempted to use it? In case I'm curious what is your use case and how we can help? :-)

As per this PR, I would decline it if there is no compelling reason, we decided to keep the scopes in the `dependencyManagement` section and this is serving us well enough at this stage of the project.

electrum:
@andreaTP I work on [Trino](https://github.com/trinodb/trino), a distributed SQL query engine. I'm investigating using WASM to support user-defined functions written in Python, JavaScript, etc., as an alternative to our native [SQL routines](https://trino.io/docs/current/routines/introduction.html).  Chicory is interesting since Trino is also written in Java, though with the new [FFM API](https://docs.oracle.com/en/java/javase/22/core/foreign-function-and-memory-api.html), using Wasmer or another library is an option. I'm currently in the exploration phase and am implementing enough of WASI to run Python.

nilslice:
@electrum - that's great to hear! happy to help out wherever we can. 

if you're exploring the FFM route, I'd also suggest taking a look at another project of ours, [Extism](https://github.com/extism/extism), which is far easier to implement (though we don't support Python yet) compared to a runtime directly, and we have a [Java SDK](https://github.com/extism/java-sdk).

PR #278, opened 2024-04-16
Add most of `linking.wast`

andreaTP:
This is one of the remaining 2 `wast` files we excluded so far.
The implementation is a little sprawling and changes the public API, happy to go into the detail of each decision when necessary.


runtime/pom.xml
```diff
@@ -41,7 +41,7 @@
         <artifactId>test-gen-plugin</artifactId>
         <version>${project.version}</version>
         <configuration>
-          <wastToProcess>labels.wast,
+          <orderedWastToProcess>labels.wast,
```
andreaTP, line 44: The tests are completely not self-contained, and the execution depends on the ordered sequence of tests.
In this way is going to be a little harder to execute something in isolation, but this implementation better reflects the very nature of those tests.

runtime/src/main/java/com/dylibso/chicory/runtime/HostGlobal.java
```diff
@@ -20,12 +25,34 @@ public HostGlobal(String moduleName, String fieldName, Value value, MutabilityTy
         this.fieldName = fieldName;
     }
 
+    public HostGlobal(
```
andreaTP, line 28: We are defining a "proxy" `HostGlobal`, this is one possible way to implement the functionality.

runtime/src/main/java/com/dylibso/chicory/runtime/Module.java
```diff
@@ -76,63 +75,18 @@ public Logger logger() {
...
+        return this.instantiate(hostImports, true, true);
     }
 
-    public Instance instantiate(HostImports hostImports, boolean start) {
+    public Instance instantiate(HostImports hostImports, boolean initialize, boolean start) {
         var globalInitializers = new Global[] {};
         if (this.module.globalSection() != null) {
             globalInitializers = this.module.globalSection().globals();
         }
-        var globals = new Value[globalInitializers.length];
```
andreaTP, line 91: this is duplicated code, the logic should be executed during `Instance.initialize`

runtime/src/main/java/com/dylibso/chicory/runtime/TableInstance.java
```diff
@@ -0,0 +1,58 @@
...
+import com.dylibso.chicory.wasm.types.Table;
+import com.dylibso.chicory.wasm.types.Value;
+import com.dylibso.chicory.wasm.types.ValueType;
+import java.util.HashMap;
+import java.util.Map;
+
+public class TableInstance {
+
+    private Table table;
+    private Map<Integer, Instance> instances;
```
andreaTP, line 13: A Table holds function references, but we should invoke the functions on the Instance they belong to.

test-gen-plugin/src/main/java/com/dylibso/chicory/maven/JavaTestGen.java
```diff
@@ -151,27 +144,39 @@ public CompilationUnit generate(
...
-                            new NameExpr(lastModuleVarName + ".instance()"));
+                            new NullLiteralExpr(),
+                            Modifier.Keyword.PUBLIC,
+                            Modifier.Keyword.STATIC);
+
+                    var instantiateMethodName = "instantiate_" + lastInstanceVarName;
+                    var instantiateMethod =
+                            testClass.addMethod(instantiateMethodName, Modifier.Keyword.PUBLIC);
+                    // It needs to be a test to be executed
+                    instantiateMethod.addAnnotation("Test");
```
andreaTP, line 158: this artificially increases the number of tests we are executing, but I have no better way to keep the assertions separated and in order.

test-gen-plugin/src/main/java/com/dylibso/chicory/maven/JavaTestGen.java
```diff
@@ -219,24 +219,13 @@ public CompilationUnit generate(
             }
         }
 
-        if (testClass.getMethods().size() == 0) {
```
andreaTP, line 222: Now that instantiation is executed as a test this is not needed anymore.

andreaTP:
Last but not least, thanks a lot to @evacchi who have been instrumental in getting this to the finish line!

runtime/src/main/java/com/dylibso/chicory/runtime/HostGlobal.java
```diff
@@ -2,9 +2,14 @@
 
 import com.dylibso.chicory.wasm.types.MutabilityType;
 import com.dylibso.chicory.wasm.types.Value;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
 
 public class HostGlobal implements FromHost {
     private Value value;
+    private Supplier<Value> getValue;
+    private Consumer<Value> setValue;
```
mariofusco, line 11: I'm not understanding why this Supplier/Consumer pair is necessary (and I find it a bit confusing) can you please explain? And if that's really necessary can we at least get rid of the fixed value field and replace it with a supplier of that value?

andreaTP: It's not necessary, it's just one possible implementation to support a "shared HostGlobal".
With this design, you can easily implement a "proxy" HostGlobal that defers to some other value.
I'm expecting very little to no usage of this feature in "real world" applications TBH.

But I understand the concern, I'll try again a more "direct" approach and see how it goes 👍

andreaTP: Introduced `GlobalInstance` as an indirection layer to be able to share Globals across Modules.
Let me know if this implementation looks better or feel free to propose an alternative design.

mariofusco (approved):

bhelx:
This is awesome @andreaTP !

fuzz/src/test/java/com/dylibso/chicory/fuzz/SingleReproTest.java
```diff
@@ -32,7 +32,7 @@ void singleReproducer() throws Exception {
                 smith.run(seed.substring(0, Math.min(seed.length(), 32)), "test.wasm", types);
 
         var module = Module.builder(targetWasm).build();
-        var instance = module.instantiate(new HostImports(), false);
+        var instance = module.instantiate(new HostImports(), true, false);
```
bhelx, line 35: @andreaTP Not really important for this PR, but just a comment for later as we're looking at the a 1.0 API. I think booleans as params can get kind of confusing. I blame myself for starting this. Might make more sense to either have differently named instantiate functions or to have some kind of option builder type.

andreaTP: Yes, completely agree, I'll take the time to run some cleanup at some point, and this is something that needs to improve.

PR #280, opened 2024-04-17
Initial JMH setup

andreaTP:
In this PR we introduce a basic setup for running JMH performance tests.

Here: https://github.com/andreaTP/chicory/actions/runs/8721176330 you can find the expected result, but it needs to be merged to `main` to have it working in this repo.

Additionally, we need to enable the [nightly.link](https://nightly.link/) GitHub App to have it fully working.


andreaTP:
Thanks to @franz1981 for the awesome `jmh.morethan.io` JMH visualizer link and inspiration for this development.

franz1981:
FYI https://labs.quansight.org/blog/2021/08/github-actions-benchmarks

In short, if it uses the github machines (not the paid ones) - need to have "several passes averaged together" - usually geometric mean is a nice thing here. Or we risk to die by 1000 noise cut.

What the fuse team has done instead is to run A/B test on the same exact instance thinking that it would have reduce the noise, but IDK yet if the noise happen too by running the same thing on the same machine in a period of time, too

andreaTP:
Thanks @franz1981 for chipping in!
I added the CI setup for essentially 2 reasons:
- have a common way to share benchmark results
- evaluate how much the CI is going to give variable results, I expect major changes are going to be still noticeable

It remains on us to read the results and decide from there, basically just an easily accessible "rule-of-thumb", nothing more.
The "local" setup should cover the rest of the use cases.

PR #281, opened 2024-04-18
finishing `bulk.wast`

andreaTP:
Finishing the last few `bulk.wast` missing tests.

### DIFF
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/Machine.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/Machine.java
@@ -1599,13 +1599,23 @@ class Machine {
 
         var table = instance.table(tableidx);
 
-        if (size < 0
-                || elementidx > instance.elementCount()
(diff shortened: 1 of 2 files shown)


runtime/src/main/java/com/dylibso/chicory/runtime/Machine.java
```diff
@@ -1599,13 +1599,23 @@ private static void TABLE_INIT(MStack stack, Instance instance, long[] operands)
...
+                        || (size > 0
+                                && (currentElement == null
+                                        || !(currentElement instanceof PassiveElement)))
+                        || elemidx + size > currentElementCount
+                        || end > table.size());
+
+        if (isOutOfBounds) {
             throw new WASMRuntimeException("out of bounds table access");
+        } else if (size == 0) {
+            return;
```
bhelx, line 1618: nit: this is a little bit hard to follow and might make sense to break into some variables.

bhelx (approved):
Thanks!

PR #283, opened 2024-04-18
assert_invalid infrastructure

andreaTP:
Opening in draft to show the kind of work that is needed to complete validation.


wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java
```diff
@@ -863,9 +864,51 @@ private static Instruction parseInstruction(ByteBuffer buffer) {
         }
         var operandsArray = new long[operands.size()];
         for (var i = 0; i < operands.size(); i++) operandsArray[i] = operands.get(i);
+        verifyAlignement(op, operandsArray);
```
andreaTP, line 867: This makes all the `align.wast` spec to pass.

danielperano (approved):

runtime/pom.xml
```diff
@@ -145,16 +145,65 @@
...
-            SpecV1LinkingTest.test109, SpecV1LinkingTest.test110
+            <!-- Those tests rely on other tests failures ... -->
+            SpecV1LinkingTest.test83, SpecV1LinkingTest.test86, SpecV1LinkingTest.test94,
+            SpecV1LinkingTest.test97, SpecV1LinkingTest.test115, SpecV1LinkingTest.test116,
+            SpecV1LinkingTest.test118, SpecV1LinkingTest.test121, SpecV1LinkingTest.test122
           </excludedTests>
-          <excludedValidationTests>binary.wast,binary-leb128.wast</excludedValidationTests>
+          <excludedMalformedTests>binary.wast,
+            binary-leb128.wast</excludedMalformedTests>
+          <excludedInvalidTests>block.wast,
```
mariofusco, line 155: I'm not understanding the difference between malformed and invalid tests, can you please explain?

andreaTP: They come straight from the `wast` representation and they are differentiated there, here I'm just preserving the original distinction.

From what I see I can attempt:
- `assert_malformed` -> there is a problem in the binary representation
- `assert_invalid` -> the binary representation is valid but there is a semantic issue e.g. a `block` that should return i32 returns f64 instead

test-gen-plugin/src/main/java/com/dylibso/chicory/maven/JavaTestGen.java
```diff
@@ -200,22 +205,29 @@ public CompilationUnit generate(
                         }
                     }
                     break;
+                case REGISTER:
+                    // should be irrelevant
+                    break;
                 case ASSERT_MALFORMED:
```
bhelx, line 211: What does this mean exactly? Should we log a message here? No strong opinion just curious.

bhelx (approved):
Looks good to me :)

test-gen-plugin/src/main/java/com/dylibso/chicory/maven/JavaTestGen.java
andreaTP: We are not doing anything for the `register` and is the same in wazero.
We still pass the tests 🙂 I'm becoming more confident that this is not useful at all in the way we are running the tests.

.gitignore
```diff
@@ -8,4 +8,4 @@ wasi-testsuite/
 dependency-reduced-pom.xml
 jmh-result.json
 scripts/jmh-tmp
-main/
+/main/
```
andreaTP, line 11: @electrum I understand now the motivation of your change in #288 I think this is good enough for now.

electrum: Yep, this is good, thanks

wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java
```diff
@@ -863,9 +864,51 @@ private static Instruction parseInstruction(ByteBuffer buffer) {
...
         return new Instruction(address, op, operandsArray);
     }
 
+    private static void verifyAlignement(OpCode op, long[] operands) {
```
electrum, line 871: Note there is a typo here, `verifyAlignement` should be named `verifyAlignment`

andreaTP: Ouch, thanks for noticing, I'll fix it when I'm back in front of a PC.

PR #288, opened 2024-04-22
Cleanup .gitignore

electrum:


.gitignore
```diff
@@ -8,4 +8,3 @@ wasi-testsuite/
 dependency-reduced-pom.xml
 jmh-result.json
 scripts/jmh-tmp
-main/
```
andreaTP, line 11: I agree the naming is not ideal but this is the folder where the baseline chicory repository gets cloned for jmh tests, I would keep it in `gitignore`

andreaTP (commented):
I think we should keep `main`, apart from that, good to go 👍 thanks!

.gitignore
andreaTP: it should be `/main/` and is fixed in the `main` branch already.

andreaTP:
@electrum please rebase and it's good to go, thanks 👍

electrum:
I see you changed `main/` to `/main/` so this is good. The previous one was blocking `src/main/java`.

electrum:
I rebased and dropped the second commit.

PR #300, opened 2024-05-03
Refactor the Maven setup

andreaTP:
With this the `aot` testsuite is running but, as expected, it fails.


danielperano (approved):

danielperano:
Beautiful - this is way better (I wanted to merge the test infrastructure into a common test suite, but hadn't gotten that far yet).

PR #301, opened 2024-05-03
BR_IF / BR / BR_TABLE are control frames

### LINKED ISSUE #294: Stack unwinds incorrectly  when jumping from `if` block to the outermost block using a `br_if`
Given the following code:

```wat
(module
  (type $t0 (func (param i32) (result i32)))

  (func $innerFunc (type $t0) (param $arg0 i32) (result i32)
    (local $resultCode i32)
    block $outerBlock
      i32.const -1 ;; These values must be dropped from the stack
      i32.const -2 ;; when the br_if jump is triggered

      i32.const -3
      if (result i32)
        local.get $arg0
        local.tee $resultCode
        br_if $outerBlock
        i32.const 100
      else
        i32.const 200
      end
      local.set $resultCode

      drop
      drop
    end
    local.get $resultCode)

  (func $main (export "main") (type $t0) (param $arg0 i32) (result i32)
    i32.const 0
    local.get $arg0
    call $innerFunc
    i32.add)

  (memory $M0 1))
```

When executing 'main' on Chicory with the argument '5', the function returns the incorrect result of 3.
In the "innerFunc", the  `br_if` jump  to the outer block is triggered and after its execution the values -1 and -2 remain on the stack:
```
stack = {ArrayDeque@2296}  size = 4
 0 = {Value@2311} "5@i32"
 1 = {Value@2312} "-2@i32"
 2 = {Value@2313} "-1@i32"
 3 = {Value@2314} "0@i32"
```

When executed on `wasm-interp` the result is correct:
```shell
$ wasm-interp test.wasm -r "main" -a "i32:5"
main(i32:5) => i32:5
```

andreaTP:
Fix #294 

I confirm that the analysis is correct and this was a bug.

```
wasmtime ./wasm-corpus/src/test/resources/compiled/issue294.wat.wasm --invoke main 5
5
```

Thanks for the great reproducer @illarionov ! :pray:


bhelx:
Thank you for the highly detailed report @illarionov 🎉

wasm-corpus/run.sh
```diff
@@ -4,9 +4,9 @@
 # run with no args to compile everything
 
 if [[ "$1" = "rebuild" ]]; then
-  docker build . --platform linux/amd64 -t chicory/wasm-corpus
+  docker build . --platform linux/amd64 -t docker.io/chicory/wasm-corpus
 else
   # Optionally takes the args `lang` (ex: wat, rust) and `file` (ex: br.wat)
   # both default to all
-  docker run --platform linux/amd64 -v $(pwd)/src/test/resources:/usr/code --rm chicory/wasm-corpus $1 $2
+  docker run --platform linux/amd64 -v $(pwd)/src/test/resources:/usr/code --security-opt label=disable --rm docker.io/chicory/wasm-corpus $1 $2
```
electrum, line 11: Why do we need `--security-opt label=disable`?

bhelx (approved):

PR #302, opened 2024-05-04
Update MStack to use pollFirst() instead of pop()

illarionov:
This fixes the exception thrown on stack underflow.
ArrayDeque#pop throws a NoSuchElementException when stack is empty, whereas pollFirst returns null in such cases.


bhelx:
Good catch, thank you!

bhelx (approved):

PR #309, opened 2024-05-07
Apply spotless during build

mstepan:
1. I propose always formatting code during the build process, this will completely remove any code formatting issues.
2. Added a few additional patterns for IntelliJ IDEA-related files in `.gitignore`
3. Proper java version specified in README(b/c I initially used 17 LTS)
4. Minor update related to 1st build added to README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,6 +246,7 @@ Some nice to have but probably separate items:
 ### Building the Runtime
 
 Contributors and other advanced users may want to build the runtime from source. To do so, you'll need to have Maven installed.
+`Java version 11` required for a proper build. You can download and install [Java 11 Temurin](https://adoptium.net/temurin/releases/?version=11)
 
 Basic steps:
 
--- a/README.md
+++ b/README.md
@@ -246,11 +246,12 @@ Some nice to have but probably separate items:
 ### Building the Runtime
 
 Contributors and other advanced users may want to build the runtime from source. To do so, you'll need to have Maven installed.
-`Java version 11` required for a proper build. You can download and install [Java 11 Temurin](https://adoptium.net/temurin/releases/?version=11)
+`Java version 11+` required for a proper build. You can download and install [Java 11 Temurin](https://adoptium.net/temurin/releases/?version=11)
 
 Basic steps:
 
-* `mvn clean install` to run all of the project's tests and install the library in your local repo
+* For the 1st build you need to create `testsuite` folder manually `mkdir testsuite`
+* `mvn clean install` to run all the project's tests and install the library in your local repo
 * `mvn spotless:apply` to autoformat the code
 * `./scripts/compile-resources.sh` will recompile and regenerate the `resources/compiled` folders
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -246,6 +246,7 @@ Some nice to have but probably separate items:
 ### Building the Runtime
 
 Contributors and other advanced users may want to build the runtime from source. To do so, you'll need to have Maven installed.
+`Java version 11` required for a proper build. You can download and install [Java 11 Temurin](https://adoptium.net/temurin/releases/?version=11)
```
andreaTP, line 249: Have you faced issues with other Java versions?
The requirement should be 11+ I'm currently on 17 and I use 21 sometimes too.

pom.xml
```diff
@@ -364,9 +364,9 @@
         </configuration>
         <executions>
           <execution>
-            <id>format</id>
+            <id>format-sources</id>
             <goals>
-              <goal>check</goal>
+              <goal>apply</goal>
```
andreaTP, line 369: I have mixed feelings here, I see pros and cons in going down this path.
The current setup already prevents committing code that is not formatted, as it will break in CI.

I defer the final decision to someone else in the team.

andreaTP (commented):
@mstepan thanks for your interest in the project!

There are 3 changes in this PR:
- `.gitignore`: completely fine, if you extract to a standalone PR I can merge it immediately
- Java requirement: Java should be 11+ and please, report any other issue you found
- apply fmt on compile by default: I'm split about this change, let see if someone have a stronger opinion

pom.xml
mstepan: Let's wait for 3rd opinion and I will split PR to smaller as needed.

andreaTP: > recently working with Golang team I found it is so natural to auto-format code

I have to appreciate that there is a bit of a gap in-between Go formatting IDE/Editors integration and Java, but I'm fine either way.

bhelx: I don't have a strong opinion either. I'm comfortable with the Go style of doing things. But Go also has a single, simple formatter that ships with it.

I lean towards saying this is positive, though small. I think it's more likely for people to be irritated that their PR fails because of code formatting than they would be irritated that the code gets auto-formatted under their feet.

danielperano (commented):
I hit an issue just yesterday that makes me a bit hesitant to reformat on compile - I had imported a long list of classes from the same package, and IntelliJ helpfully collapsed it down into a star import that we don't allow. Spotless removed the star import but didn't expand it into the list of single imports, breaking compilation. I had to manually re-import everything as single imports (which took a minute of fighting IntelliJ's urge to re-collapse them all again - I really need to figure out how to turn that off!). It was quite annoying, and would be even more so as part of a compile when it might not be expected.

I'm not dead set against it, but due to the differences in tooling (& language, for that matter) between Go & Java I think we might occasionally run into some annoying quirks that wouldn't happen in Go so I'm not really for it either.

mstepan:
Thanks for the feedback. I will remove spotless reformatting and will just leave startup and gitignore files.

mstepan:
Closed in favour of https://github.com/dylibso/chicory/pull/311

Issue #31, opened 2023-10-12
Missing specs


andreaTP:
Editable WIP tracking issue.
Tick the checkbox if you start working on something, remove the line when is done:

- [x] elem.wast
- [x] linking.wast

andreaTP:
updated 👍

andreaTP:
@danielperano ticked `data.wast` as you are on it.

PR #310, opened 2024-05-07
Add elem.wast spec

andreaTP:
Adding almost all of `elem.wast` which is the last wast file missing :slightly_smiling_face:

### DIFF
--- a/runtime/pom.xml
+++ b/runtime/pom.xml
@@ -143,7 +143,6 @@
             <test>SpecV1LinkingTest.test118</test>
             <test>SpecV1LinkingTest.test121</test>
             <test>SpecV1LinkingTest.test122</test>
-            <test>SpecV1ElemTest.test95</test>
           </excludedTests>
           <excludedMalformedWasts>
             <wast>binary-leb128.wast</wast>
--- a/runtime/pom.xml
+++ b/runtime/pom.xml
@@ -58,6 +58,7 @@
             <wast>conversions.wast</wast>
             <wast>custom.wast</wast>
             <wast>data.wast</wast>
+            <wast>elem.wast</wast>
             <wast>endianness.wast</wast>
             <wast>exports.wast</wast>
             <wast>f32.wast</wast>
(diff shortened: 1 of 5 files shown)


runtime/pom.xml
```diff
@@ -142,6 +143,7 @@
             <test>SpecV1LinkingTest.test118</test>
             <test>SpecV1LinkingTest.test121</test>
             <test>SpecV1LinkingTest.test122</test>
+            <test>SpecV1ElemTest.test95</test>
```
andreaTP, line 146: This test is failing because we need to keep the root instance of exported Global FuncRefs.

There is a little design issue to overcome:
- a `FuncRef` should be able to be associated with an `Instance`
- `FuncRef` is defined in the `wasm` module
- `Instance` is defined in the `runtime` module and is not visible in the `wasm` module

For Tables I initially defined `TableInstance` in the `runtime` module, but this approach is falling short here where initializers etc. are involved.

Happy to hear thoughts! cc. @bhelx @danielperano

andreaTP: Resisting hard to break the abstraction layer in between `runtime` and `wasm` modules.
This is a possible implementation that does the job: https://github.com/dylibso/chicory/pull/310/commits/3f5aa57ffffd6ffa4cb77177f1021e2d512878dd

thomasdarimont:
Does this then close https://github.com/dylibso/chicory/issues/31 ?

andreaTP:
Yes! This fixes #31 :tada: 
We need to open another tracking issue for validation now :slightly_smiling_face:

bhelx (approved):
🚀 WOOO

bhelx:
Incredible effort @andreaTP and everyone who participated in implementing these :)

andreaTP:
you started with this @bhelx we are growing your legacy now :-P

bhelx:
I remember when this test suite was just a terrible ruby script 😆

andreaTP:
ruby FTW :+1:

PR #311, opened 2024-05-07
README updated with 1st build instructions. IntelliJ IDEA files added to ignore.

mstepan:
1. README file updated with startup instructions. 
2. Additional Inte…lliJ IDEA *.impl files added to gitignore

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,10 +246,12 @@ Some nice to have but probably separate items:
 ### Building the Runtime
 
 Contributors and other advanced users may want to build the runtime from source. To do so, you'll need to have Maven installed.
+`Java version 11+` required for a proper build. You can download and install [Java 11 Temurin](https://adoptium.net/temurin/releases/?version=11)
 
 Basic steps:
 
-* `mvn clean install` to run all of the project's tests and install the library in your local repo
+* For the 1st build you need to create `testsuite` folder manually `mkdir testsuite`
+* `mvn clean install` to run all the project's tests and install the library in your local repo
 * `mvn spotless:apply` to autoformat the code
 * `./scripts/compile-resources.sh` will recompile and regenerate the `resources/compiled` folders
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -246,10 +246,12 @@ Some nice to have but probably separate items:
 ### Building the Runtime
 
 Contributors and other advanced users may want to build the runtime from source. To do so, you'll need to have Maven installed.
+`Java version 11+` required for a proper build. You can download and install [Java 11 Temurin](https://adoptium.net/temurin/releases/?version=11)
 
 Basic steps:
 
-* `mvn clean install` to run all of the project's tests and install the library in your local repo
+* For the 1st build you need to create `testsuite` folder manually `mkdir testsuite`
```
andreaTP, line 253: This is not correct, the Maven plugin will perform the right operation.

andreaTP (changes requested):
You should not create an empty `testsuite` folder, are you facing any issue?

For the rest LGTM, thanks for making the changes.

README.md
mstepan: If I remove the `testsuite` folder and try to run `mvn clean install` it fails with:
```
[ERROR] Failed to execute goal com.dylibso.chicory:test-gen-plugin:999-SNAPSHOT:wasm-test-gen (default) on project runtime: Failed to list wast files in ~/chicory/testsuite -> [Help 1]
```

andreaTP: Thanks for reporting! This is a bug/regression.
Fixed in: #313

andreaTP (approved):
LGTM, thanks for making the changes!

mstepan:
I believe it's expected that PR will be merged by someone with the WRITE permission. I don't have WRITE permission, so can't merge current PR.

andreaTP:
Correct, as a general rule of thumb, I try to leave a some time to other maintainers for additional reviews or comments.
In this case the risk is extremely low and I merge it now :+1: 
Thanks for your help @mstepan

PR #312, opened 2024-05-07
Experimental AOT Compiler

danielperano:
This PR provides the minimum infrastructure to compile and run the WASM I32 spec (`i32.wast`).

I propose several features for AOT compilation which this PR demonstrates:
1. When a WASM opcode cannot be implemented as a single JVM instruction, the implementation shall be placed in a static method in a class that is shared by both the interpreter and AOT implementations.
2. The AOT shall cleanly separate generators for opcode implementations wherever possible. An interpreter-style switch loop may be used for traversal of the bytecode structures (blocks), but should not be used for anything that can go into a standalone generator and still maintain clarity of intent.
3. The AOT should handle calls to static, shared opcode implementations automatically, except for registration/discovery of the opcode implementations.
4. AOT'd code should be accessible via the same API as interpreted code. In the future, full compilation to standalone `.class` files (or compilation to runtime-loaded classes without writing to files) should be supported as an optional feature. These standalone classes should expose their API in a manner consistent with their use in native code, rather than requiring use of a Chicory API.


aot/src/main/java/com/dylibso/chicory/aot/AotEmitters.java
```diff
@@ -0,0 +1,147 @@
...
+
+    public static void I32_SUB(AotContext ctx, Instruction ins, MethodVisitor asm) {
+        asm.visitInsn(Opcodes.ISUB);
+    }
+
+    public static void I32_XOR(AotContext ctx, Instruction ins, MethodVisitor asm) {
+        asm.visitInsn(Opcodes.IXOR);
+    }
+
+    public static BytecodeEmitter intrinsify(OpCode opcode, Class<?> staticHelpers) {
```
bhelx, line 123: I think you explained it well in the meeting this week, but could we get some inline docs about what inftinsifying is in this context? any added docs on some of the meatier pieces will be helpful as people start to contribute or go to fix bugs.

danielperano: I had somewhat documented these concepts in a couple other scattered areas, but added some documentation here as well.

runtime/src/main/java/com/dylibso/chicory/runtime/MachineUtil.java
```diff
@@ -0,0 +1,89 @@
+package com.dylibso.chicory.runtime;
+
+import com.dylibso.chicory.wasm.exceptions.ChicoryException;
+import com.dylibso.chicory.wasm.types.Instruction;
+import com.dylibso.chicory.wasm.types.Value;
+import com.dylibso.chicory.wasm.types.ValueType;
+import java.util.Arrays;
+import java.util.List;
+
+public class MachineUtil {
```
andreaTP, line 10: I'm not great with names, but wdyt about `MachineConstants` or `CompileTimeConstants`(and change the methods to e.g. `computeValue`), just to show the purpose of the class instead of using "Util".

runtime/src/main/java/com/dylibso/chicory/runtime/Module.java
```diff
@@ -106,11 +107,37 @@ public Instance instantiate() {
...
+    public Instance instantiate(HostImports hostImports) {
+        return this.instantiate(hostImports, true, true);
+    }
+
+    public Instance instantiate(
+            HostImports hostImports, Function<Instance, Machine> machineFactory) {
+        return this.instantiate(hostImports, machineFactory, true, true, null);
+    }
+
+    public Instance instantiate(
```
andreaTP, line 126: I recently cleaned up a little this class to make it mandatory to use the Builder Pattern.
This should be `withMachineFactory` and remove the other `instantiate` overloads.

andreaTP (approved):
A few nitpicks, but looks great work to me!
You are free to:
- merge immediately and we will follow up
- ask me to address the issues and we merge after
- address the comments and merge at your convenience

PR #313, opened 2024-05-07
Fix the testsuite download order in the test-gen plugin

andreaTP:
Reported here: https://github.com/dylibso/chicory/pull/311#discussion_r1592728957


electrum (approved):
Thanks for fixing this. I missed this case when testing.

andreaTP:
Thanks for checking in @electrum :+1:

No worries the code is pretty maintainable.
For context in this issue:

We are not testing this aspect in CI for two reasons:
- avoid rate limiting on the GH Api
- we expect to catch possible regressions early

One possible mitigation is to run this developer flow nightly... more a nice to have than necessary from my pov.

bhelx (approved):

PR #314, opened 2024-05-07
PoC for JSR223 script-engine

### LINKED ISSUE #149: Implement a compatibility layer for JSR 223
https://docs.oracle.com/javase/8/docs/technotes/guides/scripting/prog_guide/api.html

thomasdarimont:
Initial attempt for implementing a JSR223 compatible script-engine based on chicory.

Fixes #149

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -43,6 +43,7 @@
     <module>wasi</module>
     <module>wasm-corpus</module>
     <module>cli</module>
+    <module>scriptengine</module>
   </modules>
 
   <scm>
--- a/scriptengine/pom.xml
+++ b/scriptengine/pom.xml
@@ -1,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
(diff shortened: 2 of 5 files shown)


thomasdarimont:
Hi @andreaTP unfortunately, I don't have time to work on this at the moment.

I still think it would be useful to have wasm support per ScriptEngine I had the impression that the JSR223 API is not meant to be used with "binary" scripts as most ways to supply a script only offer string based methods. That's why I resorted to provide the wasm as base64 encoded string.

Of course we could also support WAT here, but that might be quite inefficient.

While thinking about it, another (crazy) idea came to my mind:

We could just add wasm support to Nashorn via chicory :)


We could provide a small JS wrapper that provides the minimal JavaScript web assembly API core which uses chicory underneath to load and execute the webasebly. 
See: https://developer.mozilla.org/en-US/docs/WebAssembly/Using_the_JavaScript_API

This would allow us to leverage the ScriptEngine Bindings support of Nashorn and would work nicely with the scripting support in existing Frameworks :)

andreaTP:
Thanks for getting back @thomasdarimont !
Even if we compromise on performance, maybe having Wat support in the script engine is the most compelling integration.

Talking about the other idea on Nashorn, totally, I have been exploring it in Rhino: https://github.com/rhino/rhino-wasm/pull/1
but it requires more work than I anticipated (the API surface is big enough) and it's up for grabs if anyone wants.

andreaTP:
Let me close this PR for now, @thomasdarimont feel free to re-open it or open an alternative one.

I'm fully supportive in making Chicory available through JSR223 by embedding it(or having it as an optional extension) in JS runtimes like Nashorn and Rhino.

At the moment I don't have the bandwidth and a clear use case to push this forward myself either, but if anyone reading this thread is interested I'll be happy to assist.

PR #318, opened 2024-05-08
Enable the implementation of optional runtime monitoring

andreaTP:
Alternative implementation for #308


bhelx (approved):

andreaTP:
Merged #308 instead.

PR #295, opened 2024-04-29
docs: Add roadmap, update READMEs and contributing guide

bhelx:
I had some time this weekend to start working on readmes and add a roadmap. Here are my notes from the meeting where we identified this problem:

> the primary theme that kept emerging is we need to be more transparent about our roadmap and our motivations. we have a lot of ideas and features that we have signaled interest in, but it's time to put out clear signals about what we're aiming for this year, what would be nice to have community support for, and what we are not going to do. I'm taking it as my task to update the readme and get this started. 

I also took the time to add a readme to the wasi package. Mainly to document how it works but also document our plans for it. 

I'm at a conference this week but should be able to continue improving this through the week. Feel free to drop comments or ideas in this draft state.

PR #319, opened 2024-05-08
MStack internals changed.

mstepan:
1. `MStack` class internals changed from ArrayDeque usage to hand-crafter Stack using an array.
2. Additional unit tests added.
3. Minor improvements in JMH scripts.

The current JMH benchmark doesn't show any improvements (no changes), but that's because `MStack` is not used extensively during micro-benchmark tests.

I want to receive initial feedback if it makes sense or not.

### DIFF
--- a/scripts/build-jmh-main.sh
+++ b/scripts/build-jmh-main.sh
@@ -1,10 +1,11 @@
-#! /bin/bash
+#!/usr/bin/env bash
 set -euxo pipefail
 
 SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 
 (
-  cd "${SCRIPT_DIR}/.."
-  git clone --depth 1 --branch main https://github.com/dylibso/chicory.git main
-  mvn -Pbenchmarks spotless:apply clean package -DskipTests
+  cd "${SCRIPT_DIR}/.." || exit 1
+  rm -rf main
+  git clone --depth 1 --branch main https://github.com/dylibso/chicory.git main || exit 1
+  mvn -Pbenchmarks spotless:apply clean package -DskipTests || exit 1
 )
--- a/scripts/run-jmh.sh
+++ b/scripts/run-jmh.sh
@@ -1,6 +1,6 @@
-#! /bin/bash
+#!/usr/bin/env bash
 set -euxo pipefail
 
 SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 
-java -jar "${SCRIPT_DIR}/../jmh/target/benchmarks.jar" -rf json
+java -jar "${SCRIPT_DIR}/../jmh/target/benchmarks.jar" -rf json || exit 1
--- a/jmh/README.md
+++ b/jmh/README.md
@@ -6,7 +6,7 @@ This module is used to run JMH performance tests on Chicory and make it easy to 
 
 There a few additional tools that are needed to see the benchmark results:
 
-- `wget`
+- [wget](https://www.gnu.org/software/wget/)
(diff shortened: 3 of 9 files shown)


mstepan:
Screenshot 2024-05-08 at 16 24 34

andreaTP:
Thanks for pushing on this side too @mstepan !
My takes:

- this implementation is slightly more complicated than the current implementation, if there is no performance improvement, we should keep the simpler implementation ( [ref to the upcoming docs](https://github.com/dylibso/chicory/pull/295/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055R20) )
- the JMH benchmarks are immature, maybe tweaking the parameters and/or implementing a new test to stress the Stack you can get better/more representative results
- seems that [in CI](https://github.com/dylibso/chicory/actions/runs/9002634918/attempts/2#summary-24731845290) we have a performance degradation with those changes (this data needs to be validated locally)
- [here](https://github.com/dylibso/chicory/pull/251/files#diff-3dc68b619be666678282bd7ddf1a8fc699b29104bb78a4fad308ee4172b2e833) you can find a past attempt to improve performance you might find interesting

Last, but not least, let me try to summon my favorite performance experts @franz1981 @mariofusco :pray:

mariofusco:
My personal point of view on this is that we're (over)complicating the implementation only for a theoretical performance gain that at moment is totally not demonstrated. Unless we have a few JMH benchmarks clearly evidencing the improvement, I'm quite against this change.

mstepan:
I completely agree that there should be clear evidence to merge any improvements. Let's keep this PR open for a while, I will try to play with the codebase and JMH and see if more interesting results can be demonstrated.

scripts/run-jmh.sh
```diff
@@ -1,6 +1,6 @@
-#! /bin/bash
+#!/usr/bin/env bash
```
bhelx, line 1: It might be smoother to move some of these script improvements to their own PR. I don't normally mind drive by changes but i suspect there will be some discussion around the main purpose of the PR and don't want these improvements to sit unused

bhelx:
I have a similar feeling as the others. I really appreciate the attention to detail and desire to improve this! I support changing the name for instance (FWIW i named it MStack and deeply regret it 😆 ).

But ultimately, in this phase of the project, we're trying to stick with built in java implementations as much as possible unless there is a massive performance increase or a big usability improvement. Whereas this change feels like it could be a later stage change as things mature a bit and getting a little performance increase is worth adding the complexity. Because the cost of the added complexity goes down and the cost of getting performance gains goes up as the project matures.

andreaTP:
> Let's keep this PR open for a while

No problem at all, I convert it to Draft and label as `wip`, take your time and thanks for the exploration!

mstepan:
Partially work move to https://github.com/dylibso/chicory/pull/322

scripts/build-jmh-main.sh
```diff
@@ -1,10 +1,11 @@
...
+#!/usr/bin/env bash
 set -euxo pipefail
 
 SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 
 (
-  cd "${SCRIPT_DIR}/.."
-  git clone --depth 1 --branch main https://github.com/dylibso/chicory.git main
-  mvn -Pbenchmarks spotless:apply clean package -DskipTests
+  cd "${SCRIPT_DIR}/.." || exit 1
```
electrum, line 7: The `set -e` above does this automatically:

> set: set [--abefhkmnptuvxBCHP] [-o option] [arg ...]
>     -e  Exit immediately if a command exits with a non-zero status.

mstepan:
We conclude that this PR is not needed (at least no obvious benefits right now).

PR #320, opened 2024-05-09
[FUZZ] Crash reproducer

bhelx:
Fuzz testing failed and there are new reproducers available.


andreaTP:
Seems like the nightly CI is a bit more stable lately.
Let's postpone looking into those.

PR #321, opened 2024-05-09
Adds wasi func name to missing exception

bhelx:
This is a small change that makes this a little easier to debug when you hit one of these unsupported WASI calls.


andreaTP (approved):
lgtm, thanks!

PR #322, opened 2024-05-09
Minor improvements in JMH scripts

mstepan:
1. JMH scrips improved. Shebang added and https://github.com/koalaman/shellcheck/wiki/SC2086


runtime/src/test/java/com/dylibso/chicory/runtime/MStackTest.java
```diff
@@ -0,0 +1,91 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import com.dylibso.chicory.wasm.types.Value;
+import org.junit.jupiter.api.Test;
+
+public class MStackTest {
```
andreaTP, line 11: @bhelx you mentioned to regret calling this `MStack` do you have an alternative name proposal?

andreaTP (changes requested):
Thanks for this contribution @mstepan !

I'm fine with all the scripts improvements, thanks :+1: 
I would leave the test for the `MStack` out as, with the current implementation, we will be testing the Java standard library.

runtime/src/test/java/com/dylibso/chicory/runtime/MStackTest.java
mstepan: `OperandStack` is the proper name IMHO. At least this will clearly distinguish it from CallStack.

mstepan: `MStackTest` is kinda of testing some minor logic inside the MStack, like throwing appropriate exception etc. It would be much easier to replace in future if needed, b/c each time running the whole integration tests suite is rather time-consuming.

bhelx: Not specifically. Something like `ValueStack` or `OperandStack` both make sense to me because that's kind of how we use it at the moment. In the spec, the stack is used for more than operands though. But I'm not sure if we want to put labels and activations on the stack so maybe it will always only be values for us. https://www.w3.org/TR/wasm-core-1/#stack

What do you think @andreaTP ? Naming is obviously one of my weak spots :)

andreaTP: > like throwing appropriate exception etc.

This is covered by the testsuite.

> It would be much easier to replace in future if needed

Happy to re-evaluate when necessary, in the future, we don't want to introduce complexity and code to be maintained if it's not useful with the current implementation.

> Naming is obviously one of my weak spots

Same here :sweat_smile: what about `WasmStack`?
No super-strong opinions, though!

bhelx: IDK WasmStack feels a bit too generic for me. It does feel like right now it is a stack for holding operands. thus `OperandStack` makes sense. I recommended `ValueStack` because it holds `Value`s (both the chicory type and the wasm type). So it feels closer to the code. Although we instructions do have operands then name of the property is `operands()`, they manifest in the code as `Value`s.

I don't like MStack because it doesn't mean anything. I literally just didn't want to conflict with `Stack` which i was using at the time 😆 . I think in my head it was `M(achine)Stack`.

bhelx: Okay since we're kind of not coming to a conclusion and we don't have strong opinions, i'm just gonna say `OperandStack` is good for us

runtime/src/test/java/com/dylibso/chicory/runtime/OperandStackTest.java
```diff
@@ -0,0 +1,86 @@
...
+        assertEquals(0, stack.size());
+
+        stack.push(Value.i32(5));
+        stack.push(Value.i32(10));
+
+        assertEquals(2, stack.size());
+    }
+
+    @Test
+    public void pushWithResizing() {
```
bhelx, line 24: taking a quick look at this test, It feels like it's testing `ArrayDeque` and not necessarily our code. I think we can keep it scoped to error the conditions for now or whatever is in our impl of OperandStack.

runtime/src/test/java/com/dylibso/chicory/runtime/MStackTest.java
andreaTP: Not fussy, but `Operands` conflicts with the notion of "immediate values" that we call `Operands`.
If we call the Stack `OperandStack` I suggest we rename `operands` used in the machine to something different.

To be honest, `M(achine)Stack` looks like the best option to me.

mstepan:
All comments addressed + rebased.

andreaTP (approved):
LGTM, thanks as always!

PR #325, opened 2024-05-10
Add Bill of Materials (BOM) artifact

electrum:
This makes it easier to use and ensures that the same version is used for all artifacts:
```xml
<dependencyManagement>
  <dependencies>
    <dependency>
      <groupId>com.dylibso.chicory</groupId>
      <artifactId>bom</artifactId>
      <version>XXX</version>
      <type>pom</type>
      <scope>import</scope>
    </dependency>
  </dependencies>
</dependencyManagement>

<!-- no need to specify versions for the artifacts -->
<dependencies>
  <dependency>
    <groupId>com.dylibso.chicory</groupId>
    <artifactId>log</artifactId>
  </dependency>
  <dependency>
    <groupId>com.dylibso.chicory</groupId>
    <artifactId>runtime</artifactId>
  </dependency>
  <dependency>
    <groupId>com.dylibso.chicory</groupId>
    <artifactId>wasi</artifactId>
  </dependency>
  <dependency>
    <groupId>com.dylibso.chicory</groupId>
    <artifactId>wasm</artifactId>
  </dependency>
</dependencies>

PR #327, opened 2024-05-10
Generate a BOM with `sundrio`

andreaTP:
No super strong opinions, but for this mundane tasks we can make use of some magic.
Inspiration from the [Fabric8 Kubernetes Client](https://github.com/fabric8io/kubernetes-client/blob/b295118c4cdc3fb4f6a83029d9543edde899d4ac/pom.xml#L1530).
Alternative to #325

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -26,6 +26,7 @@
 
   <developers>
     <developer>
+      <id>chicorians</id>
       <name>The Chicory Authors</name>
       <email>oss@dylibso.com</email>
       <organization>Dylibso, Inc.</organization>
@@ -83,6 +84,7 @@
     <commons-lang.version>3.14.0</commons-lang.version>
     <commons-io.version>2.16.1</commons-io.version>
     <jmh.version>1.37</jmh.version>
+    <sundrio.version>0.200.0</sundrio.version>
 
     <maven-plugin-api.version>3.9.6</maven-plugin-api.version>
     <maven-plugin-annotations.version>3.13.0</maven-plugin-annotations.version>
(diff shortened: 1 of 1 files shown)


andreaTP:
Cannot quickly get the spotless exclusion to work to pass the CI, will look deeper into if we decide to move on with this.

electrum:
This is cool, and I like that it's automatic, but I have some concerns about this plugin. It generates the module and then builds it inline during execution of the goal (which is why Spotless fails). This is an interesting technique, but I've never seen a plugin that does this, so I wonder if this can cause any problems. When running with `mvn -T` or [Maven Daemon](https://github.com/apache/maven-mvnd) (which I always use), it prints this warning:
```
[WARNING] *****************************************************************
[WARNING] * Your build is requesting parallel execution, but this         *
[WARNING] * project contains the following plugin(s) that have goals not  *
[WARNING] * marked as thread-safe to support parallel execution.          *
[WARNING] * While this /may/ work fine, please look for plugin updates    *
[WARNING] * and/or request plugins be made thread-safe.                   *
[WARNING] * If reporting an issue, report it against the plugin in        *
[WARNING] * question, not against Apache Maven.                           *
[WARNING] *****************************************************************
[WARNING] The following plugins are not marked as thread-safe in Chicory:
[WARNING]   io.sundr:sundr-maven-plugin:0.200.0
[WARNING] 
[WARNING] Enable debug to see precisely which goals are not marked as thread-safe.
[WARNING] *****************************************************************
```
Looking at the plugin code, it's actually not thread safe:
```java
private static final Set<String> GENERATED_ARTIFACT_IDS = Collections.synchronizedSet(new HashSet<String>());
```
Another consideration is that the generated BOM contains all modules, including the internal Maven plugins. I intentionally only included the public modules in the hand-generated BOM.

Given these issues, the hand-written BOM seems preferable. It's simple, reliable and easy to understand. I expect that adding modules will be rare, so adding an entries to the BOM should not be a burden.

PR #328, opened 2024-05-10
[Do not merge] Text format with wat2wasm

andreaTP:
Warning, this PR horribly breaks the execution of tests from the IDE (IntelliJ) with the following error:

> annotation processing is not supported for module cycles. Please ensure that all modules from cycle [wat2wasm,runtime,wasi] are excluded from annotation processing

But it provides a viable solution to parsing wasm text format :slightly_smiling_face: .

The dots:
- `wabt` ships the binaries compiled to `wasm` [here](https://github.com/WebAssembly/wabt/releases/download/1.0.34/wabt-1.0.34-wasi.tar.gz)
- `wat2wasm` read a file and transform it to it's corresponding binary format
- @electrum contributed some amazing work supporting FS operations in `wasi`

Wrapping things together:
- added a module that depends on the latest published version of Chicory and shade the dependency in a uber-jar
- the module contains the wasm version of `wat2wasm` and a little wrapper to work with `jimfs`
- we are using the wrapped dependency to parse the `wat` files generated by the testsuite
- this way we can actually run the tests targeting `wat` modules preprocessing using `wat2wasm`

The error mapping is hacked around, but testing it is pointless as we will be checking `wat2wasm` instead of Chicory.
The biggest issue is the compatibility with IntelliJ.


andreaTP:
Seems like the latest iteration works nicely in the IDE too (after deleting all the caches and restarting).

Next steps:

- verify it doesn't break the IDE experience for more users
- evaluate the tradeoff: +571 tests would not be disabled, but the tests' execution is taking about 30 seconds more

thomasdarimont:
Just checked out you branch and ran the tests in the runtime module. Works fine for me in intellij.

wat2wasm/src/main/java/com/dylibso/chicory/wat2wasm/Wat2Wasm.java
```diff
@@ -0,0 +1,73 @@
...
+import java.nio.file.FileSystem;
+import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
+import java.util.List;
+
+public class Wat2Wasm {
+    private static final Logger logger = new SystemLogger();
+
+    public static byte[] parse(File file) {
+        var module = Module.builder("wat2wasm").build();
```
bhelx, line 30: Would it be possible to memoize this? it shouldn't change over the execution of the program.

andreaTP:
@bhelx rebased and ready for final review.

andreaTP:
I try to take advantage of you being around @bhelx and @danielperano :innocent: 
I think this has been discussed already and we can move on with it, wdyt?

danielperano (approved):

danielperano:
I think this is a brilliant idea because it gets us built-in `.wat` support that is essentially free.

bhelx (approved):
Yes, I think running tests a little slower is okay. In a way this might help us see the relative speed changes better :)

andreaTP:
A good motivation for someone to start to look into low hanging performance improvements, maybe :innocent: cc. @mariofusco @enebo

PR #329, opened 2024-05-12
Some small improvements for bindings authors

bhelx:
When interfacing with a libsqlite, i realized we don't really expose some of the types and interfaces that would make interoping with C simpler. This doesn't really address that but just adds some improvements around how we expose reading and writing strings (and adds support for c strings).

Perhaps a longer term solution would be to provide some wrapper interfaces that can wrap the memory. It would be nice to do things like, return a pointer which has these reader and writer methods on it. Like an FFI interface. But i'm not sure that belongs in the Memory class and I think that's a broader tasks. It could get bloated.

### DIFF
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/Memory.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/Memory.java
@@ -11,6 +11,7 @@ import com.dylibso.chicory.wasm.types.PassiveDataSegment;
 import com.dylibso.chicory.wasm.types.Value;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
+import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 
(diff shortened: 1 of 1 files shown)


runtime/src/main/java/com/dylibso/chicory/runtime/Memory.java
```diff
@@ -117,12 +118,40 @@ public void initPassiveSegment(int segmentId, int dest, int offset, int size) {
...
+    }
+
+    public void writeCString(int offset, String str, Charset charSet) {
+        writeString(offset, str + '\0', charSet);
+    }
+
+    public String readCString(int addr, Charset charSet) {
+        int c = addr;
+        while (read(c) != '\0') {
+            c++;
```
bhelx, line 148: perhaps a better way to do this? not sure whether it's more efficient to also be reading in the string in this loop or only get the count.

runtime/src/main/java/com/dylibso/chicory/runtime/Memory.java
```diff
@@ -117,12 +118,40 @@ public void initPassiveSegment(int segmentId, int dest, int offset, int size) {
...
-        write(offset, data.getBytes(StandardCharsets.UTF_8));
+        writeString(offset, data, StandardCharsets.UTF_8);
     }
 
     public String readString(int addr, int len) {
-        return new String(readBytes(addr, len), StandardCharsets.UTF_8);
+        return readString(addr, len, StandardCharsets.UTF_8);
+    }
+
+    public String readString(int addr, int len, Charset charSet) {
```
bhelx, line 133: was kind of going back and forth on it. On the one hand, we could let the caller handle some of this charset stuff. But then we'd just be exposing a bytes interface which i think is a little too low level. Granted, maybe we'll take this stuff out eventually and put in some kind of bindings package.

andreaTP (approved):
A bit of mixed feelings here.
On one hand, those are just little helpers to make the life of a binding authors suck a little less.
At the same time, we have always been thinking about a "higher level API", is it maybe time to start a separate module?

All in all, I'm good with the changes, let see where this goes.

bhelx:
>  At the same time, we have always been thinking about a "higher level API", is it maybe time to start a separate module?

I agree. This is just a temporary add. I think it's time to create a bindings package of some kind. I'll write up an issue for it and we can start working on it.

PR #330, opened 2024-05-13
Test the README.md file in CI

andreaTP:
I keep forgetting to update the examples that we have in the Readme file.
Thanks to `jbang` we can actually test it :slightly_smiling_face: so that we can catch and fix the documentation along with the implementation.


bhelx (approved):
I'm surprised this works! Looks good :)

PR #332, opened 2024-05-13
AOT tidy-up

danielperano:


andreaTP (approved):
Sorry for the delay, thanks for the follow up @danielperano ! LGTM

Issue #224, opened 2024-02-04
Text format support for a Java WebAssembly runtime


kanha-gupta:
The task is to construct a Parser which converts Text format WASM Module and Generating output through AST using Domain classes already available in the project.
ANTLR is a suggested Parsing framework for the task.

kanha-gupta:
Hi @andreaTP, as you mentioned that we can either enable core parser functionalities first or enable custom parsers.
What do you think is the right approach ?

andreaTP:
No super strong opinion as far as we can get this effort started.

A possible approach would be to extract an `interface` for the [Parser](https://github.com/dylibso/chicory/blob/59bdb0dab7017f4dd778ecf0d054e6e3c2ce6d50/wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java#L63) functionality.
The `BinaryParser` implementation will be the current implementation, and we should make the `Parser` instance injectable [here](https://github.com/dylibso/chicory/blob/59bdb0dab7017f4dd778ecf0d054e6e3c2ce6d50/runtime/src/main/java/com/dylibso/chicory/runtime/Module.java#L489).

In this way implementing new Parsers should be straightforward.
wdyt? 

A couple of notes:
- the Parser is not supposed to "generate output" but rather create a valid instance of a [Module](https://github.com/dylibso/chicory/blob/main/wasm/src/main/java/com/dylibso/chicory/wasm/Module.java).
- I suggested ANTLR because the generated code doesn't have dependencies and that's something we value in this project

cc. @mariofusco

bhelx:
I'm fine with antlr if @andreaTP is. Keep in mind that WAT is basically just s-expressions (with some caveats) so the parser really should be the easier part if you want to just do it by hand. But if Antlr gives us a better experience that sounds good to me too.

One thing I do want to note as a possible conflict is https://github.com/dylibso/chicory/pull/208 as we're kind of refactoring some of the core wasm encoder and decoder layer. It might not be a problem, but maybe something we should talk about in the team meeting tomorrow.

kanha-gupta:
@andreaTP @bhelx I will come up with a proposal soon. I will attach Screenshots of what is working best in my local environment too. 
Thanks a lot for your guidance :)

electrum:
Note that ANTLR generated code depends on `antlr4-runtime` which has a strong dependency on the ANTLR generator version, so you would probably want to shade it.

andreaTP:
[...]

PR #337, opened 2024-05-16
Using wabt compiled to wasm

andreaTP:
Here we move away from using `wabt` tools from the command line using a `ProcessBuilder` and we run the tools using Chicory itself.
Meanwhile, I updated `wabt` to 1.0.35.

A couple of observations:

- `wasm-objdump` fails in IntelliJ for me, looks like a problem with the classpath, I would be ok to remove its execution now that we pass the entire `testsuite` but open to suggestions
- performance are slightly worst as expected, it might be a motivation to start looking at them sooner, and they should not be completely terrible

Happy to receive early feedback!


andreaTP:
The execution time seems to be about the same in CI! That's good news!

danielperano (approved):

andreaTP:
I just found out that even `wat2wasm` fails to be executed in IntelliJ.
Probably not a big issue but let me know if this hurts someone workflow.

I simplified a bunch of things and done some cleanup.

It would be great if @electrum can review the `wabt` module setup with shading and inclusions/exclusions :pray:

andreaTP:
Additional note that this is much slower on my machine, likely the execution of the external process is better parallelized.
The difference is not observable in CI, probably, because there are only 2 cores.

electrum:
I strongly advise against this shading approach. Using a shaded module as a dependency for another module within the same reactor build doesn't work as expected, because Maven doesn't understand shading and won't use the dependency-reduced POM for the dependent modules. Similarly, IntelliJ won't work with this, either. The exclusion hacks might make the Maven build work, but won't work with IntelliJ, and it feels fragile and hard to understand.

The only reliable approach that I've found is to do shading in a separate project. So you could have a `chiocry-wabt` project that depends on the latest release and produces a shaded artifact with a dependency-reduced POM. Then `chicory` would depend on this artifact, without the need for exclusions. We have several such sidecar shaded artifact projects for Trino. While it's annoying to manage these separate projects, they don't change frequently, and it brings the benefit of being completely reliable for the main project.

Alternatively, since this seems to only be used for tests, is to move the tests into a separate integration test module. This has the downside of the tests not being in the same module as the code, but is reliable and avoids all of the above problems. It would also allow running with the current version of Chicory, which is good for testing purposes, and will be important if a newer version of those tools requires a bug fix in Chicory (if using a release version, the tool couldn't be upgraded until a new version of Chicory is released).

electrum:
I updated your PR in #341 that splits out the tests into separate modules. The first commit is a squash of all commits from this PR, then another commit with my changes. Though it's probably easier to look at the entire thing as one diff. Feel free to squash my commit as this is mostly your work.

andreaTP:
Thanks a lot @electrum for your intervention and advice here!

I noticed that you went ahead and already proactively implemented the last option(effectively the only one that you could tackle on your own), this is excellent but slightly increases the complexity of this setup further.

> The only reliable approach that I've found is to do shading in a separate project. So you could have a chiocry-wabt project

I think that this proposal is still the cleanest/most desirable, but we need @bhelx buy in to proceed or we have to host the additional repository on a personal account.

electrum:
Having a separate repository that uses a shaded version the previous release seems more complicated than splitting out a `runtime-tests` module to run the WASM testsuite.

Shading and using the previous release version to avoid a circular dependency is an ugly workaround — it’s not a clean solution. It seems like a last resort when you have no better options.

If the concern here is being able to use [text format](https://github.com/dylibso/chicory/issues/224) — which isn’t an issue for users of the library as they won’t have a circular dependency — then we can just write the parser. I can look at that later today.

andreaTP:
Ok, given your answer here and in the other PR, I think my mind is formed.
Let's move on with your fixes in #341 and iterate when necessary.

Thanks for bearing with me @electrum

PR #339, opened 2024-05-17
Limit MAX_PAGES based on maximum JVM array size

electrum:
Chicory implements memory using ByteBuffers which are backed by byte arrays, and arrays are limited to slightly less than 2^31 values.

### DIFF
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/Memory.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/Memory.java
@@ -1,6 +1,7 @@
 package com.dylibso.chicory.runtime;
 
 import static com.dylibso.chicory.runtime.ConstantEvaluators.computeConstantValue;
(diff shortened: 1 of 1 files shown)


electrum:
Thanks, fixed. I should have ran the tests before pushing 🤦

The test failures made me realize that `MemoryLimits` is the wrong place for this, as that represents the parsed WASM model. I moved the limit to `runtime.Memory`, since that handles the runtime implementation. I included your calculation in the comment, since that seems like a good explanation of what we're doing here. I also changed both `MAX_PAGES` and `PAGE_SIZE` to be literals, as that seems more direct and easier to read.

Also, I added a check to ensure that we're not shrinking the memory size. I can do that as a separate commit or PR if you feel that is better.

andreaTP (approved):
Lgtm, thanks a lot for noticing and fixing this!

Issue #335, opened 2024-05-15
Enable all `assert_invalid` tests


andreaTP:
Opening this issue to gather as much help as possible from the community :slightly_smiling_face: 

`asset_invalid` tests are, mostly, exercising the static type checking capabilities of the Wasm runtime.
Those tests are important to verify that we will not be trying to execute invalid modules and, ultimately, hitting undefined behaviors.
They will make Chicory much more safe!

The section of the spec where you find those rules is [here](https://webassembly.github.io/spec/core/valid/index.html).
We started the implementation following the steps of wazero which is performing this validation, mostly, in the [func_validation.go](https://github.com/tetratelabs/wazero/blob/5a8a053bff0ae795b264de9672016745cb842070/internal/wasm/func_validation.go) file.
In this codebase we started implementing the same checks in the [`TypeValidator`](https://github.com/dylibso/chicory/blob/main/runtime/src/main/java/com/dylibso/chicory/runtime/TypeValidator.java).
It's structured as a stack based interpreter like the actual runtime but, instead of checking the values, it verifies that the types are matching i.e. if you load an `I32` on the Stack and the subsequent instruction is supposed to pop an `F64` we have to throw an exception accordingly.

The process to work on this issue is iterative:

1. remove/comment one `wast` file name from this list in the [`runtime/pom.xml`](https://github.com/dylibso/chicory/blob/0ee666206d39ffe45ba6d4f8e8ae50d4d42179d1/runtime/pom.xml#L155) (the complexity of tackling those vary a lot, check a different file if you are getting stuck too much or the required code increases a lot)
1. run a full `mvn spotless:apply clean install`
1. check the failures as they are, likely, going to be related with the type validation logic
1. fix the tests! If enabling an entire `wast` file turns out to be a lot of work, you can disable specific tests adding the names in [this section of the pom.xml](https://github.com/dylibso/chicory/blob/0ee666206d39ffe45ba6d4f8e8ae50d4d42179d1/runtime/pom.xml#L140)
1. submit a PR! :tada: 

We will be grateful of any contribution!
If you start to work on something, you can mention the file/files you are working on with a comment on this Issue, to decrease the chances that we spend duplicated effort.

thomasdarimont:
Thanks @andreaTP for this very useful writeup :)

I just gave this a spin and I'm already stuck :-/. Some guidance would really help to get into gears with this. 
 I tried to enable ` <wast>br.wast</wast>` and for this I need to implement validation for the BR op code.

[...]

PR #340, opened 2024-05-17
Add some memory assert_invalid and dummy control flow

andreaTP:
Partially targets #335


runtime/src/main/java/com/dylibso/chicory/runtime/Instance.java
```diff
@@ -217,6 +217,10 @@ public Instance initialize(boolean start) {
             }
         }
 
+        if (start && module.export(START_FUNCTION_NAME) != null) {
```
bhelx, line 220: Can you explain this change?

runtime-tests/pom.xml
```diff
@@ -171,6 +171,7 @@
             <wast>func.wast</wast>
             <wast>func_ptrs.wast</wast>
             <wast>global.wast</wast>
+            <wast>i32.wast</wast>
```
electrum, line 174: Is this a regression?

runtime/src/main/java/com/dylibso/chicory/runtime/Instance.java
electrum: This means we are now doing validation before running the start function -- which seems correct.

runtime/src/main/java/com/dylibso/chicory/runtime/Module.java
```diff
@@ -226,11 +226,9 @@ protected Instance instantiate(
 
         Memory memory = null;
         if (module.memorySection() != null) {
```
electrum, line 228: When we cleanup the APIs, we should consider changing all of these nullables to `Optional`

wasm/src/main/java/com/dylibso/chicory/wasm/types/MemoryLimits.java
```diff
@@ -63,14 +65,12 @@ public MemoryLimits(int initial, int maximum) {
      * @param shared {@code true} if the limits apply to a shared memory segment, or {@code false} otherwise
      */
     public MemoryLimits(int initial, int maximum, boolean shared) {
-        if (initial < 0 || initial > maximum) {
-            throw new IllegalArgumentException(
-                    "initial must be >= 0 and <= maximum, but was " + initial);
+        if (initial > MAX_PAGES || maximum > MAX_PAGES || initial < 0 || maximum < 0) {
```
electrum, line 68: Could do a separate check for negative, unless this error message is required per specification.
```java
if (initial < 0 || maximum < 0) {
    throw new InvalidException("memory size is negative");
} else if (initial > MAX_PAGES || maximum > MAX_PAGES) {

electrum (approved):
Looks like a great start! I'm learning a lot by reading through this code.

runtime-tests/pom.xml
andreaTP: not really, just "more correct" checks, expect a bit of forth and back until we have a super solid control flow.

runtime/src/main/java/com/dylibso/chicory/runtime/Module.java
andreaTP: Sadly `Optional`s are a leaky abstraction in Java, I would be happy to discuss this change, I might be overthinking something.

wasm/src/main/java/com/dylibso/chicory/wasm/types/MemoryLimits.java
andreaTP: > unless this error message is required per specification.

It's the case, for context `< 0` usually happens during overflows other than real negative input passed by the user.

andreaTP:
I merge this as a checkpoint, will iterate on it, feel free to open separate issues if I miss/forget to follow up on something required!

PR #342, opened 2024-05-18
Use resources from wasm-corpus as a normal Maven artifact

electrum:
IntelliJ doesn't allow using the same content root for multiple modules and doesn't allow saving changes to the project structure. Instead, we can make `wasm-corpus` provide the resources as a normal Maven artifact and reference it as a test dependency.

We use `src/main/resources` so that the resources are in the artifact's main JAR. Otherwise, we would would need to depend on the `test-jar`. This is correct because from the perspective of the `wasm-corpus` module, the resources are what the artifact provides. The fact that they're used in tests is a function of the consuming modules. This is similar to how a library like `junit` is a normal Maven library that happens to be used for testing.

With the existing structure, IntelliJ has this error when trying to save project structure settings:

Screenshot 2024-05-18 at 11 57 17 AM


bhelx (approved):
I hadn't noticed this problem, but makes sense to me. Thanks for updating the docs too!

PR #343, opened 2024-05-19
Implement I64, F32 and F64 opcodes for AOT

electrum:


andreaTP (approved):
LGTM! Nice job!

danielperano (approved):
Thanks, looks great!

Eventually we'll want to move (un)boxing into a calling layer above the AOT (so that AOT'd methods can call each other without boxing overhead) but for now this is perfect.

PR #348, opened 2024-05-23
rework Wat2Wasm to generate bytecode also from a String of code

mariofusco:


wabt/src/main/java/com/dylibso/chicory/wabt/Wat2Wasm.java
```diff
@@ -30,30 +33,45 @@ public final class Wat2Wasm {
...
+
+    public static byte[] parse(String wat) {
+        try (InputStream is = new ByteArrayInputStream(wat.getBytes(StandardCharsets.UTF_8))) {
+            return parse(is, "temp.wast");
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
+    }
+
+    private static byte[] parse(InputStream is, String fileName) {
```
electrum, line 51: The `fileName` parameter is unused. We can remove it.

wabt/src/main/java/com/dylibso/chicory/wabt/Wat2Wasm.java
```diff
@@ -30,30 +33,45 @@ public final class Wat2Wasm {
...
                 Path target = fs.getPath("tmp");
                 java.nio.file.Files.createDirectory(target);
-                Path path = target.resolve(file.getName());
-                copy(fis, path, StandardCopyOption.REPLACE_EXISTING);
-                wasiOpts.withDirectory(target.toString(), target);
+                Path path = target.resolve("test.wast");
+                copy(is, path, StandardCopyOption.REPLACE_EXISTING);
 
-                wasiOpts.withArguments(List.of("wat2wasm", path.toString(), "--output=-"));
+                WasiOptions wasiOpts =
```
electrum, line 66: I like how this is now a chained builder. Much nicer to read!

wabt/src/main/java/com/dylibso/chicory/wabt/Wat2Wasm.java
```diff
@@ -30,30 +33,45 @@ public final class Wat2Wasm {
...
-                            Jimfs.newFileSystem(
-                                    Configuration.unix().toBuilder()
-                                            .setAttributeViews("unix")
-                                            .build())) {
-
-                var wasiOpts = WasiOptions.builder();
 
-                wasiOpts.withStdout(stdoutStream);
-                wasiOpts.withStderr(stdoutStream);
+            try (FileSystem fs =
```
electrum, line 57: We could merge this try-with-resources into the above, then add an normal try block. So it would look like this:
```java
try (var stdoutStream = ...;
        var stderrStream = ...;
        var fs = ...) {
    try {
        ...
    } catch (WASMMachineException e) {
        ...
    }
}
```
This is just a suggestion, since this is a problem with the original code.

electrum (approved):

wabt/src/main/java/com/dylibso/chicory/wabt/Wat2Wasm.java
mariofusco: I tried to, but unfortunately that's not possible because `stdoutStream` is used also in the inner catch block.

andreaTP (approved):
Thanks for the nice cleanup! :+1:

mariofusco:
@andreaTP This wasn't meant to be only a cleanup, but also a first attempt to demonstrate how we can evaluate a WebAssembly function in a very simple (yet admittedly quite inefficient) way, without using any memory mapping tricky code as I saw in other examples. For instance I believe that this [Camel integration example](https://lburgazzoli.github.io/posts/2024-01-14_apache_camel_meets_wasm_part_1/) could be more easily rewritten using this technique, or am I missing something?

andreaTP:
It's a matter, as usual, of tradeoffs.
This mechanism can be used in the Camel integration example, but giving access to the `wasi` layer does have security and sandboxing implications.

Depending on your use-case, it can be perfectly ok or not.

PR #349, opened 2024-05-24
[FUZZ] Crash reproducer

bhelx:
Fuzz testing failed and there are new reproducers available.

PR #355, opened 2024-05-28
Reduce 349

andreaTP:
Opening this PR for 2 reasons:
- it shows a valid way to reduce a fuzz reproducer
- it goes full circle, using `wasm-interp` from the wabt tools to compare with the bare result of Chicory

Supersede #349


bhelx (approved):

runtime/src/test/java/com/dylibso/chicory/runtime/ModuleTest.java
```diff
@@ -321,6 +321,17 @@ public void issue294_BRTABLE() {
         assertEquals(4, main.apply()[0].asInt());
     }
 
+    @Test
+    public void fuzzCrashReproY4ye0lx() {
```
electrum, line 325: I'm confused. What is this supposed to test? Since all the tests pass, it's not crashing.

andreaTP: This test demonstrates that the interpreter is emitting the correct result, and the problem is in "mimicking" the `wasmtime` output.
I will follow up on this eventually, I'm considering moving to `wasm-interp` as an "Oracle" as it has a better handling of Wasm types.

andreaTP:
We need first to re-enable the fuzz tests based on `wasm-interp`

PR #356, opened 2024-05-29
Cleaner separation in Module.Builder

### LINKED ISSUE #344: `runtime.Module` should not be mutable
Noticed here: https://github.com/dylibso/chicory/pull/341#issuecomment-2118979440

>  it would be nice to only parse a WASM module once

andreaTP:
Suggestion from @lburgazzoli , these changes makes the builder API for building modules much nicer and better defined.
The `runtime.Module` class is now immutable and this should fix #344 if I'm not missing details.

To be able to re-use(hence memoize) Parsed WASM Modules I added a builder from `wasm.Module`, this way one can instantiate a "dummy" module and re-use it; the usage is in the `wabt` tools module.

### DIFF
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/Module.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/Module.java
@@ -36,22 +36,33 @@ import java.util.function.Supplier;
 public class Module {
     public static final String START_FUNCTION_NAME = "_start";
     private final com.dylibso.chicory.wasm.Module module;
-    private NameCustomSection nameSec;
 
     private final HashMap<String, Export> exports;
     private final Logger logger;
 
-    private boolean initialize = true;
-    private boolean start = true;
-    // TODO: turn the default to true
-    // Type validation needs to remain optional until it's finished
-    private boolean typeValidation = false;
-    private ExecutionListener listener = null;
-    private HostImports hostImports;
-    private Function<Instance, Machine> machineFactory;
+    private final boolean initialize;
+    private final boolean start;
+    private final boolean typeValidation;
+    private final ExecutionListener listener;
+    private final HostImports hostImports;
+    private final Function<Instance, Machine> machineFactory;
 
-    protected Module(com.dylibso.chicory.wasm.Module module, Logger logger) {
+    private Module(
+            com.dylibso.chicory.wasm.Module module,
+            Logger logger,
+            Function<Instance, Machine> machineFactory,
+            HostImports hostImports,
+            ExecutionListener listener,
+            boolean initialize,
+            boolean start,
+            boolean typeValidation) {
         this.logger = logger;
+        this.machineFactory = machineFactory;
+        this.hostImports = hostImports;
+        this.listener = listener;
+        this.initialize = initialize;
+        this.start = start;
+        this.typeValidation = typeValidation;
         this.module = validateModule(module);
         this.exports = new HashMap<>();
         if (module.exportSection() != null) {
@@ -83,52 +94,7 @@ public class Module {
         return logger;
     }
 
-    public Module withInitialize(boolean init) {
-        this.initialize = init;
-        return this;
-    }
-
-    public Module withStart(boolean s) {
-        this.start = s;
-        return this;
-    }
-
-    public Module withTypeValidation(boolean v) {
-        this.typeValidation = v;
-        return this;
-    }
-
-    /*
-     * This method is experimental and might be dropped without notice in future releases.
-     */
-    public Module withUnsafeExecutionListener(ExecutionListener listener) {
-        this.listener = listener;
-        return this;
-    }
-
-    public Module withHostImports(HostImports hostImports) {
-        this.hostImports = hostImports;
-        return this;
-    }
-
-    public Module withMachineFactory(Function<Instance, Machine> machineFactory) {
-        this.machineFactory = machineFactory;
-        return this;
-    }
-
     public Instance instantiate() {
-        if (hostImports == null) {
-            hostImports = new HostImports();
-        }
-
-        return this.instantiate(hostImports, initialize, start, listener);
-    }
-
-    protected Instance instantiate(
-            HostImports hostImports,
-            boolean initialize,
-            boolean start,
-            ExecutionListener listener) {
         var globalInitializers = new Global[] {};
         if (this.module.globalSection() != null) {
             globalInitializers = this.module.globalSection().globals();
@@ -256,10 +222,6 @@ public class Module {
                 default:
                     break;
             }
-        }
-
-        if (machineFactory == null) {
-            machineFactory = InterpreterMachine::new;
         }
 
         return new Instance(
@@ -399,9 +361,7 @@ public class Module {
     }
 
     public NameCustomSection nameSection() {
-        if (nameSec != null) return nameSec;
-        nameSec = this.module.nameSection();
-        return nameSec;
+        return this.module.nameSection();
     }
 
     public com.dylibso.chicory.wasm.Module wasmModule() {
@@ -416,6 +376,16 @@ public class Module {
      */
     public static Builder builder(InputStream input) {
         return new Builder(() -> input);
+    }
+
+    /**
+     * Creates a {@link Builder} for the specified {@link com.dylibso.chicory.wasm.Module}
+     *
+     * @param wasmModule the already parsed Wasm module
+     * @return a {@link Builder} for reading the module definition from the specified input stream
+     */
+    public static Builder builder(com.dylibso.chicory.wasm.Module wasmModule) {
+        return new Builder(wasmModule);
     }
 
     /**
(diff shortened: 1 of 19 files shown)


electrum:
This is great! I'll review later today. 

Regarding being able to share, I thought I noticed some mutable objects (Table?) in the `wasm` package, so I don't think it's safe yet.

electrum:
Another API consideration is that the AOT compilation should be tied to the immutable parsed WASM and shared across instances. Currently, it's constructed per instance.

andreaTP:
Ahah, the CI check on the Readme is great to prevent my mistakes :-)
https://github.com/dylibso/chicory/actions/runs/9288031336/job/25558461170?pr=356

andreaTP:
> Regarding being able to share, I thought I noticed some mutable objects (Table?) in the wasm package, so I don't think it's safe yet.

Interesting, yes, `Table` is indeed mutable as it holds indexes that are set etc. at runtime.
It should be doable to move those bits to `TableInstance` in the runtime module.

@electrum do you have an idea or suggestion on how to get a full list of those issues?

andreaTP:
I reviewed the entire `wasm` submodule, and removed mutability everywhere but:

- `Instruction`: is still mutable as it will require a little more work and I prefer to separate in a follow up PR
- `wasm.Module` and all the sections are still kind of mutable(section's fields are final references to mutable lists), I would go down a path like what I proposed in #211 but would like to get feedback before I rework this aspect cc. @electrum and @bhelx

electrum:
> do you have an idea or suggestion on how to get a full list of those issues?

I don't know of a good way, other than a thorough review of the code, which I'm happy to do.

We could start using [Error Prone](https://errorprone.info/) and mark the classes as [Immutable](https://errorprone.info/bugpattern/Immutable), which should provide enforcement and safety/security guarantees.

electrum:
> `Instruction`: is still mutable as it will require a little more work and I prefer to separate in a follow up PR

Absolutely. I'm not proposing to fix everything at once. We'll keep making progress and get there eventually.

> `wasm.Module` and all the sections are still kind of mutable(section's fields are final references to mutable lists), I would go down a path like what I proposed in https://github.com/dylibso/chicory/pull/211 but would like to get feedback before I rework this aspect

Making everything immutable with builders makes sense to me. It's the typical approach and I can't see a better way.

andreaTP:
> I don't know of a good way, other than a thorough review of the code, which I'm happy to do.

:pray: 

> We could start using [Error Prone](https://errorprone.info/) and mark the classes as [Immutable](https://errorprone.info/bugpattern/Immutable), which should provide enforcement and safety/security guarantees.

All in for it as long as we make sure we don't add external dependencies (I assume it's possible).

> Making everything immutable with builders makes sense to me. It's the typical approach and I can't see a better way.

Thanks for the feedback, noted, we get back to it.

electrum:
> All in for it as long as we make sure we don't add external dependencies (I assume it's possible).

We can make the Error Prone annotations dependency optional, which means it is non-transitive and won't be added to the user's classpath. They're only needed at compile time for the Error Prone checks. The JVM ignores missing annotations when loading classes.

andreaTP:
> The JVM ignores missing annotations when loading classes.

I might be wrong but they still need to be on the classpath to compile IIRC.
Happy to hear otherwise or to automate a mechanism that will take the benefits without the need for the dependency (a processor running before `package` to clean things up? Alternatives?).

electrum:
The annotation classes are on the class path to compile Chicory, but won’t be on the class path for Chicory users, as we will make them optional (non-transitive) in Maven. This is normal for these types of annotations.

The compiled Chicory classes will contain references to these annotations, but these are ignored by the JVM if the annotation classes are not present.

If the Chicory user is also using Error Prone and has the annotation classes in their project’s class path, then they might also get benefits from the annotations. For example, if they have an `@Immutable` class that contains a reference to a Chicory `@Immutable` class, then Error prone can verify the whole thing is immutable. (Just an example. In practice, it’s likely not useful. But hopefully this helps in understanding how it works.)

wasm/src/main/java/com/dylibso/chicory/wasm/types/FunctionSection.java
```diff
@@ -1,17 +1,15 @@
 package com.dylibso.chicory.wasm.types;
 
-import java.util.Arrays;
-import java.util.Objects;
+import java.util.ArrayList;
 
 public class FunctionSection extends Section {
-    private int[] typeIndices;
-    private int count;
+    private final ArrayList<Integer> typeIndices;
```
electrum, line 6: Using lists instead of arrays is a nice improvement.

Note that it's typically better to declare as the interface type, so `List<Integer>`, as the calling code doesn't (and shouldn't) depend on the concrete type.

wasm/src/main/java/com/dylibso/chicory/wasm/types/NameCustomSection.java
```diff
@@ -12,30 +12,59 @@
...
+    private final ArrayList<NameEntry> dataNames;
+    private final ArrayList<NameEntry> tagNames;
 
     /**
      * Construct a new, empty section instance.
      */
-    public NameCustomSection() {}
+    public NameCustomSection(
+            String moduleName,
+            ArrayList<NameEntry> funcNames,
```
electrum, line 31: If we make these `List<X>` rather than `ArrayList<X>`, then the caller can pass in any list type. We should also make defensive copies, to ensure that the class is immutable:
```java
this.moduleName = List.copyOf(moduleName);

File: wasm/src/main/java/com/dylibso/chicory/wasm/types/NameCustomSection.java [hunk 3]
```diff
@@ -12,30 +12,59 @@
...
+    private final ArrayList<ListEntry<NameEntry>> labelNames;
+    private final ArrayList<NameEntry> tableNames;
+    private final ArrayList<NameEntry> memoryNames;
+    private final ArrayList<NameEntry> globalNames;
+    private final ArrayList<NameEntry> elementNames;
+    private final ArrayList<NameEntry> dataNames;
+    private final ArrayList<NameEntry> tagNames;
 
     /**
      * Construct a new, empty section instance.
```
electrum, line 27: Comment needs to be updated:

> Construct a section instance from the specified contents.

electrum (approved):
This is great! As I mentioned in comments, I don't think the API is right yet, but this change is a big improvement and is moving in the right direction. The cleanup of `runtime.Module` makes it easier to see what the API could look like. We can merge this change now and continue to iterate.

wabt/src/main/java/com/dylibso/chicory/wabt/Wast2Json.java
```diff
@@ -33,6 +33,12 @@ public boolean isLoggable(Logger.Level level) {
                     return false;
                 }
             };
+    private static final com.dylibso.chicory.wasm.Module wasmModule =
+            Module.builder(Wast2Json.class.getResourceAsStream("/wast2json"))
+                    .withInitialize(false)
+                    .withStart(false)
+                    .build()
+                    .wasmModule();
```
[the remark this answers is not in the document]
lburgazzoli: I think this would be the way to go

wasm/src/main/java/com/dylibso/chicory/wasm/types/NameCustomSection.java
lburgazzoli: +1

runtime/src/main/java/com/dylibso/chicory/runtime/Module.java
```diff
@@ -36,22 +36,33 @@
...
+            boolean start,
+            boolean typeValidation) {
         this.logger = logger;
+        this.machineFactory = machineFactory;
+        this.hostImports = hostImports;
+        this.listener = listener;
+        this.initialize = initialize;
+        this.start = start;
+        this.typeValidation = typeValidation;
         this.module = validateModule(module);
```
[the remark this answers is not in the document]
andreaTP: It's possible that we can move it earlier in the chain, but I leave this out of this PR as we can probably review "where" the validation logic happens later on.

wabt/src/main/java/com/dylibso/chicory/wabt/Wast2Json.java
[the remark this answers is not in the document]
andreaTP: Completely agree.
This will allow for maximum reuse and e.g. swapping `HostImports` for different `Module` executions with WASI.

side note:
I keep being confused by:
`com.dylibso.chicory.wasm.Module`
`com.dylibso.chicory.runtime.Module`
but I don't have better names to differentiate too :cry:

wasm/src/main/java/com/dylibso/chicory/wasm/types/FunctionSection.java
andreaTP: Yes, too much copy-pasting! Fixing it :slightly_smiling_face:

wasm/src/main/java/com/dylibso/chicory/wasm/types/NameCustomSection.java
andreaTP: good catch, thanks :+1:

wasm/src/main/java/com/dylibso/chicory/wasm/types/NameCustomSection.java
andreaTP: right, thanks for noticing, I'm not great at maintaining comments :sweat:

andreaTP (commented):
> I don't think the API is right yet, but this change is a big improvement and is moving in the right direction.

Agree, let's refine on top as there is some more work needed here.

PR #361, opened 2024-05-31
Review the type validator control flow

andreaTP:
Still not perfect, but happy to discuss the approach etc.
Looks a bit better than before, almost no NPEs on unrelated specs.

### DIFF
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/TypeValidator.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/TypeValidator.java
@@ -32,9 +32,7 @@ public class TypeValidator {
     }
 
     private static <T> List<T> clone(List<T> list) {
-        var res = new ArrayList<T>();
-        res.addAll(list);
-        return res;
+        return new ArrayList<>(list);
     }
 
     private void popAndVerifyType(ValueType expected) {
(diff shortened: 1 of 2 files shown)


runtime/src/main/java/com/dylibso/chicory/runtime/TypeValidator.java
```diff
@@ -6,28 +6,45 @@
...
+    }
+
+    private static <T> void push(List<T> list, T elem) {
+        list.add(elem);
+    }
+
+    private static <T> List<T> clone(List<T> list) {
+        var res = new ArrayList<T>();
+        res.addAll(list);
+        return res;
```
electrum, line 37:
```suggestion
        return new ArrayList<>(list);
```

electrum (approved):

runtime/src/main/java/com/dylibso/chicory/runtime/TypeValidator.java
mariofusco: I'm not even sure if this deserves a separated method tbh.

mariofusco (approved):

runtime/src/main/java/com/dylibso/chicory/runtime/TypeValidator.java
andreaTP: Possibly you are right @mariofusco , just a lot of forth and back to get there, I'll review the entire class at the end, there are other things to be tightened up ...

PR #362, opened 2024-05-31
Implement more AOT opcodes

electrum:
Some notes:

* Use `java.lang.reflect.Method` rather than manually building descriptors. IntelliJ understands `Class.getMethod(...)` for usage reporting, refactoring, code completion, etc., and will warn if the method does not exist.
* Use static trampoline methods rather than generating code to rearrange items on the stack. This is simpler, easier to read, results in smaller bytecode, and should be completely inlined by the JIT.
* We can code generate globals with host accessors if the boxing turns out to be a performance issue.

### DIFF
--- a/aot/pom.xml
+++ b/aot/pom.xml
@@ -77,7 +77,11 @@
             <wast>inline-module.wast</wast>
             <wast>int_exprs.wast</wast>
             <wast>int_literals.wast</wast>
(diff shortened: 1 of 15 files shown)


bhelx (approved):
I like it! Lots of good simplifications.

PR #363, opened 2024-05-31
Format code with Palantir Java Format

electrum:
This is a proposal to use [Palantir Java Format](https://github.com/palantir/palantir-java-format). It's based on `google-java-format` but is more lambda friendly. See the examples in the README and compare the formatted code. This resolves my personal complaint about the excessive vertical formatting of `google-java-format`.

The only actual code change here is the configuration in `pom.xml`. The rest is the result of running the formater. The changes are easier to see if you view the diff with "hide whitespace".

This uses `<style>AOSP</style>` which maintains the 100 character limit. This change brings the option of using  `<style>PALANTIR</style>` which has a 120 character limit.

### DIFF
--- a/aot/src/main/java/com/dylibso/chicory/aot/AotEmitters.java
+++ b/aot/src/main/java/com/dylibso/chicory/aot/AotEmitters.java
@@ -29,11 +29,10 @@ public class AotEmitters {
         protected final Map<OpCode, BytecodeEmitter> emitters = new EnumMap<>(OpCode.class);
 
         public Builder intrinsic(OpCode opCode, BytecodeEmitter emitter) {
-            BytecodeEmitter wrapped =
-                    (ctx, ins, asm) -> {
-                        emitter.emit(ctx, ins, asm);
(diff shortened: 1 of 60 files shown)


bhelx (approved):
I'm okay with this. I don't have a strong opinion on any particular format. My main concern is only that one be in place and be enforced. I typically don't like these changes because they create a lot of noise, though this one doesn't have a huge surface area so it's acceptable. The only thing I'd ask is that this is the last time we change it 😆

andreaTP (approved):
Fully aligned with what @bhelx said, no strong preference here.
Ideally this PR should have 2 commits, one with only the `pom.xml` changes and one applying it, but it's a nitpick, no action required.

electrum:
@andreaTP Good point about two commits -- updated for that

danielperano (approved):

danielperano:
I like it!

thomasdarimont (approved):

bhelx:
Closing as discussed until we find a replacement for the Palantir formatter.

PR #364, opened 2024-06-02
Format code using 120 characters

electrum:
This is a variant of #363 that uses the 120 character line width. The last commit uses the same code style but changes from 100 to 120 characters, so you can see that commit to compare just the line with change, or view the whole PR to see the difference from the existing style.


danielperano (approved):
I like this one too - I like 120 slightly better than 100, but both are visually pleasing formats to my eye.

bhelx:
See #363

PR #369, opened 2024-06-03
Implement CALL for AOT

electrum:
This changes the compiler to generate a single, non-instantiable class for the entire WASM module, with a static method for each WASM method. The `Memory` and `Instance` parameters are last, which means their values are pushed onto the JVM stack last, thus methods can be called without any stack manipulation (or a much more complicated compiler).

### DIFF
--- a/aot/src/main/java/com/dylibso/chicory/aot/AotUtil.java
+++ b/aot/src/main/java/com/dylibso/chicory/aot/AotUtil.java
@@ -4,6 +4,8 @@ import static com.dylibso.chicory.wasm.types.Value.REF_NULL_VALUE;
 import static org.objectweb.asm.Type.getInternalName;
 import static org.objectweb.asm.Type.getMethodDescriptor;
 
+import com.dylibso.chicory.runtime.Instance;
+import com.dylibso.chicory.runtime.Memory;
 import com.dylibso.chicory.wasm.types.FunctionBody;
 import com.dylibso.chicory.wasm.types.FunctionType;
 import com.dylibso.chicory.wasm.types.Value;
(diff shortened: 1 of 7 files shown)


aot/src/main/java/com/dylibso/chicory/aot/AotUtil.java
```diff
@@ -70,6 +73,23 @@ public static Class<?> jvmType(ValueType type) {
         }
     }
 
+    public static int loadTypeOpcode(ValueType type) {
```
andreaTP, line 76: - nitpick - would it make sense to have those methods in the `ValueType` class?
`Util` looks a bit too broad of a scope to keep pushing things into, if possible.

andreaTP (approved):
LGTM, great work! Thanks!

aot/src/main/java/com/dylibso/chicory/aot/AotUtil.java
electrum: `ValueType` is in the WASM module. This utility method (and all the others here) are specific to AOT bytecode generation. This one has a dependency on the ASM library for the constants that it returns.

I completely agree with your concern about `Util` classes in general -- we've had "no modules named `common`" rule in other projects. The `AotUtil` class has methods that are shared between `AotMachine` and `AotEmitters`. This particular method is currently only used in `AotMachine`, but that class is already 600+ lines long, so putting the method next to similar methods in `AotUtil` seems cleaner for now.

We should make `AotUtil` and most things in the module to be package-private. I can do a cleanup for that. We need to do more of that for other places in Chicory, so that the public APIs aren't exposing internal things that might change. Some things might need to be public for technical reasons but should be documented as internal-only so that people don't depend on them.

andreaTP: Thanks for explaining, make sense.
I'm happy to hear that the direction is aligned and we can clean up things as we go :+1:

PR #370, opened 2024-06-03
Make AOT classes package-private and final

electrum:
This minimizes the public surface area of the module.


bhelx (approved):
Looks pretty straight forward!

danielperano (approved):

PR #373, opened 2024-06-05
Implement CALL_INDIRECT for AOT

electrum:
This is ugly and expensive, as it uses boxing for everything, even functions inside the current module. We can replace this with `invokedynamic` later.

### DIFF
--- a/aot/src/main/java/com/dylibso/chicory/aot/AotEmitters.java
+++ b/aot/src/main/java/com/dylibso/chicory/aot/AotEmitters.java
@@ -33,7 +33,10 @@ import static com.dylibso.chicory.aot.AotMethods.TABLE_SIZE;
 import static com.dylibso.chicory.aot.AotMethods.THROW_OUT_OF_BOUNDS_MEMORY_ACCESS;
 import static com.dylibso.chicory.aot.AotMethods.THROW_TRAP_EXCEPTION;
 import static com.dylibso.chicory.aot.AotMethods.VALIDATE_BASE;
+import static com.dylibso.chicory.aot.AotUtil.StackSize;
 import static com.dylibso.chicory.aot.AotUtil.boxer;
+import static com.dylibso.chicory.aot.AotUtil.callIndirectMethodName;
+import static com.dylibso.chicory.aot.AotUtil.callIndirectMethodType;
 import static com.dylibso.chicory.aot.AotUtil.emitInvokeStatic;
 import static com.dylibso.chicory.aot.AotUtil.emitInvokeVirtual;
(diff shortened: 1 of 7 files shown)


bhelx:
I'll take a look in the morning. At first glance it seems understandable at least and it is a step towards a complete implementation. We can always come back and make it faster.

aot/src/main/java/com/dylibso/chicory/aot/AotEmitters.java
```diff
@@ -1,17 +1,9 @@
...
 import static com.dylibso.chicory.aot.AotMethods.*;
-import static com.dylibso.chicory.aot.AotUtil.boxer;
-import static com.dylibso.chicory.aot.AotUtil.emitInvokeStatic;
-import static com.dylibso.chicory.aot.AotUtil.emitInvokeVirtual;
-import static com.dylibso.chicory.aot.AotUtil.methodNameFor;
-import static com.dylibso.chicory.aot.AotUtil.methodTypeFor;
-import static com.dylibso.chicory.aot.AotUtil.stackSize;
-import static com.dylibso.chicory.aot.AotUtil.unboxer;
-import static com.dylibso.chicory.aot.AotUtil.validateArgumentType;
+import static com.dylibso.chicory.aot.AotUtil.*;
```
andreaTP, line 4: - unrelated - this shows a problem with the spotless setup, as [this rule](https://github.com/dylibso/chicory/blob/df74f2a258de42e5eebfda4f6e34bc9784477f5d/pom.xml#L388) should prevent star imports.
In this specific case, it reduces the pain of having an `Util` class, which is something I would possibly discourage.

andreaTP: oh, maybe the regexp is not capturing `import static`?

electrum: I didn't realize that star imports were supposed to be disallowed in Chicory. We've always banned them in other projects. I had to change my IntelliJ settings for Chicory to *not* remove the star imports.

Note that `AotMachine` has `OpCode.*` and expanding that will add ~200 imports, since it has one per opcode. IntelliJ hides imports, so I don't care either way, just FYI.

electrum: I put up a PR to remove wildcard imports and updated this PR to not add a new wildcard import.

bhelx (approved):

andreaTP:
@electrum sorry if I missed it, would you mind capturing the suggestion to use `invokedynamic` in an issue so that we won't forget about it? :pray:

PR #374, opened 2024-06-05
include nop.wast assert invalid

andreaTP:
I hope you don't mind me taking over your very welcome effort @thomasdarimont  .
I think that this is what's missing for `nop`


thomasdarimont (approved):
Thanks for looking into it :)

LGTM

PR #378, opened 2024-06-07
Implement multiple return values for AOT

electrum:


aot/src/test/java/com/dylibso/chicory/imports/SpecV1ElemHostFuncs.java
```diff
@@ -0,0 +1,106 @@
...
+import com.dylibso.chicory.runtime.HostGlobal;
+import com.dylibso.chicory.runtime.HostImports;
+import com.dylibso.chicory.runtime.HostTable;
+import com.dylibso.chicory.runtime.TableInstance;
+import com.dylibso.chicory.wasm.types.Limits;
+import com.dylibso.chicory.wasm.types.Table;
+import com.dylibso.chicory.wasm.types.Value;
+import com.dylibso.chicory.wasm.types.ValueType;
+
+public class SpecV1ElemHostFuncs {
```
andreaTP, line 17: I assume that we are copy-pasting those HostFuncs from the `runtime-tests` module, correct?

If this is the case, we can probably refactor to avoid duplication (no need to do it in this PR though).

andreaTP (approved):
Amazing progress, thanks!
LGTM!

aot/src/test/java/com/dylibso/chicory/imports/SpecV1ElemHostFuncs.java
electrum: Correct -- that was in the first version of AOT. I'll put up a PR to clean this up.

PR #382, opened 2024-06-08
Add `quick-build` profile to install skipping codegen and tests

evacchi:
Add property `wasm-test-gen.skip` to the codegen plugin, and configure a `quick-build` profile to skip all tests and install the artifacts. The new profile defaults to `clean install` so to install the whole repo it should now be enough to just:

    $ mvn -Dquicky


andreaTP (approved):
LGTM! Thanks for sending over this improvement!

PR #383, opened 2024-06-09
docs: fix wasi readme

bhelx:


andreaTP (approved):
Thanks, looks good!
We should test this readme as we do with the main at root probably. But not needed for this PR to land.

PR #394, opened 2024-06-19
Cleanup Javadoc for ControlTree

electrum:


wasm/src/main/java/com/dylibso/chicory/wasm/ControlTree.java
```diff
@@ -45,8 +48,8 @@
  * 18     local.set 1
  *        )
  * 19   local.get 1)
- *
- * This is very sub-optimal at the moment
+ * </pre>
+ * This is very suboptimal at the moment.
```
andreaTP, line 52: I think that we can leave this out now, we rewrote the algorithm to not be greedy and rely on callbacks

electrum: Nice, removed

andreaTP (approved):
LGTM, thanks!

PR #396, opened 2024-06-19
Replace usage of FileUtils with JDK Files

electrum:


andreaTP (approved):
LGTM, thanks for the cleanup!

PR #400, opened 2024-06-20
Review type validator control flow

andreaTP:
Here I switched to this suggested implementation:
https://webassembly.github.io/spec/core/appendix/algorithm.html

There is still something breaking that I don't fully understand, removing `block.wast` from `excludedInvalidWasts` shows a couple of points where this breaks.

@electrum since you have been working on the control flow recently, I would love review/fixes/suggestions by you :pray:

### DIFF
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/Instance.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/Instance.java
@@ -341,6 +341,9 @@ public class Instance {
     }
 
     public Element element(int idx) {
+        if (idx < 0 || idx >= elements.length) {
+            throw new InvalidException("unknown elem segment " + idx);
+        }
         return elements[idx];
     }
 
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/Instance.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/Instance.java
@@ -289,6 +289,10 @@ public class Instance {
         if (idx < importedGlobalsOffset) {
             return imports.global(idx).instance().getValue();
         }
+        var i = idx - importedGlobalsOffset;
+        if (i < 0 || i >= globals.length) {
+            throw new InvalidException("unknown global");
+        }
         return globals[idx - importedGlobalsOffset].getValue();
     }
 
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/Instance.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/Instance.java
@@ -136,7 +136,7 @@ public class Instance {
 
                 Value offset = computeConstantValue(this, ae.offset());
                 if (offset.type() != ValueType.I32) {
-                    throw new ChicoryException("Invalid offset type in element");
+                    throw new ChicoryException("Invalid offset type in element " + offset.type());
                 }
                 List<List<Instruction>> initializers = ae.initializers();
                 for (int i = 0; i < initializers.size(); i++) {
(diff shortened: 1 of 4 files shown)


andreaTP:
Thanks to @evacchi pairing, we found the last missing detail and now it finally clicked.
Ready for review!

runtime/src/main/java/com/dylibso/chicory/runtime/TypeValidator.java
```diff
@@ -5,128 +5,146 @@
...
-        var val = list.get(list.size() - 1);
-        list.remove(list.size() - 1);
-        return val;
+    private static class CtrlFrame {
+        private final OpCode opCode;
+        private final List<ValueType> startTypes;
+        private final List<ValueType> endTypes;
+        private final int height;
+        private boolean unreachable;
+        private boolean hasElse;
```
evacchi, line 34: maybe worth adding a few comments on the meaning of these fields

runtime/src/main/java/com/dylibso/chicory/runtime/TypeValidator.java
```diff
@@ -155,140 +175,102 @@ private static ValueType getLocalType(List<ValueType> localTypes, int idx) {
...
                         break;
                     }
                 case IF:
                     {
-                        popAndVerifyType(ValueType.I32);
-                        push(prevStack, clone(valueTypeStack));
+                        popVal(ValueType.I32);
+                        // fallthrough
+                    }
+                case LOOP: // t1* -> t2*
```
evacchi, line 195:
```suggestion
                case LOOP: // t1* -> t2*
                     // fallthrough
```

evacchi (approved):
some nits, otherwise yay 👍

runtime/src/main/java/com/dylibso/chicory/runtime/TypeValidator.java
andreaTP: In this case the comment is a placeholder instead of `break`, any reason to push it down?

runtime/src/main/java/com/dylibso/chicory/runtime/TypeValidator.java
andreaTP: done

runtime/src/main/java/com/dylibso/chicory/runtime/TypeValidator.java
evacchi: Oh, I am not suggesting to push it down, I am suggesting to add it on LOOP too, since its body is empty

runtime/src/main/java/com/dylibso/chicory/runtime/Instance.java
```diff
@@ -211,8 +211,11 @@ public Instance initialize(boolean start) {
             // TODO: can be parallelized?
             for (int i = 0; i < this.functions.length; i++) {
                 if (this.function(i) != null) {
-                    new TypeValidator()
-                            .validate(this.function(i), this.types[this.functionType(i)], this);
+                    var funcType = this.functionType(i);
+                    if (funcType >= this.types.length) {
+                        throw new InvalidException("unknown type");
```
mariofusco, line 216: Nitpick: wouldn't be better to also add in the message of the exception the id of the unknown type? This could help for debuggability. Same thing for global and elements after in this file.

andreaTP: done :+1:

PR #409, opened 2024-06-25
Update milestones, fix zulip badge

bhelx:


andreaTP (approved):
I would be good with less tags/mentions (to me, in particular ;-P ) thanks...
It's a nitpick though, LGTM, thanks for taking care of this!

andreaTP (approved):
LGTM, and either I disable the test reports comments or we open PRs only from forks

PR #414, opened 2024-06-28
doc: Add simple doc for AotMachine usage

bhelx:


aot/README.md
```diff
@@ -1,3 +1,18 @@
...
+
+```java
+// ...
+import com.dylibso.chicory.aot.AotMachine;
+// ...
+
+var module = Module
+        .builder("mymodule.wasm")
+        .withMachineFactory(
+            instance -> new AotMachine(instance.module().wasmModule(), instance)
```
bhelx, line 15: Looking at this as a user, could we not just narrow AotMachine to only take the instance parameter?

andreaTP: In the tests is slightly different:
https://github.com/dylibso/chicory/blob/530f6e5d0c03158613663f6c92c3631297fd4705/aot/src/test/java/com/dylibso/chicory/testing/TestModule.java#L79

That said, I think that this API is just instrumental to enable the usage of the test-gen  infrastructure.
I'm not against merging it, but we need a few more steps here to provide something "usable".
Cc. @danielperano @electrum

bhelx:
I copy pasted from the test but then got confused because it appeared to be using the module before it was “built”. I think it works because that module is an instance var in scope? I’m assuming the intention is to use the variables coming in to the lambda.

bhelx:
Also yes, not super concerned about the api I just want to document it so people can try it.

andreaTP (approved):

aot/README.md
electrum: Yeah, we should change the API to just take `Instance` for now. I put up a PR for that: https://github.com/dylibso/chicory/pull/416

The API really needs to be split so that we can compile separately from binding to the instance. Compiling only requires `com.dylibso.chicory.wasm.Module`. `Instance` is used afterwards to bind the instance to the compiled methods. This would allow us to have a Maven plugin that compiles the bytecode ahead of time to ship with the application, cache the compiled classes across JVMs, etc.

electrum, line 15: After merging #416:

```suggestion
        .withMachineFactory(AotMachine::new)
```

electrum:
> I think it works because that module is an instance var in scope?

It "works" because the `MachineFactory` is not used during the construction of `runtime.Module`. If it was, this would fail with NPE. It is used later, when the module is instantiated. The simplified code looks like this:

```java
this.module = Module.builder()
        .withMachineFactory(instance -> new AotMachine(module.wasmModule(), instance))
        .build();
```

The reference `module.wasmModule()` inside the lambda reads the field `module`. If the field isn't set at the time the lambda is created, then why isn't the captured value `null`? Furthermore, why are we allowed to capture this field before it is set, when Java requires lambda captures to be effectively final?

The answer is that the lambda isn't capturing the `module` field! The reference to `module` has an implicit `this`, so it is actually doing `this.module.wasmModule()`. The lambda is capturing `this`, which is of course final. The lambda is used after `module` is set, thus the `this.module` reference reads the correct value.

Our APIs shouldn't rely on Java puzzlers, so let's fix this!

andreaTP:
@bhelx do you want to test this piece of code with `jbang` as we do in the other readme?
I can follow up with it if necessary.

bhelx:
> @bhelx do you want to test this piece of code with jbang as we do in the other readme?
I can follow up with it if necessary.

I think it's okay for the time being since it's still experimental. When we move it to the main readme we can have it tested then. Up to you though. Doesn't hurt to add a test.

bhelx:
> Our APIs shouldn't rely on Java puzzlers, so let's fix this!

Always happy to see when my Java ignorance actually helps make things better 😆

PR #416, opened 2024-06-30
Simplify AotMachine constructor

electrum:


andreaTP (approved):
Much nicer thanks! LGTM!

Still not 100% convinced about the dance we keep doing(me included) `instance.module().wasmModule()` but not a concern for this PR :+1:

PR #421, opened 2024-07-01
Most of assert invalid wast files

andreaTP:
I started enjoining too much the process, probably ... :smile: 

Screenshot from 2024-07-01 17-10-34

### DIFF
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/ConstantEvaluators.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/ConstantEvaluators.java
@@ -1,7 +1,9 @@
 package com.dylibso.chicory.runtime;
 
 import com.dylibso.chicory.wasm.exceptions.ChicoryException;
+import com.dylibso.chicory.wasm.exceptions.InvalidException;
 import com.dylibso.chicory.wasm.types.Instruction;
+import com.dylibso.chicory.wasm.types.MutabilityType;
 import com.dylibso.chicory.wasm.types.Value;
 import com.dylibso.chicory.wasm.types.ValueType;
 import java.util.Arrays;
@@ -56,7 +58,13 @@ public class ConstantEvaluators {
                     }
                 case GLOBAL_GET:
                     {
-                        return instance.readGlobal((int) instruction.operands()[0]);
+                        var idx = (int) instruction.operands()[0];
+                        if (idx < instance.imports().globalCount() && instance.imports().global(idx).mutabilityType() != MutabilityType.Const) {
+                            throw new InvalidException(
+                                    "constant expression required, initializer expression"
+                                            + " cannot reference a mutable global");
+                        }
+                        return instance.readGlobal(idx);
                     }
                 case END:
                     {
@@ -64,13 +72,13 @@ public class ConstantEvaluators {
                     }
                 default:
                     {
-                        throw new ChicoryException(
-                                "Non-constant instruction encountered: " + instruction);
+                        throw new InvalidException(
+                                "constant expression required, but non-constant instruction encountered: " + instruction);
                     }
             }
         }
         if (tos == null) {
-            throw new ChicoryException("No constant value loaded");
+            throw new InvalidException("type mismatch, expected constant value");
         }
         return tos;
     }
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/ConstantEvaluators.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/ConstantEvaluators.java
@@ -58,14 +58,21 @@ public class ConstantEvaluators {
                 case GLOBAL_GET:
                     {
                         var idx = (int) instruction.operands()[0];
-                        if (idx < instance.imports().globalCount()
-                                && instance.imports().global(idx).mutabilityType()
-                                        != MutabilityType.Const) {
+                        if (idx < instance.imports().globalCount()) {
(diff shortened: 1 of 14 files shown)


runtime/src/main/java/com/dylibso/chicory/runtime/InterpreterMachine.java
```diff
@@ -1787,13 +1786,6 @@ private static void TABLE_GET(MStack stack, Instance instance, long[] operands)
 
     private static void GLOBAL_SET(MStack stack, Instance instance, long[] operands) {
         var id = (int) operands[0];
-        var mutabilityType =
```
andreaTP, line 1790: - note to self - At the end of this process, we should review the Machine to remove any other redundant check.

runtime/src/main/java/com/dylibso/chicory/runtime/ConstantEvaluators.java
```diff
@@ -51,26 +52,45 @@ public static Value computeConstantValue(Instance instance, List<Instruction> ex
                     }
                 case REF_FUNC:
                     {
-                        tos = Value.funcRef((int) instruction.operands()[0]);
+                        var idx = (int) instruction.operands()[0];
+                        instance.function(idx);
```
electrum, line 56: `function()` returns null for negative values, so this will succeed. Should this fail for negative values?

runtime/src/main/java/com/dylibso/chicory/runtime/Instance.java
```diff
@@ -129,73 +132,82 @@ public Instance initialize(boolean start) {
...
                 }
+            } else if (el instanceof DeclarativeElement) {
+                var de = (DeclarativeElement) el;
+
+                List<List<Instruction>> initializers = de.initializers();
+                for (int i = 0; i < initializers.size(); i++) {
+                    final List<Instruction> init = initializers.get(i);
+                    var value = computeConstantValue(this, init);
+                    var func = function(value.asFuncRef());
+                    if (de.type() == ValueType.FuncRef && func != null) {
```
electrum, line 191: Related to the above comment, should `function()` throw on invalid values rather than returning null? With the current code, an invalid function index here is ignored.

runtime/src/main/java/com/dylibso/chicory/runtime/Instance.java
```diff
@@ -297,8 +317,8 @@ public Value readGlobal(int idx) {
             return imports.global(idx).instance().getValue();
         }
         var i = idx - importedGlobalsOffset;
-        if (i < 0 || i >= globals.length) {
-            throw new InvalidException("unknown global " + i);
+        if (i < 0 || i >= globals.length || globals[idx - importedGlobalsOffset] == null) {
```
electrum, line 320: How do we have a null global for a valid array index?

runtime/src/main/java/com/dylibso/chicory/runtime/InterpreterMachine.java
electrum: Yes! When writing the AOT, I noticed various checks in the interpreter that seemed like they should have been handled during validation. I should have kept notes.

runtime/src/main/java/com/dylibso/chicory/runtime/Module.java
```diff
@@ -61,15 +61,24 @@ private Module(
         this.start = start;
         this.typeValidation = typeValidation;
         this.module = validateModule(module);
-        this.exports = new HashMap<>();
-        int cnt = module.exportSection().exportCount();
+        this.exports = genExports(module.exportSection());
+    }
+
+    private static HashMap<String, Export> genExports(ExportSection export) {
```
electrum, line 67: Nit: we could change `exports` to be declared as `Map` instead of `HashMap`, then this could return `Map`

electrum (approved):

runtime/src/main/java/com/dylibso/chicory/runtime/Module.java
andreaTP: right! done

runtime/src/main/java/com/dylibso/chicory/runtime/ConstantEvaluators.java
andreaTP: Addressed in this way: https://github.com/dylibso/chicory/pull/421/commits/a423f05ab0ac5dd55a193b4909e74cfc0c117ce6

electrum: Nice, that looks better

runtime/src/main/java/com/dylibso/chicory/runtime/Instance.java
electrum: I just get nervous seeing nulls in unexpected places. If this is expected, then no problem

andreaTP: It's actually a good catch: https://github.com/dylibso/chicory/pull/421/commits/9a751db8b2df27ae1547b013dc65913bf699f60f

Thanks!

runtime/src/main/java/com/dylibso/chicory/runtime/InterpreterMachine.java
andreaTP: Tracked in #422

PR #425, opened 2024-07-03
Minor optimizations for AOT method size

electrum:


aot/src/main/java/com/dylibso/chicory/aot/AotMachine.java
```diff
@@ -462,14 +464,23 @@ private Class<?> compileClass(FunctionSection functions) {
 
         classWriter.visitEnd();
 
-        return loadClass(className, classWriter.toByteArray());
+        try {
+            return loadClass(className, classWriter.toByteArray());
+        } catch (MethodTooLargeException e) {
```
andreaTP, line 469: :+1:

andreaTP (approved):
LGTM thanks!

To review this kind of PRs I do need to squint my eyes.
It would be great to be able to compare examples of the generated bytecode.
Ideally, we can have [approval tests](https://github.com/approvals/ApprovalTests.Java), on a few selected demo modules.

electrum:
I rebased this on #426 and added additional approval tests, which makes the bytecode changes easy to see.

andreaTP (approved):
LGTM!
As a general advice, approval tests are nice but tend to become high maintenance (e.g. generating larger diffs on PRs).
My rule of thumb is to stay with the minimum amount to show a proper diff when something changes but don't try to track each and every detail.
I'm ok with this if it's the first "corpus", but, please, don't abuse them :-)

electrum:
Agreed! We have similar types of tests in Trino, and used to have a lot more for query plans, but deleted them due to the maintenance problem. I think we'll need to play with this to determine the right balance.

After rebasing this PR and seeing the diffs to the approval tests, it was more interesting and helpful than I expected, so I'm glad that you added this.

PR #426, opened 2024-07-03
Basic approval tests for AOT generated code

andreaTP:
Following up: https://github.com/dylibso/chicory/pull/425#pullrequestreview-2156046227

### DIFF
--- a/aot/src/test/java/com/dylibso/chicory/approvals/PackageSettings.java
+++ b/aot/src/test/java/com/dylibso/chicory/approvals/PackageSettings.java
@@ -1,0 +1,6 @@
+package com.dylibso.chicory.approvals;
+
+public class PackageSettings {
+    public String UseApprovalSubdirectory = "approvals";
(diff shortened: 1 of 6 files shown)


aot/src/test/java/com/dylibso/chicory/approvals/PackageSettings.java
```diff
@@ -0,0 +1,6 @@
+package com.dylibso.chicory.approvals;
+
+public class PackageSettings {
+    public String UseApprovalSubdirectory = "approvals";
```
electrum, line 4: We have the tests under `chicory/approvals` with the results in `chicory/approvals/approvals` which seems redundant. I can see putting the results in a subdirectory if we didn't already use an `approvals` package for the tests.

electrum (approved):
We can follow up on the duplicated directory name. I'll merge this for now so that I can rebase #425 which might change as the result of this.

PR #428, opened 2024-07-05
Builder helper class for HostImports

### LINKED ISSUE #357: Host Functions not very composable
I've had this issue where working with host functions and imports can be awkward. For example. if i want to support both wasi and some custom host functions, i need to compose them like this:

```java
HostFunction[] extra = createMyCustomHostFunctions();
HostFunction[]  wasiFuncs = wasi.toHostFunctions();
HostFunction[] allImports = new HostFunction[wasiFuncs.length + extra.length];
System.arraycopy(wasiFuncs, 0, allImports, 0, wasiFuncs.length);
System.arraycopy(extra, 0, allImports, wasiFuncs.length, extra.length);

var imports = new HostImports(allImports);
```

I can think of a number of ways to address this. A quick fix might be to allow the HostImports constructor to take an arraylist. then we turn it into an array in the constructor. Or there may be a better interface to build and compose these that we should consider.
Now might be a good time to create a builder interface for host modules. I think wazero has a decent interface for this: https://github.com/tetratelabs/wazero/blob/main/examples/import-go/age-calculator.go#L41 but of course we could do it in a more idomatic Java way. I can easily do the first solution but I felt I'd give an opportunity for some of the experienced Java developers on the team to bring their opinion on a good interface.

manusa:
Hi all,

I'm getting familiar with this repository and wanted to start by contributing a very small improvement.

In this case, I'm tackling #357, since @andreaTP said it could be a good way to start.

Here I'm adding a `HostImportsBuilder` class that can help with the situation described in #357.

If this is OK and we merge it, further refinements can be done in the HostImports class to remove redundant constructors. Similar tactics could be used for other classes too.

(might fix #357)

### DIFF
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/HostImports.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/HostImports.java
@@ -1,5 +1,9 @@
 package com.dylibso.chicory.runtime;
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
 public class HostImports {
     private static final HostFunction[] NO_HOST_FUNCTIONS = new HostFunction[0];
     private static final HostGlobal[] NO_HOST_GLOBALS = new HostGlobal[0];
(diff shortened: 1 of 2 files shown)


runtime/src/main/java/com/dylibso/chicory/runtime/HostImportsBuilder.java
```diff
@@ -0,0 +1,94 @@
+package com.dylibso.chicory.runtime;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+public final class HostImportsBuilder {
```
andreaTP, line 7: Nit: we have not "streamlined" yet an approach in this repo:
- external Builder class
- internal `.Builder`

I have a slight preference for the latter but happy with whatever we find an agreement on.
Cc. @electrum @danielperano @bhelx

andreaTP (approved):
One minor nit that doesn't need to be tackled here and LGTM, an improvement over the current situation anyhow!

andreaTP (approved):

runtime/src/main/java/com/dylibso/chicory/runtime/HostImportsBuilder.java
electrum: I prefer the internal. Seems cleaner and more common.

manusa: I changed the commit to reflect the second approach.

andreaTP (approved):
LGTM thanks!

PR #434, opened 2024-07-11
refactor: remove annoying obsolete log

ablax:
System.out.println("Skipping Section with ID due to configuration: " + sectionId);
Is a public log that we cannot hide, so it would be optimal to use the provided logger to log it.

That set aside, I belive firmly that the provided logger should be optional and nullable. If you disagree with me, I can return the logger to be nonNull, but the sysout should be changed to logger.

### DIFF
--- a/wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java
+++ b/wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java
@@ -72,6 +72,10 @@ public final class Parser {
     private final BitSet includeSections;
     private final Logger logger;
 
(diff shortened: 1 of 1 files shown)


Zurcusa (approved):

andreaTP (changes requested):
Thanks for your interest in the project!

I don't think that we should make the logger nullable, but I agree that we should use the logger instead of System.out, good catch!
Other than that, it should be possible to implement an alternative Listener to better customize the behavior.

andreaTP (approved):
LGTM thanks!

PR #439, opened 2024-07-23
Emitboxarguments forloop

DavideCeto:
AotMachine class - emitBoxArguments method: slots ArrayList has been moved into for loop, only for optimization pourposes.

### DIFF
--- a/aot/src/main/java/com/dylibso/chicory/aot/AotMachine.java
+++ b/aot/src/main/java/com/dylibso/chicory/aot/AotMachine.java
@@ -56,7 +56,6 @@ import java.io.PrintWriter;
 import java.lang.invoke.MethodHandle;
 import java.lang.invoke.MethodType;
 import java.util.ArrayDeque;
-import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


andreaTP (approved):
lgtm, thanks!

aot/src/main/java/com/dylibso/chicory/aot/AotMachine.java
```diff
@@ -593,21 +593,18 @@ private static void compileHostFunction(int funcId, FunctionType type, MethodVis
...
-        for (ValueType type : types) {
-            slots.add(slot);
-            slot += slotCount(type);
-        }
-
         // box the arguments into Value[]
         asm.visitLdcInsn(types.size());
         asm.visitTypeInsn(Opcodes.ANEWARRAY, getInternalName(Value.class));
         for (int i = 0; i < types.size(); i++) {
+            slots.add(slot);
```
electrum, line 600: We don't need the list at all anymore

andreaTP (approved):

PR #440, opened 2024-07-23
Finish all the validation for the interpreter

andreaTP:
I'll clean this up tomorrow, there are probably leftovers around as I was going back and forth fixing stuffs.

Screenshot from 2024-07-23 19-10-13

### DIFF
--- a/aot/src/test/java/com/dylibso/chicory/imports/SpecV1ElemHostFuncs.java
+++ b/aot/src/test/java/com/dylibso/chicory/imports/SpecV1ElemHostFuncs.java
@@ -10,76 +10,53 @@ import com.dylibso.chicory.runtime.HostImports;
 import com.dylibso.chicory.runtime.HostTable;
 import com.dylibso.chicory.runtime.TableInstance;
 import com.dylibso.chicory.wasm.types.Limits;
+import com.dylibso.chicory.wasm.types.MutabilityType;
 import com.dylibso.chicory.wasm.types.Table;
 import com.dylibso.chicory.wasm.types.Value;
 import com.dylibso.chicory.wasm.types.ValueType;
 
 public class SpecV1ElemHostFuncs {
 
+    private static GlobalInstance glob = new GlobalInstance(Value.i32(123));
+
     public static HostImports fallback() {
-        return new HostImports();
-    }
-
-    public static HostImports testModule3() {
-        return new HostImports(
-                new HostTable[] {
-                    new HostTable(
-                            "spectest",
-                            "table",
-                            new TableInstance(new Table(ValueType.FuncRef, new Limits(1))))
-                });
-    }
-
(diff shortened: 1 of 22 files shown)


aot/src/test/java/com/dylibso/chicory/imports/SpecV1ElemHostFuncs.java
```diff
@@ -10,14 +10,30 @@
...
 import com.dylibso.chicory.wasm.types.Table;
 import com.dylibso.chicory.wasm.types.Value;
 import com.dylibso.chicory.wasm.types.ValueType;
 
 public class SpecV1ElemHostFuncs {
 
     public static HostImports fallback() {
-        return new HostImports();
+        return HostImports.builder()
+                .addGlobal(
```
andreaTP, line 22: note to self: this API is better than before, but is still too verbose and detailed for 90% of the use cases.

runtime-tests/src/test/java/com/dylibso/chicory/imports/SpecV1BinaryLeb128HostFuncs.java
```diff
@@ -0,0 +1,24 @@
...
+import com.dylibso.chicory.wasm.types.Value;
+import com.dylibso.chicory.wasm.types.ValueType;
+import java.util.List;
+
+public class SpecV1BinaryLeb128HostFuncs {
+    public static HostImports fallback() {
+        return HostImports.builder()
+                .addFunction(
+                        new HostFunction(
+                                (Instance instance, Value... args) -> {
```
andreaTP, line 15: note to self: `HostFunction` constructor is the only `Host` constructor that doesn't take the two strings for the name as first arguments

aot/src/test/java/com/dylibso/chicory/imports/SpecV1ElemHostFuncs.java
bhelx: Agreed, i have some words to say about this in #441

electrum:
I'm traveling for the next two days and can review when I return home.

It will be easier to review if you can split out the API refactoring to a separate commit from the functional changes.

andreaTP:
Thanks for the feedback @electrum !

I might try to separate in 2 commits, but is challenging, apart from a few trivial things, the biggest part of it is driven by this change: https://github.com/dylibso/chicory/pull/440/files#diff-fadb7963eded2e2c77d87d70b45f81ab8b4ea0cffff5a68695cead4732c5f6b1R225

We had a bug in the tests generation and now the generated code looks different.

bhelx (approved):

electrum (approved):

andreaTP:
Thanks for the reviews! Appreciated!

PR #443, opened 2024-07-26
Fix issue 401

### LINKED ISSUE #401: Validate interpreter handling of stack unwinds
While debugging `fac-ssa` from `fac.wast`, I noticed that the stack continues to grow. It appears that the interpreter is not removing the initial loop parameters when unwinding. We could validate that the stack is empty when the function returns.

andreaTP:
I ended up rewriting the entire control flow in a way that maps closely to the `TypeValidator` implementation.

Fix #401


bhelx:
The SNES emulator was crashing after 100 frames or so on the interpreter. Now working on this PR!

Screenshot 2024-07-26 at 9 10 00 AM


I think this PR fixed the stack effect bug :)

bhelx (approved):
Looks good to me! Would be good to get a second pair of eyes on someone more familiar with the control flow algorithm.

runtime/src/main/java/com/dylibso/chicory/runtime/StackFrame.java
```diff
@@ -89,30 +87,63 @@ public boolean terminated() {
...
 
-    public void endOfNonControlBlock() {
-        if (currentInstruction.depth() > 0) {
-            stackSizeBeforeBlock.pop();
+    public void jumpTo(int newPc) {
+        pc = newPc;
+    }
+
+    public static void doControlTransfer(CtrlFrame ctrlFrame, MStack stack) {
+        var endResults = ctrlFrame.startValues + ctrlFrame.endValues; // fix 401
```
electrum, line 132:
```java
// unwind stack

electrum (approved):

mariofusco (approved):
Nice rework, much clearer now.

wasm-corpus/src/main/resources/wat/fac.wat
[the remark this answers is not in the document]
andreaTP: done

runtime/src/main/java/com/dylibso/chicory/runtime/StackFrame.java
andreaTP: changed

Issue #441, opened 2024-07-23
Version 1.0 API


bhelx:
This issue is for collecting ideas and opinions about what our 1.0 API should look like. The primary focus is on the user-facing API. Though some decisions may have consequences for the internal APIs. We need to choose this direction and get out a beta in the coming months. From there we can iterate until we feel ready to public a 1.0. Once we can publish a stable API, we can start building some of the nicer layers on top and we can better support people in production.

This issue will have both a goals and a tasks section. I'll add items to both as we discuss in the team meetings and get a consensus on the ideas. Please issue your ideas and proposals as comments.

## Goals

- Security
  + Are we doing our best to make it clear how to use Chicory securely?
  + Are there any footguns that might make it easy to apply an insecure pattern?
- Stability
  + Are we ensuring that our API allows all the things users may need to do?
  + Are we ensuring that our API can support future specs without breaking it?
- Familiarity and Compatibility
  + Are we borrowing from other runtimes as much as possible?
  + Will the API be familiar to someone who has used a Wasm runtime in anger before?
- Performance
  + We don't need the most performant interface
  + But does what we choose lock us into performance we cannot upgrade?
  + How can we bring performance improvements without interrupting users?
  
  
## Tasks

TODO

andreaTP:
Additional task here, we should, at least verify, that we are not breaking Java's "module system" good practices for package naming etc.

andreaTP:
I think we can now close this one now, the bulk of the work has been done and there are issues opened for the rest.
Feel free to re-open if you think there is still something missing.

PR #445, opened 2024-07-26
RFC - Human friendly HostImports' builders

andreaTP:
Since code is worth a thousand words, I started coding a bunch of `HostImports`' `Builders` to see how far I could get with a small corpus of examples.

Notes:
- I avoided the usage of reflection
- This implementation is mostly bound to the [specialized functional interfaces](https://docs.oracle.com/en%2Fjava%2Fjavase%2F11%2Fdocs%2Fapi%2F%2F/java.base/java/util/function/package-summary.html) it will be a bit of code to map everything, but I think it's something manageable
- The focus has been solely on DX(e.g. let's keep the discussion around performance to a later iteration)

Happy to hear comments!

### DIFF
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/HostImports.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/HostImports.java
@@ -1,8 +1,21 @@
 package com.dylibso.chicory.runtime;
 
+import com.dylibso.chicory.wasm.types.Limits;
+import com.dylibso.chicory.wasm.types.MemoryLimits;
+import com.dylibso.chicory.wasm.types.MutabilityType;
+import com.dylibso.chicory.wasm.types.Table;
+import com.dylibso.chicory.wasm.types.Value;
+import com.dylibso.chicory.wasm.types.ValueType;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.function.Function;
+import java.util.function.IntConsumer;
+import java.util.function.IntSupplier;
+import java.util.function.IntUnaryOperator;
+import java.util.function.LongConsumer;
+import java.util.function.LongSupplier;
+import java.util.function.LongUnaryOperator;
 
 public class HostImports {
     private static final HostFunction[] NO_HOST_FUNCTIONS = new HostFunction[0];
@@ -137,6 +150,10 @@ public class HostImports {
         return new Builder();
     }
 
+    public static Builder builder(HostImports imports) {
+        return new Builder(imports);
+    }
+
     public static final class Builder {
         private List<HostFunction> functions;
         private List<HostGlobal> globals;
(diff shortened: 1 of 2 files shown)


runtime/src/test/java/com/dylibso/chicory/runtime/HostImportsTest.java
```diff
@@ -177,5 +181,82 @@ void addIndex() {
...
             }
         }
+
+        @Nested
+        class DSL {
+
+            @Test
+            void withIndex() {
+                var moduleName = "module";
+                var fieldName = "filed";
```
electrum, line 191: Typo "filed"?

runtime/src/test/java/com/dylibso/chicory/runtime/HostImportsTest.java
```diff
@@ -177,5 +181,82 @@ void addIndex() {
...
+                HostImports.builder()
+                        .withNewImport(moduleName, fieldName)
```
electrum, line 193: I find this style of nested builder to be strange, especially since only one method can be called. For nested builders, I've found that having a `Consumer` argument is a good approach, but that doesn't seem useful here. The only thing we are saving by having `withNewImport()` is the two name arguments.

All of the `with` methods on `Builder` replace the entire collection. For example, `withGlobals()` replaces all of the globals, whereas `addGlobal()` adds additional globals. So the name `withNewImport()` seems inconsistent.

Instead, what if we put new methods directly on `Builder`:
```java
.addGlobal(moduleName, fieldName, Value.i32(1))
.addGlobal(moduleName, fieldName, MutabilityType.Var, Value.i32(1))
.addMutableGlobal(moduleName, fieldName, Value.i32(1))
.addMemory(moduleName, fieldName)
.addMemory(moduleName, fieldName, new MemoryLimits(1))
.addMemory(moduleName, fieldName, 1) // I'd drop these as the MemoryLimits version seems more clear
.addMemory(moduleName, fieldName, 1, 2)
.addTable(moduleName, fieldName)
.addTable(moduleName, fieldName, ValueType.ExternRef)
.addTable(moduleName, fieldName, new Limits(1))
.addTable(moduleName, fieldName, 1) // same, I'd drop these and use the Limits version
.addTable(moduleName, fieldName, 1, 2)
// does type inference work out if we name all of these "addFunction"?
.addProcedure(moduleName, fieldName, () -> System.out.println("hello world"))
.addProcedure(moduleName, fieldName, (Instance inst) -> () -> System.out.println("hello world"))
.addSupplier(moduleName, fieldName, () -> 1)

Comment (andreaTP) [reply runtime/src/test/java/com/dylibso/chicory/runtime/HostImportsTest.java:193]: Thanks for your input @electrum !
A few different subjects:

- `add` instead of `with`: agree will change accordingly, thanks for noticing!
- `two name arguments`: personally, I feel it awkward to have to type those "names" along with the implementation, the nested builder was an attempt to fix it.
- `new methods directly on Builder`: this works for sure, I'm afraid is an improvement less impactful than I originally thought

File: runtime/src/main/java/com/dylibso/chicory/runtime/HostImports.java [hunk 3]
```diff
@@ -146,6 +163,245 @@ public static final class Builder {
...
+                        new HostFunction(
+                                (Instance instance, Value... args) ->
+                                        new Value[] {Value.i32(f.getAsInt())},
+                                moduleName,
+                                fieldName,
+                                List.of(),
+                                List.of(ValueType.I32)));
+            }
+
+            public Builder withSupplier(Function<Instance, IntSupplier> consumer) {
```
andreaTP, line 311: The more I look at this approach, the less I'm convinced this is the way to go.

```java
public Builder withSupplier(Function<Instance, IntSupplier> consumer)
```
and
```java
public Builder withSupplier(Function<Instance, LongSupplier> consumer)
```
are going to be not distinguishable after type erasure.

Currently, I think that leaving only:

```java
public Builder withFunction(Function<Instance, Function<Value[], Value[]>> consumer)
```

and generating the binding out of real modules with something like: https://github.com/andreaTP/chicory-bindgen-poc is the best bet.

bhelx:
Something I wanted to chat about in regards to #441 is that building a imports are really just another module that you link to. It feels like we could only have one way to build a module internally. Then we could maybe have some helpers to quickly build these when you just want a couple host functions. I'm just not sure there needs to be a `HostImports` object at all.

andreaTP:
@bhelx thanks for sharing! Super interesting POV indeed.
A couple of things on top of my head:

- other than "using the Parser" we don't have a "great" API to build Modules internally, this seems a prerequisite to your proposal
- we should make it extremely convenient to define Host Functions

I'm happy to see further exploration of the idea, thought! Seems appealing!

andreaTP:
A couple of additional cents:

> building a imports are really just another module that you link to

correct, at the moment all the `HostXXX` are implementing the `FromHost` `interface`, here we are basically adding the 2 module and field names.
One option is to make those fields optional on the `Memory`, `Global`, etc. classes, or there is a better option?

bhelx:
The way i see it, host imports are just a type of Module that whose instance is not a wasm instance, but lives on the host. But it could / should have the same API as any other module you might want to link (e.g. other wasm modules which must be instantiated as wasm). You can run into situations where you might need to link up a mix of both host and wasm modules to instantiate a module.

I need some time to study how others runtimes do it, but ideally there could be a low-level, imperative API for building up an instance, and a high level API for when you just want to throw some host functions at a module.

We'd need some things like:

1. A builder API for building a Module (host or wasm)
2. An api for instantiating these
  * we mostly have this, my only point in bringing it up is cases where we need to instantiate a linked wasm module
3. Some kind of Linker which can dynamically link any set of wasm (or host) modules
  * The linker could handle the complexity of lining up all the imports / exports, validating, and instantiating the modules

For the higher level, shorthand API, where you maybe just want to pass some host functions. We can support passing functions to the instantiation process (kind of like how we do now). You wouldn't need the linker you could just use the normal module builder and instantiator.

bhelx:
This is a lower-level reference doc, but here is the API for the wasmtime linker https://docs.wasmtime.dev/api/wasmtime/struct.Linker.html

andreaTP:
I see, at the moment cross-linking modules is possible but challenging for sure.
Do you have/want to craft real-world use cases where this functionality is needed?

I'm a bit afraid that it might be a bit "too early" to "future-proof" something we don't entirely grasp.

runtime/src/test/java/com/dylibso/chicory/runtime/HostImportsTest.java
electrum: > I feel it awkward to have to type those "names" along with the implementation

This test might look worse due to repeating the same variables. In actual usage, the field names will need to be unique (and the builder should validate that)? So if we change it to use constants, does it look better?

```java
.addGlobal("test", "g1", Value.i32(1))
.addGlobal("test", "g2", MutabilityType.Var, Value.i32(1))
.addMutableGlobal("test", "gmut", Value.i32(1))
.addMemory("test", "m1")
.addMemory("test", "mlim", new MemoryLimits(1))
.addTable("test", "t1")
.addTable("test", "tref", ValueType.ExternRef)
.addTable("test", "t3", new Limits(1))
```

Another problem is that the code formatter forces each chained call to be on a separate line. If we were formatting the code by hand, it would look better:
```java
.add("test", "g1").global(Value.i32(1))
.add("test", "g2").global(MutabilityType.Var, Value.i32(1))
.add("test", "gmut").mutableGlobal(Value.i32(1))
.add("test", "m1").memory()
.add("test", "mlim").memory(new MemoryLimits(1))
.add("test", "t1").table()
.add("test", "tref").table(ValueType.ExternRef)
.add("test", "t3").table(new Limits(1))
```
Which I agree looks slightly cleaner, but we can't format like that...

andreaTP: I love this!

```java
.add("test", "g1").global(Value.i32(1))
```
I believe that we should not be limited by our own formatter, and we can always disable it when necessary :smirk:

evacchi:
I'm starting to feel like `HostImports` (now "`ExternalValues`") could be subclassed(*) to `HostModule` (#482) and this builder would be the "low-level" builder for a `HostModule`. Then #496 would be the higher-level, user-friendly version of it.

(*) well not necessarily, we can just let the `HostModule` have a `toExternalValues()` method.

andreaTP:
This is too outdated, let's close it and move on, the `Store` helps a lot in smoothening the experience and the upcoming code-gens are going to become the standard way to define `ExternalValues`.

PR #446, opened 2024-07-26
[chore] add wasm.Module Builders to conveniently run the parser

andreaTP:


wabt/src/main/java/com/dylibso/chicory/wabt/Wat2Wasm.java
```diff
@@ -30,11 +30,9 @@
 public final class Wat2Wasm {
     private static final Logger logger = new SystemLogger();
     private static final com.dylibso.chicory.wasm.Module wasmModule =
-            Module.builder(Wat2Wasm.class.getResourceAsStream("/wat2wasm"))
-                    .withInitialize(false)
-                    .withStart(false)
-                    .build()
-                    .wasmModule();
+            com.dylibso.chicory.wasm.Module.builder(Wat2Wasm.class.getResourceAsStream("/wat2wasm"))
```
andreaTP, line 33: I'm not 100% convinced of having 2 `Module` classes around, but I don't have a "clearly better" proposal to change the name.

wasm/src/main/java/com/dylibso/chicory/wasm/Module.java
```diff
@@ -151,4 +166,150 @@ public ElementSection elementSection() {
...
+     *
+     * @param classpathResource the name of the resource
+     * @return a {@link Builder}  for reading the module definition from the specified resource
+     */
+    public static Builder builder(String classpathResource) {
+        return new Builder(
+                () -> {
+                    InputStream is =
+                            Thread.currentThread()
+                                    .getContextClassLoader()
```
electrum, line 239: Relying on the thread context class loader is problematic, as we saw with `Wast2Json`. We should consider removing this API and let callers use `Class.getResourceAsStream()` or similar, as appropriate. This avoids having a hidden dependency on the context class loader. If the user wants that, it will be explicit in their application code.

wasm/src/main/java/com/dylibso/chicory/wasm/Module.java
```diff
@@ -151,4 +166,150 @@ public ElementSection elementSection() {
...
+            this.logger = logger;
+            return this;
+        }
+
+        public Builder withType(ModuleType type) {
+            this.moduleType = type;
+            return this;
+        }
+
+        public Module build() {
```
electrum, line 292: Update `com.dylibso.chicory.runtime.Module.Builder#build` to use this?

electrum (approved):

wabt/src/main/java/com/dylibso/chicory/wabt/Wat2Wasm.java
electrum: I think we can remove `runtime.Module` as it's basically just a builder for `Instance`.

wasm/src/main/java/com/dylibso/chicory/wasm/Module.java
andreaTP: I agree, doing the changes

wabt/src/main/java/com/dylibso/chicory/wabt/Wat2Wasm.java
andreaTP: I'm trying this approach, I'll update this PR as soon as I'm done, thanks for the suggestion!

aot/src/main/java/com/dylibso/chicory/aot/AotMachine.java
```diff
@@ -39,7 +39,7 @@
 import com.dylibso.chicory.runtime.OpcodeImpl;
 import com.dylibso.chicory.runtime.StackFrame;
 import com.dylibso.chicory.runtime.exceptions.WASMRuntimeException;
-import com.dylibso.chicory.wasm.Module;
```
andreaTP, line 42: Changed to avoid the overlap with `java.lang.Module`

cli/src/main/java/com/dylibso/chicory/cli/Cli.java
```diff
@@ -66,31 +68,19 @@ public void run() {
...
-                                + type.params().size()
-                                + " parameters, but found: "
-                                + arguments.length);
-            }
-            var params = new Value[type.params().size()];
-            for (var i = 0; i < type.params().size(); i++) {
-                params[i] = new Value(type.params().get(i), Long.valueOf(arguments[i]));
+            var params = new Value[arguments.length];
+            for (var i = 0; i < arguments.length; i++) {
+                // TODO: FIXME - it's hard to extract the exported function signature
```
andreaTP, line 81: this is the only "rough edge" that, in my opinion, is left from this PR.

wabt/src/main/java/com/dylibso/chicory/wabt/Wat2Wasm.java
andreaTP: Done, I'm happy with the result!

aot/src/main/java/com/dylibso/chicory/aot/AotMachine.java
electrum: I don't think we need to rename, as `java.lang.Module` is not typically used in application code. We use `com.google.inject.Module` everywhere and the name overlap doesn't cause problems.

That said, I'm not against the rename if we think it's better. Let's do it as a separate commit or PR.

wasm/src/main/java/com/dylibso/chicory/wasm/WasmModule.java
```diff
@@ -0,0 +1,294 @@
...
+                () -> {
+                    try {
+                        return Files.newInputStream(path);
+                    } catch (IOException e) {
+                        throw new IllegalArgumentException("Error opening file: " + path, e);
+                    }
+                });
+    }
+
+    public static class Builder {
```
electrum, line 245: I dislike having a builder just to read a file. The builder only has two options: logger and file format.

The logger is only used in one place, to warn about ignored section IDs. If anyone cares about that, we could collect them and have a method `getIgnoredSectionIds()`. Or perhaps we should fail, since an unknown ID means the file is malformed or non-compliant. We can see what other implementations do. Thus, we can remove the logger.

The file format is not used today. It's also unclear if we would want to load the text format using the same API. The browser API only supports the binary format. Using `Wat2Wasm` or similar to covert to binary first seems like a reasonable answer. Another future option is to add auto-detection: try to parse as text if the file does not start with the binary magic value.

With those changes, we can have a simpler API:
```java
WasmModule module = WasmModule.load(path);
```
All of the `builder()` methods become `load()` methods that return `WasmModule`.

README.md
```diff
@@ -171,7 +171,7 @@ copyFileFromWasmCorpus("count_vowels.rs.wasm", "count_vowels.wasm");
 Build and instantiate this module:
 
 ```java
-Instance instance = Module.builder(new File("./count_vowels.wasm")).build().instantiate();
+Instance instance = Instance.builder(WasmModule.builder(new File("./count_vowels.wasm")).build()).build();
```
electrum, line 174:
```suggestion
var module = WasmModule.load(new File("./count_vowels.wasm"));
Instance instance = Instance.builder(module).build();
```

README.md
```diff
@@ -278,7 +278,7 @@ Now we just need to pass this host function in during our instantiation phase:
 ```java
 import com.dylibso.chicory.runtime.HostImports;
 var imports = new HostImports(new HostFunction[] {func});
-var instance = Module.builder(new File("./logger.wasm")).withHostImports(imports).build().instantiate();
+var instance = Instance.builder(WasmModule.builder(new File("./logger.wasm")).build()).withHostImports(imports).build();
```
electrum, line 281:
```suggestion
var module = WasmModule.load(new File("./logger.wasm"));
var instance = Instance.builder(module).withHostImports(imports).build();```

File: aot/README.md [hunk 9]
```diff
@@ -13,9 +13,10 @@ To enable use the AotMachine factory when building the module:
...
 // ...
-import com.dylibso.chicory.runtime.Module;
+import com.dylibso.chicory.wasm.WasmModule;
+import com.dylibso.chicory.runtime.Instance;
 import com.dylibso.chicory.aot.AotMachine;
 // ...
-
-var module = Module.builder("compiled/basic.c.wasm").withMachineFactory(AotMachine::new).build();
+var is = Thread.currentThread().getContextClassLoader().getResourceAsStream("compiled/basic.c.wasm");
+Instance.builder(WasmModule.builder(is).build()).withMachineFactory(AotMachine::new).build();
```
electrum, line 21:
```suggestion
var input = getClass().getResourceAsStream("/compiled/basic.c.wasm");
var module = WasmModule.load(input);
Instance.builder(module).withMachineFactory(AotMachine::new).build();
```

aot/src/test/java/com/dylibso/chicory/approvals/ApprovalTest.java
```diff
@@ -63,12 +64,16 @@ public void verifyTrap() {
 
     private static void verifyGeneratedBytecode(String name) {
         var instance =
-                Module.builder("compiled/" + name)
+                Instance.builder(
+                                WasmModule.builder(
+                                                Thread.currentThread()
+                                                        .getContextClassLoader()
+                                                        .getResourceAsStream("compiled/" + name))
+                                        .build())
```
electrum, line 72:
```suggestion
                Instance.builder(
                                WasmModule.load(
                                                ApprovalTest.class.getResourceAsStream("/compiled/" + name)))
```

runtime/src/test/java/com/dylibso/chicory/runtime/ModuleTest.java
```diff
@@ -185,35 +198,18 @@ public void shouldCountVowels() {
...
         // check with: wasmtime src/test/resources/wasm/basic.c.wasm --invoke run
-        var instance = Module.builder("compiled/basic.c.wasm").build().instantiate();
+        var instance =
+                Instance.builder(WasmModule.builder(load("compiled/basic.c.wasm")).build()).build();
         var run = instance.export("run");
         var result = run.apply()[0];
         assertEquals(42, result.asInt());
     }
 
-    // @Test
```
electrum, line 194: Why commented out?

wasi/README.md
```diff
@@ -166,9 +167,8 @@ var wasi = new WasiPreview1(logger, WasiOptions.builder().build());
 // turn those into host imports. Here we could add any other custom imports we have
 var imports = new HostImports(wasi.toHostFunctions());
 // create the module and connect imports
-var module = Module.builder(new File("hello-wasi.wasm")).withHostImports(imports).build();
 // this will execute the module if it's a WASI command-pattern module
-module.instantiate();
+Instance.builder(WasmModule.builder(new File("hello-wasi.wasm")).build()).withHostImports(imports).build();
```
electrum, line 171: With the removal of the module builder, this becomes
```suggestion
var module = WasmModule.load(new File("hello-wasi.wasm");
Instance.builder(module).withHostImports(imports).build();
```

wasi/README.md
```diff
@@ -206,10 +206,8 @@ var wasi = new WasiPreview1(logger, wasiOpts);
 var imports = new HostImports(wasi.toHostFunctions());
 
 // greet-wasi is a rust program that greets the string passed in stdin
-var module = Module.builder(new File("greet-wasi.wasm")).withHostImports(imports).build();
-
 // instantiating will execute the module if it's a WASI command-pattern module
-module.instantiate();
+Instance.builder(WasmModule.builder(new File("greet-wasi.wasm")).build()).withHostImports(imports).build();
```
electrum, line 210:
```suggestion
var module = WasmModule.load(new File("greet-wasi.wasm"));
Instance.builder(module).withHostImports(imports).build();
```

wasm/src/main/java/com/dylibso/chicory/wasm/WasmModuleType.java
```diff
@@ -0,0 +1,6 @@
+package com.dylibso.chicory.wasm;
+
+public enum WasmModuleType {
```
electrum, line 3: If we remove the builder, I don't think this enum will be needed

electrum:
Sorry for the churn. Each round of improvements makes it easier to see the problems in the existing API. It keeps getting better.

aot/src/main/java/com/dylibso/chicory/aot/AotMachine.java
andreaTP: ok, I revert the change in this PR and we keep the conversation going.

cli/src/main/java/com/dylibso/chicory/cli/Cli.java
andreaTP: The validation is performed while building an `Instance`, the function signatures are accessible in the raw `wasm.Module`, just not as convenient as it used to be.

andreaTP:
> Sorry for the churn. Each round of improvements makes it easier to see the problems in the existing API. It keeps getting better.

No apology! Enjoining the process so far, even if it needs more iterations, I like the direction is going :+1:

runtime/src/test/java/com/dylibso/chicory/runtime/ModuleTest.java
```diff
@@ -5,33 +5,39 @@
...
         var iterFact = instance.export("iterFact");
         var result = iterFact.apply(Value.i32(5))[0];
         assertEquals(120, result.asInt());
     }
 
     @Test
     public void shouldSupportBrTable() {
-        var instance = Module.builder("compiled/br_table.wat.wasm").build().instantiate();
+        var instance =
+                Instance.builder(WasmModule.builder(load("compiled/br_table.wat.wasm")).build())
```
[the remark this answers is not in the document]
andreaTP: :+1:

wasm/src/main/java/com/dylibso/chicory/wasm/WasmModule.java
```diff
@@ -0,0 +1,294 @@
...
+
+    public DataCountSection dataCountSection() {
+        return dataCountSection;
+    }
+
+    public MemorySection memorySection() {
+        return memorySection;
+    }
+
+    public void setMemorySection(MemorySection memorySection) {
```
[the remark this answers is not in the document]
andreaTP: agree, but let's follow up, this is already big enough for an iteration.

wasm/src/main/java/com/dylibso/chicory/wasm/WasmModule.java
```diff
@@ -0,0 +1,294 @@
...
+     */
+    public static Builder builder(InputStream input) {
+        return new Builder(() -> input);
+    }
+
+    /**
+     * Creates a {@link Builder} for the specified {@link WasmModule}
+     *
+     * @param wasmModule the already parsed Wasm module
+     * @return a {@link Builder} for reading the module definition from the specified input stream
```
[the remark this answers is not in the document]
andreaTP: > It won't make sense when this class is immutable.

completely agree, removed (IIRC it was added to facilitate an "in-between" situation)

wasm/src/main/java/com/dylibso/chicory/wasm/WasmModule.java
andreaTP: I do understand and love this comment, I'm a little worried of applying it straight away(i.e. in this PR) as it will change a lot again ...
Let me go through the rest of the comments and see how far we can get.

andreaTP: I ended up doing those changes, I'll stick with the `builder` since handling `InputStream`s, in memory `byte[]` etc. seems to be compelling use cases, but simplifying where possible and removing redundant APIs.

cli/src/main/java/com/dylibso/chicory/cli/Cli.java
andreaTP: Added an `exportType` method to fill the gap.

wasi/src/test/java/com/dylibso/chicory/wasi/WasiPreview1Test.java
```diff
@@ -20,9 +21,15 @@ public void shouldRunWasiModule() {
         var wasi =
                 new WasiPreview1(this.logger, WasiOptions.builder().withStdout(fakeStdout).build());
         var imports = new HostImports(wasi.toHostFunctions());
-        var module =
-                Module.builder("compiled/hello-wasi.wat.wasm").withHostImports(imports).build();
-        module.instantiate();
+        Instance.builder(
```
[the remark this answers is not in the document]
andreaTP: Fixed all around, please note that the only exception is, currently, in the `aot/README.md` file where I used: `ClassLoader.getSystemClassLoader()`

wasm/src/main/java/com/dylibso/chicory/wasm/WasmModuleType.java
andreaTP: I think the current iteration is "good enough" as a single PR.
We can iterate over this on top.

andreaTP:
This should be ready for the next round of review.

With the last iteration, I removed the usage of `logger` everywhere but in the `wasi` module, I think is great (and we know that it improves the compatibility with Android too), but it's probably "less future-proof".

Since this is a big and sprawling change, I aim to have, at least, @electrum AND @bhelx approvals before merging.

wasi/src/test/java/com/dylibso/chicory/wasi/WasiPreview1Test.java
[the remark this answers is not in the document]
andreaTP: edit - `InterruptionTest`s were failing and switched to `ClassLoader.getSystemClassLoader()` also there.

wasm/src/main/java/com/dylibso/chicory/wasm/WasmModule.java
electrum: We can handle all of these with overloaded `load()` methods. I wasn't suggesting to only support files. But it doesn't require a builder. I can follow up on this if you don't want to change it here.

electrum (approved):

runtime/src/test/java/com/dylibso/chicory/runtime/ModuleTest.java
```diff
@@ -185,35 +184,15 @@ public void shouldCountVowels() {
...
     public void shouldRunBasicCProgram() {
         // check with: wasmtime src/test/resources/wasm/basic.c.wasm --invoke run
-        var instance = Module.builder("compiled/basic.c.wasm").build().instantiate();
+        var instance = Instance.builder(loadModule("compiled/basic.c.wasm")).build();
         var run = instance.export("run");
         var result = run.apply()[0];
         assertEquals(42, result.asInt());
     }
 
-    // @Test
```
electrum, line 194: Why commented out?

andreaTP: Thanks for the reminder!!!

electrum:
This looks great. I'm happy to see this merged now. We can continue to iterate on the APIs.

runtime/src/test/java/com/dylibso/chicory/runtime/ModuleTest.java
andreaTP: Uncommented all the commented tests in this file, and fixed everything around to have them running.

wasm/src/main/java/com/dylibso/chicory/wasm/WasmModule.java
andreaTP: I started doing those changes, but I would like to have the discussion about the support for text format etc.
I think it is better to do it in its own PR.

I have some mixed feelings here:
- having a single `load` method is extremely convenient
- a single method makes the code less "future-proof"
- adding overloads to the "load" method becomes maintenance burden

The reason I'm heading to this direction is that, for example, Component Model changes the Magic Number:
https://github.com/WebAssembly/component-model/blob/main/design/mvp/Binary.md#component-definitions

I don't know if/when/how we would support CM(and/or other features), but it seems appealing to leave the possibility to add properties to a `Builder` even at the `Parser` level, I think it will become handy when implementing extensions to the spec.

runtime/src/main/java/com/dylibso/chicory/runtime/Instance.java
```diff
@@ -433,4 +455,512 @@ public void onExecution(Instruction instruction, long[] operands, MStack stack)
...
+        }
+
+        private Map<String, Export> genExports(ExportSection export) {
+            var exports = new HashMap<String, Export>();
+            int cnt = export.exportCount();
+            for (int i = 0; i < cnt; i++) {
+                Export e = export.getExport(i);
+                if (exports.containsKey(e.name())) {
+                    throw new InvalidException("duplicate export name " + e.name());
+                }
```
bhelx, line 776: Shouldn't these validations be happening at the Module building level? Once the module is built they cannot change right? So won't we be re-validating this stuff everytime we make an instance from the module?

runtime/src/main/java/com/dylibso/chicory/runtime/Instance.java
```diff
@@ -433,4 +455,512 @@ public void onExecution(Instruction instruction, long[] operands, MStack stack)
...
+        }
+
+        public Instance build() {
+            validateModule(module);
+            Map<String, Export> exports = genExports(module.exportSection());
+            var globalInitializers = module.globalSection().globals();
+
+            var dataSegments = module.dataSection().dataSegments();
+
+            // TODO i guess we should explode if this is the case, is this possible?
```
bhelx, line 789: I'd recommend any time we have one of these situations to just assert it. Catching it is always worth the assertion cost.

bhelx (approved):
Just a few questions, but overall I much prefer this interface. Excellent work!

runtime/src/main/java/com/dylibso/chicory/runtime/Instance.java
andreaTP: At the moment, (most of) the validation is happening at the `Instance` building level.
We can evaluate to move some of this logic down to the `Module` building level, but let's do it in a separate PR.

runtime/src/main/java/com/dylibso/chicory/runtime/Instance.java
andreaTP: This comment is floating around since ever, I think it doesn't apply anymore, I remove it.

runtime/src/main/java/com/dylibso/chicory/runtime/Instance.java
```diff
@@ -433,4 +455,512 @@ public void onExecution(Instruction instruction, long[] operands, MStack stack)
...
+
+    public static Builder builder(Module module) {
+        return new Builder(module);
+    }
+
+    public static final class Builder {
+        private final Module module;
+
+        private boolean initialize = true;
+        private boolean start = true;
```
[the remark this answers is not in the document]
andreaTP: All you said is correct, this flag means:
if the module has a start function and this flag is set to "false" it will not be run on instantiation.
I think that we found it useful in some situations to be able to tweak this aspect.

andreaTP:
Thanks everyone for the reviews!
Merging now and we can iterate on top :slightly_smiling_face:

Issue #422, opened 2024-07-02
Review Machine runtime checks


andreaTP:
When we have all the assertions included, we should review the Machine to remove unnecessary runtime checks:
https://github.com/dylibso/chicory/pull/421#discussion_r1661270231

PR #447, opened 2024-07-29
Exploration - long array primitive MStack

andreaTP:
Motivated by #422 and given that we have the full type validation implemented, I wanted to see if there is any gap into using a primitive `long[]` as the backing storage for `MStack`.

The result is promising in terms of "used types" but doesn't work properly yet.
I have a few questions before I invest more time in this:
- is the SIMD(or some other extension) implementation going to invalidate this effort?
- this implementation OOM running `wat2json`, I might have missed something or, maybe, we need, as well, WASI functions to not box in order to avoid performance impact, wdyt?

I welcome comments in any direction cc. @electrum @bhelx @enebo


andreaTP:
Superseded by #531

PR #448, opened 2024-07-29
[WIP] Attempt to revamp the closure interpreter

andreaTP:
Now that we have a mean to measure with JMH I rebased @enebo 's effort to latest `main`:
https://github.com/enebo/chicory/tree/closure_intpreter

I haven't noticed a massive performance improvement, thought, will check on the CI results again.

### DIFF
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/InterpreterMachine.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/InterpreterMachine.java
@@ -94,16 +94,25 @@ class InterpreterMachine implements Machine {
         return results;
     }
 
+    interface ExecuteInstruction {
+        void execute(
+                StackFrame frame,
+                MStack stack,
+                Instance instance,
+                ArrayDeque<StackFrame> callStack,
+                Instruction instruction,
+                long[] operands)
+                throws ChicoryException;
+    }
+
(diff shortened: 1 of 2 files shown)


electrum:
My initial thought is that this is more difficult to read, debug, etc. Given we have AOT, keeping the interpreter simpler seems better, even if it's slightly slower. It depends on the goals for the project.

andreaTP:
I don't want to push for this, I agree that this is slightly more difficult to read/debug etc., the CI also doesn't show any game-changing performance improvement ( [ref](https://github.com/dylibso/chicory/actions/runs/10149706488/attempts/1#summary-28065550857) ).

It was worth giving it a try to make sure we have not been missing an opportunity.

electrum:
If I'm reading the benchmark correctly, it's actually worse by 23% and 36%?

Doing these experiments is helpful. I think it shows that the JVM is good at optimizing and that simpler code is often easier for the JVM to understand.

From a code perspective, you could structure this similar to the existing switch statement by keeping the static methods:
```java
instructions[OpCode.I32_GE_U.ordinal()] =
        (frame, stack, instance, callStack, instruction, operands) -> I32_GE_U(stack);
```
With Java 21, this could be simplified to
```java
instructions[OpCode.I32_GE_U.ordinal()] =
        (_, stack, _, _, _, _) -> I32_GE_U(stack);
```
But I like the switch better (and seems easier for the JVM to understand).

PR #450, opened 2024-07-30
Demo usage of C# with DotNet8

andreaTP:
C# it's a single Wasi function away :slightly_smiling_face: 

(the size of the resulting binary is huge ...)


wasi/src/main/java/com/dylibso/chicory/wasi/WasiPreview1.java
```diff
@@ -170,9 +170,26 @@ public HostFunction clockResGet() {
...
-                    throw new WASMRuntimeException(
-                            "We don't yet support this WASI call: clock_res_get");
-                    // return new Value[] { Value.i32(0) };
+                    int clockId = args[0].asInt();
+                    int resultPtr = args[1].asInt();
+
+                    Memory memory = instance.memory();
+                    switch (clockId) {
+                        case WasiClockId.REALTIME:
+                            memory.writeLong(resultPtr, 1L);
```
andreaTP, line 179: I have done some reading, but I cannot come up with anything better than "just use 1" ...
I'm open to suggestions!

andreaTP:
> (the size of the resulting binary is huge ...)

added a bunch of trimming options and now looks better.

wasi/src/main/java/com/dylibso/chicory/wasi/WasiPreview1.java
```diff
@@ -170,9 +170,26 @@ public HostFunction clockResGet() {
...
+                            return wasiResult(WasiErrno.ESUCCESS);
+                        case WasiClockId.MONOTONIC:
```
electrum, line 181: These two cases can be combined
```java
case WasiClockId.REALTIME:
case WasiClockId.MONOTONIC:

wasi/src/main/java/com/dylibso/chicory/wasi/WasiPreview1.java
electrum: This seems fine to me.

electrum (approved):
Cool!

wasi/src/main/java/com/dylibso/chicory/wasi/WasiPreview1.java
andreaTP: fixed

PR #452, opened 2024-07-30
[do not merge] test python support

andreaTP:
Most of the projects to run Python on WASI are in a bad state, so, I think that we should not test in CI against outdated toolchains.

That said, this tutorial:
https://wasmlabs.dev/articles/python-wasm32-wasi/

kind of works for dummy examples.


electrum:
Note that Python 3.13 has [official support for WASI](https://github.com/python/cpython/blob/main/Tools/wasm/README.md), so that seems reasonable to target, although you have to compile from source.

wasi/src/test/java/com/dylibso/chicory/wasi/WasiPreview1Test.java
```diff
@@ -77,4 +88,46 @@ public void shouldRunTinyGoModule() {
...
+        Path inputFolder = fs.getPath("/");
+
+        // cannot get inline python to run
+        // but works when loading a file ...
+        // -- -c "import sys; from pprint import pprint as pp; \
+        //         pp(sys.path); pp(sys.platform)"
+        var script = "print(\"hello python!\")";
+
+        FileWriter fileWriter = new FileWriter(new File("/tmp/try-python-wasm").toPath().resolve("test.py").toFile());
+        PrintWriter printWriter = new PrintWriter(fileWriter);
```
electrum, line 110: Nit: you can use `Files.writeString()`

wasi/src/test/java/com/dylibso/chicory/wasi/WasiPreview1Test.java
```diff
@@ -77,4 +88,46 @@ public void shouldRunTinyGoModule() {
...
         assertEquals(result.asInt(), 42);
     }
+
+    @Test
+    public void shouldRunPythonModule() throws Exception {
+        // implementation of this tutorial:
+        var fakeStdout = new MockPrintStream();
+        var fakeStderr = new MockPrintStream();
+        FileSystem fs =
+                Jimfs.newFileSystem(
```
electrum, line 98: Should we add an `<optional>true</optional>` dependency on Jimfs and add a utility method for this?

wasi/src/test/java/com/dylibso/chicory/wasi/WasiPreview1Test.java
andreaTP: removed

electrum:
It's possible to use wizer to improve startup time after you've run all the imports.

andreaTP:
This is not going to be merged anyhow, it served as an exercise but let's close it.

PR #454, opened 2024-07-30
Skip macos Java 11 test because of memory leak

### LINKED ISSUE #451: [CI] Macos runner are exceeding the java heap usage
https://github.com/dylibso/chicory/actions/runs/10161005690/job/28098559133?pr=449#step:7:1364

> Error: [ERROR] There was an error in the forked process
Error:  Java heap space
Error:  org.apache.maven.surefire.booter.SurefireBooterForkException: There was an error in the forked process
Error:  Java heap space

bhelx:
Skip macos11 until we can fix memory leak in #451

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -21,6 +21,9 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
         version: [11, 17]
+        exclude:
+          - os: macos-latest
+            version: 11
     steps:
       - name: Checkout sources
         uses: actions/checkout@v4


andreaTP:
I tested locally increasing the Maven memory but doesn't seems to have much impact.
I suspect a leak of resources due to `Wast2Json` or something similar ...

bhelx:
yeah. okay, maybe we just skip this one for now.

andreaTP (changes requested):
> maybe we just skip this one for now.

Drastic! :slightly_smiling_face: 
I would keep the test, if the level of attrition goes above the limit we can investigate, but, for now, the failure happens only seldom.

PR #459, opened 2024-08-12
Review exception handling

### LINKED ISSUE #455: `WasiExitException` should not be wrapped by `WASMMachineException`
The problem is observable here:
https://github.com/dylibso/chicory/blob/ee226c30dd81a673e8bd2d82e2780064e507a5fc/wasi/src/test/java/com/dylibso/chicory/wasi/WasiPreview1Test.java#L91-L92

andreaTP:
This is a more transparent approach to error handling that doesn't wrap underlying exceptions.

Fix #455


runtime/src/main/java/com/dylibso/chicory/runtime/Instance.java
```diff
@@ -297,27 +300,13 @@ public ExportFunction export(String name) {
...
-                            return machine.call(funcId, args);
-                        } catch (InvalidException e) {
-                            throw e;
-                        } catch (TrapException e) {
-                            throw new UninstantiableException("unreachable", e);
-                        } catch (Exception e) {
-                            throw new WASMMachineException(machine.getStackTrace(), e);
-                        }
-                    };
+                    return (args) -> machine.call(export.index(), args);
```
electrum, line 303: Nit: no need for parens around singular lambda argument

runtime/src/main/java/com/dylibso/chicory/runtime/InterpreterMachine.java
```diff
@@ -59,7 +59,11 @@ public static Value[] call(
             stackFrame.pushCtrl(OpCode.CALL, 0, type.returns().size(), stack.size());
             callStack.push(stackFrame);
 
-            eval(stack, instance, callStack);
+            try {
```
electrum, line 62: Nice, doing this here is cleaner

electrum (approved):
I think we can delete `WASMMachineException` after this

evacchi (approved):

andreaTP:
Removed `WASMMachineException`, good catch!

PR #461, opened 2024-08-12
Disable the type validation in wabt tools

andreaTP:
This potentially fixes/improves the situation for #451 but opens a question wrt our API.

Current situation:

You can do the parsing and cache the loaded module in memory with:
```java
private static final Module MODULE = Module.builder(...).build();
```
after that, you still need 2 steps:
 - `Instance` initialization
 - `start` function call
 the initialization will also map the `HostImports`, but, when using `wasi` we always need to re-initialize everything even when changing only the `arguments`, as we have a single `build` method:
 ```java
 Instance.builder(MODULE).withHostImports(imports).build();
 ```
 
 this is sub-optimal as there are use cases(i.e. multiple subsequent executions of a tool) that would benefit from caching the `Instance` and hot-swapping the wasi `HostImports`:
 
 Possible resolution:
 
 ```java
 private static final HostImports baseImports = ???;
 private static final Instance INSTANCE = Instance.builder(Module.builder(...).build()).withHostImports(baseImports).withStart(false).build();
 ```
 
 when using the `Instance` we should be able to fully skip the initialization(and have an effective memoization) by using, something like:
 
 ```java
 INSTANCE.updateImports(baseImports.withArgs(myNewArgs)).export("_start").apply();
 ```
 
 There are few possible alternatives to tackle the problem, I'm interested if this is something we want to fix now or something that can be postponed, I'm not sure how much this pattern is widely used. 
 Thoughts?

### DIFF
--- a/wabt/src/main/java/com/dylibso/chicory/wabt/Wast2Json.java
+++ b/wabt/src/main/java/com/dylibso/chicory/wabt/Wast2Json.java
@@ -90,7 +90,10 @@ public class Wast2Json {
                                 .withOpts(wasiOpts.build())
                                 .build()) {
                     HostImports imports = new HostImports(wasi.toHostFunctions());
-                    Instance.builder(MODULE).withHostImports(imports).build();
+                    Instance.builder(MODULE)
+                            .withTypeValidation(false)
+                            .withHostImports(imports)
+                            .build();
                 }
 
                 createDirectories(output.toPath().getParent());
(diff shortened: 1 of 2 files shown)


andreaTP:
Execution time of the tests in CI:
- before this change: 23 sec
- after this change: 16 sec

andreaTP:
A couple of additional info I forgot:
- the host mapping happens against the functions type signatures: should we go for an `HostImportSignature`?
- the need for more "dynamic" `HostFunctions` have been evident also here: https://github.com/andreaTP/opa-chicory/blob/e3d86b1c621ea58dbb9d6de359eaf1c45e27c581/core/src/main/java/com/github/andreaTP/opa/chicory/OpaWasm.java#L49 it would be great to separate the definition of the import from their actual implementation

electrum (approved):

andreaTP:
I merge this to stabilize the CI, but would love to continue the conversation regarding the correct API for caching `Instance`s

electrum:
It seems like the validation should be done when the module is loaded -- it doesn't need to be done for each instance.

PR #465, opened 2024-08-14
Make type validation use Module instead of Instance

electrum:
After #460 is merged, we can make this part of `Module.Builder`.


electrum:
@andreaTP Thanks, rebased and updated

andreaTP (approved):
LGTM! Thanks!

PR #470, opened 2024-08-18
Cleanup dependencies and enforce usage

electrum:
Ensure that
* we declare all dependencies that are used
* we do not declare unused dependencies (only declare dependencies that are used)


bhelx (approved):

andreaTP (approved):
lgtm, thanks

PR #488, opened 2024-08-27
GlobalInstance should expose MutabilityType instead of HostGlobal

evacchi:
Move the MutabilityType field from HostGlobal to GlobalInstance. Notice that GlobalInstance has both getters and setters, we might want to revisit that.

Adjust tests accordingly.

### DIFF
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/GlobalInstance.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/GlobalInstance.java
@@ -1,13 +1,21 @@
 package com.dylibso.chicory.runtime;
 
+import com.dylibso.chicory.wasm.types.MutabilityType;
 import com.dylibso.chicory.wasm.types.Value;
 
 public class GlobalInstance {
     private Value value;
     private Instance instance;
+    private MutabilityType mutabilityType;
+
+    public GlobalInstance(Value value, MutabilityType mutabilityType) {
+        this.value = value;
+        this.mutabilityType = mutabilityType;
+    }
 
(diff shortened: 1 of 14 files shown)


runtime/src/main/java/com/dylibso/chicory/runtime/GlobalInstance.java
```diff
@@ -1,13 +1,21 @@
 package com.dylibso.chicory.runtime;
 
+import com.dylibso.chicory.wasm.types.MutabilityType;
 import com.dylibso.chicory.wasm.types.Value;
 
 public class GlobalInstance {
     private Value value;
     private Instance instance;
+    private MutabilityType mutabilityType;
```
andreaTP, line 9: make it final and remove the setter?

andreaTP (changes requested):
We are going in a direction on making, as much as possible, classes immutable.
From the diff `setMutabilityType` is not used, I would remove it and make the field `final`, other than that, LGTM, thanks for the improvement!

runtime/src/main/java/com/dylibso/chicory/runtime/GlobalInstance.java
evacchi: Reason I didn't do it is for consistency with the rest of the class, I would have suggested to do it all at once later, turning the entire class immutable. But I can do it for this field if you prefer

andreaTP: thanks

andreaTP (approved):
LGTM, thanks for the cleanup!

PR #493, opened 2024-08-28
Introduce a dev profile to disable strict checks

andreaTP:


README.md
```diff
@@ -389,6 +389,7 @@ Basic steps:
 
 * `mvn clean install` to run all of the project's tests and install the library in your local repo
 * `mvn -Dquickly` to install the library skipping all tests
+* `mvn -Pdev <...goals>` to disable linters and enforcers during development
```
evacchi, line 392: I haven't tried this yet, but since you added an activation property, I think you can write, for consistency with `-Dquickly`: 

```suggestion
* `mvn -Ddev <...goals>` to disable linters and enforcers during development
```

andreaTP: it's the same, no particular preference from my side.

pom.xml
```diff
@@ -616,6 +618,22 @@
...
+      </activation>
+      <properties>
+        <spotless.check.skip>true</spotless.check.skip>
+        <checkstyle.skip>true</checkstyle.skip>
+        <enforcer.skip>true</enforcer.skip>
+        <maven.compiler.failOnWarning>false</maven.compiler.failOnWarning>
+        <maven.dependency.failOnWarning>false</maven.dependency.failOnWarning>
+      </properties>
+    </profile>
     <profile>
```
electrum, line 637: Nit: blank lines between profiles makes them easier to read

electrum (approved):
Good idea! We do something similar in other projects.

PR #495, opened 2024-08-28
Introduce the Store + fixes to validation

### LINKED ISSUE #486: Introduce the Store
Related to #482.

Before introducing HostModules and eventually Linking it probably makes sense to introduce the concept of a Store. See e.g. [Wasmtime][wasmtime], [wazero][wazero] and [the Spec][spec]

A Store is an intermediate-level abstraction that collects Wasm (function/global/memory etc) instances and host-defined state. It simplifies creating instances when there are a lot of interdependencies, but _it does not_ resolve names like a [Linker (cf. wasmtime)][wasmtime-linker] would do. In other words, the end user is still required to instantiate modules in the right order, but at least this way it will be the Store to keep around all the right globals/functions etc.

E.g. imagine you have modules x, y, z, with z depending on x and y (i.e. importing from names `x` and `y`); you would write something like (for instance)

```java
// if there are any host functions independent from Wasm modules
store.withHostFunction(...); // these could be eventually stored in a HostModule 

// Instantiate and register exports automatically into the store.
store.instantiate("x", x); 
store.instantiate("y", y); 

// We can now instantiate z
var zInstance = store.instantiate("z", z); 
// all imports of the type <"x", "foo"> or <"y", "bar"> are found in the store 
// if they are exported by x, y; so instantiation succeeds.

// zInstance is ready to use.
zInstance.call(...)
```

The API is a strawman, the following would work just as well:

```java
// if there are any host functions independent from Wasm modules
store.withHostFunction(...); // these could be eventutally stored in a Host Module if/when they are introduced.

// Instantiate and register exports automatically into the store.
Instance.builder(x).withStore(store, "x").build(); 
Instance.builder(y).withStore(store, "y").build(); 

// We can now instantiate z
var zInstance = Instance.builder(z).withStore(store).build();

// zInstance is ready to use.
zInstance.call(...)
```

[wasmtime]: https://docs.wasmtime.dev/api/wasmtime/struct.Store.html
[wasmtime-linker]: https://docs.wasmtime.dev/api/wasmtime/struct.Linker.html
[wazero]: https://github.com/tetratelabs/wazero/blob/2b12e189eeec0e4f2968e974cf25419bcb071b34/internal/wasm/store.go#L24-L36
[spec]: https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/#store%E2%91%A0

evacchi:
Closes #486 

Introduces the Store abstraction and adopts it in the test cases, cleaning up a lot of unnecessary test cases. ~~I am opening the PR in draft, as it needs rebasing anyway.~~

The change is relatively simple, it introduces the `Store` as described in #486, with a fluent interface of the type:

```java
var store = new Store()
   .addFunctions(f1, ...)
   .addGlobals(g1, ...)
```

This API is copied over from `HostImports`, although this is not a subclass; it is still "related" to it; in fact:
 
- it is possible to get the contents of a store ("converting" it, if you will) to a `HostImports` instance with `store.toHostImports()`; 
- the `Store` also accepts an entire `HostImports` instance as a shorthand for registering all of the items in it with `store.addHostImports(hostImports)`

It also provides shorthands to register all of the exports in a module as qualified "HostGlobal"s, "HostFunction"s etc. 

```java
store.register("name", instance); // now each export `e` in instance, is stored as the qualified name `name.e`
```

It is also possible to instantiate and register in one move:

```java
var instance = store.instantiate("name", module); 
```

this is equivalent to:

```java
var instance = Instance.builder(module).withHostImports(store.toHostImports()).build()
store.register("name", instance)
```

The Store is deterministic, in that its behavior is known: when clashing instances (i.e. where their qualified name `n.e` is clashing) are registered to the store, the latest registration overwrites the oldest. Internally each type(*) (globals, funcs, etc) is kept in separate LinkedHashMaps.

Introducing the `Store` enormously simplifies the definition of the spec tests, because such tests actually expect changes to "stick" across cases; i.e. they mutate sequentially a shared `store` instance adding and updating memories, functions, globals, tables.

In fact, introducing the store highlighted a bunch of issues with validation, that I am fixing contextually: even though we would not let any invalid instance slip, we would raise `UnlinkableException`s during the initialization of the instance, instead of throwing an `InvalidException` at validation time.

Potentially this allows us to avoid the second check at the time of instantiation, because we could assume that validation has passed.

(*) note: this actually means that clashes might exists if two different types have the same qualified name 🤔 
e.g. `n.e` is a func, but there is also a `n.e` that is a global 🤔


runtime/src/main/java/com/dylibso/chicory/runtime/ConstantEvaluators.java
evacchi, line 1: must revert this

runtime/src/main/java/com/dylibso/chicory/runtime/GlobalInstance.java
```diff
@@ -1,13 +1,21 @@
 package com.dylibso.chicory.runtime;
 
+import com.dylibso.chicory.wasm.types.MutabilityType;
 import com.dylibso.chicory.wasm.types.Value;
 
 public class GlobalInstance {
     private Value value;
     private Instance instance;
+    private MutabilityType mutabilityType;
```
evacchi, line 9: this is from an early rebase, should disappear in the diff once rebased on main

README.md
```diff
@@ -291,6 +291,57 @@ writeResultFile("hostFunction.result", hostFunctionResult);
...
+var store = new Store();
+// registers `console.log` defined above in the store
+store.addHostFunction(func);
+```
+
+However, the store also automatically exposes the exports of a module to the other instances that are registered.
+
+```java
+// registers `logger` into the store
+store.register("logger", instance);
```
andreaTP, line 316: where `instance` is coming from?

README.md
```diff
@@ -291,6 +291,57 @@ writeResultFile("hostFunction.result", hostFunctionResult);
...
+There is also a shorthand method to instantiate a module and register the resulting instance:
+
+```java
+var logger2 = store.instantiate("logger2", Parser.parse(new File("./logger.wasm")));
+```
+
+This is equivalent to:
+
+```java
+var hostImports = store.toHostImports();
```
andreaTP, line 329: I'm surprised that we are keeping compatibility with the "legacy" `HostImports`.
I think that using the `Store` covers already all the use cases, so, we can drop the previous lower level API.
wdyt?

aot/src/test/java/com/dylibso/chicory/testing/TestModule.java
```diff
@@ -63,17 +60,8 @@ public TestModule(Module module) {
...
-                        .build();
-        return this.instance;
-    }
-
-    public TestModule withHostImports(HostImports imports) {
-        this.imports = imports;
-        return this;
+    public Instance instantiate(Store s) {
+        HostImports hostImports = s.toHostImports();
+        return Instance.builder(module).withHostImports(hostImports).build();
```
andreaTP, line 65:
```suggestion
        return Instance.builder(module).withHostImports(hostImports).withMachineFactory(AotMachine::new).build();
```

:sweat_smile:

runtime/src/main/java/com/dylibso/chicory/runtime/Store.java
```diff
@@ -0,0 +1,166 @@
...
+        register(name, instance);
+        return instance;
+    }
+
+    /**
+     * QualifiedName is internally used to use pairs (moduleName, name) as keys in the store.
+     */
+    static class QualifiedName {
+        private final String moduleName;
+        private final String name;
```
andreaTP, line 141: nitpick: I would call it `fieldName` or something, there are already enough "name"s around here :sweat_smile:

wasm/src/main/java/com/dylibso/chicory/wasm/Module.java
```diff
@@ -256,6 +256,9 @@ public Module build() {
             validator.validateModule();
             if (validate) {
                 validator.validateFunctions();
+                validator.validateGlobals();
```
andreaTP, line 259: nice :+1:

andreaTP (approved):
A couple of nitpicks, but this is definitely going in the right direction.
If we want to merge this PR in the current state, I think that we should follow up removing the raw `HostImports` API at all from the user perspective.

The only blocking comment is to restore `withMachineFactory`: https://github.com/dylibso/chicory/pull/495/files#diff-5c1b4237718f90ef1f6d1d70f6f81c6a6883e075eb40a0bbbb25ea05fa4110a8R65

andreaTP (changes requested):
sorry for the noise, needs this: https://github.com/dylibso/chicory/pull/495/files#diff-5c1b4237718f90ef1f6d1d70f6f81c6a6883e075eb40a0bbbb25ea05fa4110a8R65

aot/src/test/java/com/dylibso/chicory/testing/TestModule.java
evacchi: ouch

README.md
evacchi: `HostImports` is just a bare container, and it is immutable. I felt it made sense to keep the distinction for now, we can always refine later :)

runtime/src/main/java/com/dylibso/chicory/runtime/Store.java
evacchi: I'll use `fieldName` for consistency, but I feel we should revisit naming names ;)

README.md
evacchi: from above 😬

evacchi: do you want me to show the instantiation in the same code snippet, or is it ok to clarify it's the `instance` defined above? notice the same issue is with `console.log` (`func`) a few lines earlier.

runtime/src/main/java/com/dylibso/chicory/runtime/Store.java
andreaTP: be my guest :pray:

README.md
andreaTP: good question, no strong opinion, either of your proposals work for me.

README.md
andreaTP: ok, let's follow up on this.

README.md
evacchi: clarified that `func` and `instance` are defined in the previous section of the README

andreaTP: thanks :+1:

andreaTP (approved):
LGTM, thanks!

andreaTP (approved):
I'm merging this one to avoid accumulating conflicts, @electrum I would still value your review, especially wrt public API and naming.

evacchi:
+1 @electrum :)

Issue #482, opened 2024-08-21
Introduce Host Modules


evacchi:
Every Wasm module is able to require some items, such as functions, at instantiation phase.
These are called _imports_, and every import is labeled by ["a two-level name space, consisting of a 
module name and a name for the entity"][syntax-import-spec].

The spec also shallowly defines the concept of [Host Function][exec-host-function-spec] as of

> a function expressed outside WebAssembly but passed to a module as an import.

Because imports are alway qualified and because, generally, related functions 
belong to the same module, they are usually qualified by the same name,
and they might even share some form of state. 
[wazero][wazero] has introduced the concept of a Host _Module_.

For instance, suppose that a module `need_add.wasm` imports a function `env.add`:

```wat
(module (import "env" "add" (func)))
```

Instead of declaring the function `<env, add>`, users define a _host module_
`env` and then declare all of the functions that belong to that module, including `add`.

wazero's Host Module is nothing more than a bundle of related functions that, for convenience, 
are defined together. On a surface level, this is a convenience for end users, who will
declare modules in a more concise way; but in reality, bundling related functions
into a host module simplifies management within the engine, and ultimately improves
lifecycle control.

This is a proposal to introduce a similar construct in Chicory.

## Benefits for End Users

End users will be able to define a collection of related host functions more concisely.

For instance, the signature for the HostFunction constructor is currently:

```java
    public HostFunction(
            WasmFunctionHandle handle,
            String moduleName,
            String fieldName,
            List<ValueType> paramTypes,
            List<ValueType> returnTypes) {
        this.handle = handle;
        this.moduleName = moduleName;
        this.fieldName = fieldName;
        this.paramTypes = paramTypes;
        this.returnTypes = returnTypes;
    }
```

We can imagine to provide a `HostModule.Builder` [in the same fashion as wazero's][host-module-builder]:

```go
	_, err := r.NewHostModuleBuilder("env").
		NewFunctionBuilder().
		WithFunc(func(v uint32) {
			fmt.Println("log_i32 >>", v)
		}).
		Export("log_i32").
		NewFunctionBuilder().
		WithFunc(func() uint32 {
			if envYear, err := strconv.ParseUint(os.Getenv("CURRENT_YEAR"), 10, 64); err == nil {
				return uint32(envYear) // Allow env-override to prevent annual test maintenance!
			}
			return uint32(time.Now().Year())
		}).
		Export("current_year").
		Instantiate(ctx)
```

[...]

PR #484, opened 2024-08-22
PoC: Introduce host modules

evacchi:
**This is a PoC meant for discussion, it is not necessarily meant for merging.**

Addresses #482 and potentially #464.

This PR refactors Wasip1 as a means to show an example and it introduces ~~(temporarily in the `wasi` package)~~ a few types:

- `HostModule` and `HostModule.Builder`. The `HostModule` interface represents a bundle of `FunctionSignature`s with a name (the name of the module). The `HostModule.Builder` returns a concrete implementation called `FunctionSignatureBundle`.
- `HostModuleInstance` is an interface representing an instance of a `HostModule`; it owns a reference to its host module "prototype" and it collects the `HostFunction`s that belong to that instance, matching with the signatures given in the `HostModule`. The `HostModuleInstance` can be also `close()`d (and potentially other lifecycle actions could be added, e.g. `instantiate()`). The `HostModuleInstance.Builder` constructs a concrete implementation called `HostFunctionBundle`.

- `FunctionSignature` is a representation of a (Host) Function's signature. This is essentially a different representation for a `FunctionImport` so the two might collapse somehow, or at least share an ancestor.
- `FunctionSignatureBundle` implements `HostModule` as explained above.
- `HostFunctionBundle` implements `HostModuleInstance` as explained above.

A `HostModule` can be constructed using the builder as such:

```java
        var wasip1 = HostModule.builder(MODULE_NAME)
                .withFunctionSignature("adapter_close_badfd", i32_t, i32_t)
                .withFunctionSignature("adapter_open_badfd", i32_t, i32_t)
                .withFunctionSignature("args_get", i32_i32_t, i32_t)
                // ...
               .build()
```

Notice how the `HostModule` does not define any implementation for the functions. The implementations are given at instantiation time. In order to create a `HostModuleInstance` we can use the `HostModuleInstance.Builder` as such:

```java
        var inst = new WasiPreview1(...);
        var wasip1Instance HostModuleInstance.builder(hostModule)
                .bind("adapter_close_badfd", inst::adapterCloseBadfd)
                .bind("adapter_open_badfd", inst::adaptedOpenBadfd)
                .bind("args_get", inst::argsGet)
                // ...
                .build();
```

Notice:

1. we need to instantiate an object instance of `WasiPreview1` so that such instance is bound to the lifecycle of the `HostModuleInstance`.
2. we can also optionally specify a close handler, that will be called to cleanup resources if any:
```java
        var inst = new WasiPreview1(...);
[...]

PR #496, opened 2024-08-29
Add experimental annotation framework for functions

electrum:
This implements a basic annotation framework, similar to the example shown in #482. It works and could be merged now, with the expectation that we'll try it out with more projects such as SQLite, and eventually extract it as a public interface in its own module.

Ideas for future improvements:

* More argument type conversions. For example, accepting `byte[]` or `String` from `(ptr i32, len i32)` arguments. This gets complicated quickly, but some basic support might cover most use cases.
* Automatic tracing of calls and results, for example, similar to `strace` on Linux. This really needs argument type conversion to be useful, so that the traces can show strings rather than useless pointers.
* Support multiple return values.

I suspect that @andreaTP will dislike the usage of `MethodHandle`. I'm thinking we can convert this to be an annotation processor that generates the `HostFunction[]` adapter, though I'm not sure what the API would look like.

The current implementation doesn't change the API of `WasiPreview1`: users create it and call `toHostFunctions()`. An annotation processor would generate a new class, and I don't think it's possible for `WasiPreview1` to statically reference that class (as it won't exist when `WasiPreview1` is compiled). I see a couple of options:

* `WasiPreview1#toHostFunctions` would call some utility method like `getHostFunctions(WasiPreview1.class)` which would reflectively find the generated class.
* Change callers to use directly reference the generated class:
  ```java
  WasiPreview1 instance = new WasiPreview1(logger, opts);
  HostFunctions[] functions = WasiPreview1.HostFunctions.get(instance);
  ```

Neither one seems ideal, so I'm hoping someone has better suggestions, if we want to go down the annotation processor route. I'm personally fine with the `MethodHandle` approach, as this is quite common in these sorts of frameworks, but there are advantages to generating code at compile time.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -38,6 +38,8 @@
     <module>aot-tests</module>
     <module>bom</module>
     <module>cli</module>
+    <module>function-annotations</module>
+    <module>function-processor</module>
     <module>fuzz</module>
     <module>jmh</module>
     <module>log</module>
@@ -143,6 +145,16 @@
       </dependency>
       <dependency>
         <groupId>com.dylibso.chicory</groupId>
+        <artifactId>function-annotations</artifactId>
+        <version>${project.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>com.dylibso.chicory</groupId>
+        <artifactId>function-processor</artifactId>
+        <version>${project.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>com.dylibso.chicory</groupId>
         <artifactId>log</artifactId>
         <version>${project.version}</version>
       </dependency>
@@ -177,6 +189,13 @@
         <groupId>com.google.errorprone</groupId>
         <artifactId>error_prone_annotations</artifactId>
         <version>${error-prone.version}</version>
+      </dependency>
+
+      <!-- Use latest Guava -->
+      <dependency>
+        <groupId>com.google.guava</groupId>
+        <artifactId>guava</artifactId>
+        <version>33.3.0-jre</version>
       </dependency>
 
       <!-- Used for WASI tests -->
@@ -224,6 +243,14 @@
         <groupId>com.approvaltests</groupId>
         <artifactId>approvaltests</artifactId>
         <version>${approvaltests.version}</version>
+        <scope>test</scope>
+      </dependency>
+
+      <!-- Used for annotation processor tests -->
+      <dependency>
+        <groupId>com.google.testing.compile</groupId>
+        <artifactId>compile-testing</artifactId>
+        <version>0.21.0</version>
         <scope>test</scope>
       </dependency>
 
--- a/pom.xml
+++ b/pom.xml
@@ -172,6 +172,13 @@
         <version>${project.version}</version>
       </dependency>
 
+      <!-- Error Prone -->
+      <dependency>
+        <groupId>com.google.errorprone</groupId>
+        <artifactId>error_prone_annotations</artifactId>
+        <version>${error-prone.version}</version>
+      </dependency>
+
       <!-- Used for WASI tests -->
       <dependency>
         <groupId>com.google.jimfs</groupId>
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -16,6 +16,11 @@
     <dependencies>
       <dependency>
         <groupId>com.dylibso.chicory</groupId>
+        <artifactId>function-annotations</artifactId>
+        <version>${project.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>com.dylibso.chicory</groupId>
         <artifactId>log</artifactId>
         <version>${project.version}</version>
       </dependency>
--- a/function-annotations/pom.xml
+++ b/function-annotations/pom.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>com.dylibso.chicory</groupId>
+    <artifactId>chicory</artifactId>
+    <version>999-SNAPSHOT</version>
+    <relativePath>../pom.xml</relativePath>
+  </parent>
+  <artifactId>function-annotations</artifactId>
+  <packaging>jar</packaging>
+  <name>Chicory - Function annotations</name>
+  <description>Experimental annotations for host functions</description>
+
+</project>
--- a/function-annotations/src/main/java/com/dylibso/chicory/function/annotations/HostModule.java
+++ b/function-annotations/src/main/java/com/dylibso/chicory/function/annotations/HostModule.java
@@ -1,0 +1,12 @@
+package com.dylibso.chicory.function.annotations;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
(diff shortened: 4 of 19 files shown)


evacchi:
Provided that I have nothing against `MethodHandle` (especially in this kind of use case), it should be fine for a class to reference a class generated by an `AnnotationProcessor`, because the processor should be invoked by `javac` automatically; e.g. consider `AutoValue` https://github.com/google/auto/blob/main/value/userguide/index.md#in-your-value-class

```
@AutoValue
abstract class Animal {
  static Animal create(String name, int numberOfLegs) {
    return new AutoValue_Animal(name, numberOfLegs);
  }

  abstract String name();
  abstract int numberOfLegs();
}
```

`AutoValue_Animal` is the generated class.

Another option is to generate the `HostFunction` "converter"; e.g. check this out for a raw sketch of the idea (here it is defined _within_ `WasiPreview1` but it could be a companion class)

https://github.com/dylibso/chicory/pull/484/files#diff-9037d84891e8043e756bfd63e2101686b67d1cad2cc2c4af3b2e6b48675acdf3R1217-R1338

electrum:
If we want to allow fast calls to host functions, we need to accept that `WasmFunctionHandle` is not the right interface, as it requires boxing everything. The annotated functions in this PR could be called directly by AOT without any conversion. With `WasmFunctionHandle`, when calling from AOT, we will box everything just to fit that interface, then unbox before calling the host function.

If we generate code at compile time that uses `WasmFunctionHandle`, then we're stuck with that (without recompiling the user code) and there's no way to make it fast. Using `MethodHandle` at runtime allows us to choose the correct adaptation for the interpreter vs AOT. Generating code also adds a dependency on the low level APIs, so they have to be stable.

andreaTP:
> Generating code also adds a dependency on the low level APIs, so they have to be stable.

correct

> without recompiling the user code) and there's no way to make it fast

I understand this, but is usually acceptable.

All in all, I now better understand where you are going @electrum , and I do not want to prevent you going down this direction.
Would it be feasible to make this code "optional"?
Or to "somehow" bake it into the `aot` module?

I'm asking as I still value the fact that the interpreter is completely "reflection free" and I'd like to keep Wasi modules runnable without having to use to reflection.
In absence of better options, we can still expose both the `MethodHandle` and `WasmFunctionHandle` APIs from 2 maven sub-modules.

As soon as this request is not a blocker I'm happy to support this effort!

electrum:
My primary goal right now is a better interface for writing host functions (the WASI functions seem much nicer after this refactor), so I'll update this to use an annotation processor, allowing us to move forward and experiment with the annotation interface.

Thanks @evacchi for the pointer to AutoValue. I'm surprised that works, but it's exactly what we need.

We can follow up on the `WasmFunctionHandle` issue. At a minimum, we would need a union between that and `MethodHandle`, but using a method handle directly has the issue of knowing the "calling convention". This works fine in the beginning when there is only one calling convention, but becomes limiting if we want support more later.

andreaTP:
Happy to hear that we have a path forward!

electrum:
@andreaTP I updated this to generate code using an annotation processor. The generated suffix of `_ModuleFactory` seems to match the direction that @evacchi is going with host modules, but happy to rename anything here. The underscore is unconventional for Java classes, but makes it stand out as a generated class (like AutoValue) and won't conflict with user code.

pom.xml
```diff
@@ -172,13 +184,27 @@
         <version>${project.version}</version>
       </dependency>
 
+      <!-- Use latest Guava -->
+      <dependency>
+        <groupId>com.google.guava</groupId>
+        <artifactId>guava</artifactId>
+        <version>33.3.0-jre</version>
```
andreaTP, line 198: This is a transitive dependency of `compile-testing`, to make sure that we don't accidentally get Guava on the classpath I would add:
```xml
<scope>test</scope>
```

there should be a very good reason to change it ...

pom.xml
```diff
@@ -172,13 +184,27 @@
...
         <groupId>com.google.jimfs</groupId>
         <artifactId>jimfs</artifactId>
         <version>1.3.0</version>
       </dependency>
 
+      <!-- Used for annotation processor tests -->
+      <dependency>
+        <groupId>com.google.testing.compile</groupId>
+        <artifactId>compile-testing</artifactId>
+        <version>0.21.0</version>
```
andreaTP, line 212: nit: `<scope>test</scope>`
but great to see `compile-testing` used here :+1:

andreaTP (approved):
A couple of nitpicks on dependencies but this is brilliant!

I think that this PR is a great start and I'm happy to move forward.
A couple of follow up:
- `docs`:
    - instructing people how to use the annotation processor
    - settle expectations toward the supported functionality(e.g. handling of "boxed" primitives like `Integer` and similar notes)
    
- next steps:
    - I'm working on something somehow ["similar"](https://github.com/dylibso/chicory/compare/main...andreaTP:chicory:bindgen) with a slightly different scope
    - I'd need a bit more(e.g. emitting the Wasm Module specific "Machine" during compilation/generation) to finalize [this PR](https://github.com/dylibso/chicory/pull/494)
    - and I'm not yet sure how all of this will fit together :blush:

pom.xml
evacchi: I suspect this might not work

evacchi (approved):
looks great!

electrum:
@andreaTP That's cool! [Takari Maven Plugin Testing Framework](https://github.com/takari/takari-plugin-testing-project) is a great way to test Maven plugins using normal unit tests. We use it for the [Trino Maven Plugin](https://github.com/trinodb/trino-maven-plugin) (though it's still on JUnit 4).

pom.xml
andreaTP: We already do it with other dependencies: https://github.com/dylibso/chicory/blob/264f57dbada145dc43d1c8ebc458a81e2857e43c/pom.xml#L220

electrum: This turns out not to work due to Guava being a dependency of Jimfs, which is used by `Wast2Json` during the invocation of `TestGenMojo`. Thus, from the perspective of `test-gen-plugin`, it is not a test dependency, so the scope shouldn't be changed.

andreaTP:
@electrum do you mind if I ask you to separate [this development](https://github.com/dylibso/chicory/pull/496/commits/dc87a4ef6f90e08f52c18d523d4fc40638c4619e) in another PR?
(not against it but I want to have the discussion about the memory layout of things).

We can merge this PR today and move on building on top.

pom.xml
andreaTP: not a big deal, but I would overwrite the scope in the `test-gen-plugin` to clearly indicate that is an expected "exception".
But fine with anything ...

electrum:
No problem, that's why I kept it as a separate commit. I dropped it from this PR.

I'll also follow up with a README section for this.

andreaTP (approved):
@electrum feel free to merge at your convenience, we can work on top of this in separate PRs.

pom.xml
electrum: Setting the scope in dependency management and then overriding it when necessary seems error prone.

As long as we review any non-test dependencies added to the core modules, we shouldn’t need to worry about things being added accidentally. The sorting based on scope makes it easy to see. They currently have no dependencies and that seems to be a project goal, so it should be easy to review.

electrum: Note that `test-gen-plugin` doesn’t depend on Guava directly. It’s a transitive dependency. So we’d have to add it as a dependency just to override the scope.

andreaTP: > currently have no dependencies and that seems to be a project goal, so it should be easy to review

It's definitely a project goal! Fine with it :+1:

andreaTP:
@electrum I took a look at the [Takari Maven Testing](https://github.com/takari/takari-plugin-testing-project) project, and looks pretty interesting, although I decided(for now), to rely on a "go-to" solution we are already using in other projects (e.g. the [fabric8 kubernetes client](https://github.com/fabric8io/kubernetes-client/tree/main/java-generator/it) ).

I do like to have the "final project like" structure, and, even if not optimal, I'm used to debugging with this setup.
Happy to share tips and tricks when necessary.

PR #498, opened 2024-08-29
Format test display name with Wast file name, line and Wasm file name

evacchi:
A minor improvement to how spec tests are generated and displayed that should help with tracking down issues.

This adds the `DisplayName(...)` annotation on most generate tests, so that they are now named `$wast_name.wast:$line` with optionally `@ $wasm_name`. For instance, `test23` in `SpecV1ImportsTest` would show up as `imports.wast:154 @ spec.19.wasm`. 

If you highlight `imports.wast:154` you should be able to jump to that file in your IDE; for instance, with IntelliJ you can hit Shift twice, the "Look anywhere" window will pop up:

Screenshot 2024-08-29 at 12 02 33

Hit enter and you are brought to the right point of the target Wast file:

Screenshot 2024-08-29 at 12 02 48

The Wasm file name is useful for reference and to disasm the binary when necessary.

Suggestions for improvements are welcome. If you don't like it, also fine, but I think something similar would be useful.


andreaTP (approved):
LGTM, thanks!

PR #499, opened 2024-08-29
Make the AOT class name configurable

andreaTP:
Follow up from #494

### DIFF
--- a/aot/src/main/java/com/dylibso/chicory/aot/AotMachine.java
+++ b/aot/src/main/java/com/dylibso/chicory/aot/AotMachine.java
@@ -307,6 +307,10 @@ public final class AotMachine implements Machine {
                     .build();
 
     public AotMachine(Instance instance) {
+        this(DEFAULT_CLASS_NAME, instance);
+    }
+
+    public AotMachine(String className, Instance instance) {
         this.instance = requireNonNull(instance, "instance");
         this.module = instance.module();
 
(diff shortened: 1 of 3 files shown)


aot/src/main/java/com/dylibso/chicory/aot/AotMachine.java
```diff
@@ -306,7 +307,15 @@ public final class AotMachine implements Machine {
                     .shared(OpCode.F64_TRUNC, OpcodeImpl.class)
                     .build();
 
+    public static Function<Instance, Machine> withClassName(String className) {
```
electrum, line 310: The reason to allow setting the class name is for actual ahead-of-time purposes. Users of `AotMachine` today will never need to set this, so I don't think we need this factory.

electrum (approved):

aot/src/main/java/com/dylibso/chicory/aot/AotMachine.java
andreaTP: It was just appealing to add it ... if you have a strong opinion, I have no problems removing it as it's just cosmetics :-)

electrum: I just don't think it's useful as it won't be used outside of tests. It seems to encourage users to use it, even though it doesn't do anything.

andreaTP: ok, removed :+1:

PR #501, opened 2024-08-29
Improve immutability of the data structure of the wasm module

andreaTP:
Opening in draft as I expect this PR to have impacts on performance.

@electrum would you like to propose improvements on top of this?


electrum (approved):

andreaTP:
@electrum now that we merged #517 should we close this one?
From your comments, looks like we only miss a couple of defensive copies of Lists, correct?

electrum:
I think most of these changes are separate from #517.

andreaTP:
Cool :+1: thanks for reviewing!

electrum:
What I meant is that we still need this PR, as most of the changes here are different from the ones in #517.

andreaTP:
Ouch, ok, I'll keep it on the Todo list.

PR #507, opened 2024-09-02
Add annotation support for string parameters

electrum:
The `@Buffer` annotation assumes a `POINTER_LENGTH` layout, which seems common. If needed, we can add an enum to support a `LENGTH_POINTER` layout, with the default being the former.

The `@Buffer` annotation could also be extended to support `byte[]`, additional character sets for strings, etc.

While there is lots more that we could do with layouts, this seems like a good step to handle a common case. It makes the logging for WASI significantly more useable, as we now have strings instead of pointers. This would allow us to add automatic logging/tracing support to the binding generator.

### DIFF
--- a/function-processor/src/test/resources/Simple.java
+++ b/function-processor/src/test/resources/Simple.java
@@ -1,17 +1,40 @@
 package chicory.testing;
 
+import static java.util.Objects.requireNonNull;
+
+import com.dylibso.chicory.function.annotations.Buffer;
+import com.dylibso.chicory.function.annotations.CString;
 import com.dylibso.chicory.function.annotations.HostModule;
 import com.dylibso.chicory.function.annotations.WasmExport;
 import com.dylibso.chicory.runtime.HostFunction;
 import com.dylibso.chicory.runtime.Memory;
 import com.dylibso.chicory.wasm.exceptions.ChicoryException;
+import java.util.Random;
 
 @HostModule("simple")
 public final class Simple {
 
+    private final Random random;
+
+    public Simple(Random random) {
+        this.random = requireNonNull(random);
+    }
+
     @WasmExport
-    public void print(Memory memory, int ptr, int len) {
-        System.out.println(memory.readString(ptr, len));
+    public void print(@Buffer String data) {
+        System.out.println(data);
+    }
+
+    @WasmExport
+    public void printx(@CString String data) {
+        System.out.println(data);
+    }
+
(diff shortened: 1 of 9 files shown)


function-processor/src/test/resources/Simple.java
```diff
@@ -1,17 +1,32 @@
...
+
+    @WasmExport
+    public void printx(@CString String data) {
+        System.out.println(data);
+    }
+
+    @WasmExport
+    public void randomGet(Memory memory, int ptr, int len) {
+        byte[] data = new byte[len];
+        new SecureRandom().nextBytes(data);
```
andreaTP, line 28: nitpick: instead of generating random data we can stick with a predefined constant to avoid flakiness and make the test reproducible.

andreaTP (approved):
I like where this PR is going, at the same time I don't know how we can identify and generate the code for this "specialization".
A note on using Random in tests and LGTM, thanks!

electrum:
> I don't know how we can identify and generate the code for this "specialization"

Not sure what you mean, can you elaborate?

function-processor/src/test/resources/Simple.java
electrum: This code is never actually called -- we only verify the generated module source code. But I'll change this to have a `Random` instance as a constructor argument, as it's likely that host modules will have instance fields in practice.

andreaTP:
> Not sure what you mean, can you elaborate?

Sure, sorry, I forgot to give context.
I see those changes useful in general, but, if I understand it correctly, they are not yet covering real-world use-cases:

- Are we assuming that the memory is always already allocated?
- How do you easily "free" a `String` allocated memory? [use-case](https://github.com/dylibso/chicory/pull/506/files#diff-6ae6c6d43974944561891d7bfc45e572fca4e97c37547b393527d5ad08ba97f0R31-R33)
- How does this play with explicit "malloc" calls? [use-case](https://github.com/dylibso/chicory/pull/506/files#diff-6ae6c6d43974944561891d7bfc45e572fca4e97c37547b393527d5ad08ba97f0R21-R24)

electrum:
Ah, I understand what you mean. I don't think those are an issue for arguments to host functions, since the memory will be already allocated. Returning strings from a host function should be the same problem as passing strings to an export function.

We would need to know

* how to allocate and free memory
* if memory should be freed (i.e. is the passed or returned value owned by the function or the caller)

PR #508, opened 2024-09-02
Remove size limit for WASI random_get

electrum:


wasi/src/main/java/com/dylibso/chicory/wasi/WasiPreview1.java
```diff
@@ -1077,13 +1078,18 @@ public int randomGet(Memory memory, int buf, int bufLen) {
...
-        byte[] data = new byte[bufLen];
-        new SecureRandom().nextBytes(data);
-        memory.write(buf, data);
+        byte[] data = new byte[min(bufLen, 4096)];
+        int written = 0;
+        while (written < bufLen) {
+            if (Thread.currentThread().isInterrupted()) {
+                throw new ChicoryException("Thread interrupted");
+            }
+            new SecureRandom().nextBytes(data);
```
andreaTP, line 1088: I think that `SecureRandom` should be instantiated outside the loop.
It would be ideal to add a "Random source" to the `WasiOptions` class, I can follow up with the changes.

wasi/src/main/java/com/dylibso/chicory/wasi/WasiPreview1.java
```diff
@@ -1077,13 +1078,18 @@ public int randomGet(Memory memory, int buf, int bufLen) {
...
-            throw new WASMRuntimeException("random_get: bufLen must be < 100_000");
-        }
 
-        byte[] data = new byte[bufLen];
-        new SecureRandom().nextBytes(data);
-        memory.write(buf, data);
+        byte[] data = new byte[min(bufLen, 4096)];
+        int written = 0;
+        while (written < bufLen) {
+            if (Thread.currentThread().isInterrupted()) {
```
andreaTP, line 1085: This is interesting, basically we should have this check in any "long enough" operation.

andreaTP (approved):
A couple of notes but lgtm, having a test would be ideal but not a requirement at this stage.
Thanks!

wasi/src/main/java/com/dylibso/chicory/wasi/WasiPreview1.java
electrum: Yeah, I was thinking we might want an `ExecutionInterruptedException` or `ExecutionAbortedException` with a static utility method to do this check

wasi/src/main/java/com/dylibso/chicory/wasi/WasiPreview1.java
electrum: Good idea! I'll add a follow up commit.

electrum: Good idea! I'll add another commit.

electrum:
@andreaTP I updated `WasiOptions` to accept `Random` and added a test which verifies that large buffers are handled and that the provided `Random` is used correctly.

electrum:
I checked `wasi-libc` and found that `random_get()` is only used for `getentropy()` which is limited to 256 bytes, so the previous limitation won't be an issue in practice, but it's good not to have arbitrary limits.

wasi/src/main/java/com/dylibso/chicory/wasi/WasiPreview1.java
andreaTP: :+1:

wasi/src/test/java/com/dylibso/chicory/wasi/WasiPreview1Test.java
```diff
@@ -123,4 +127,25 @@ public void shouldRunWasiDemoDotnetModule() throws Exception {
 
         assertEquals("Hello, Wasi Console!\n", fakeStdout.output());
     }
+
+    @Test
+    public void wasiRandom() {
+        var seed = 0x12345678;
```
andreaTP, line 133: perfect :+1:

PR #514, opened 2024-09-05
Separate non-Maven-specific code in OpCodeGen into a separate file

jmillikin:
Fixes https://github.com/dylibso/chicory/issues/513

Issue #513, opened 2024-09-05
Split OpCodeGenMojo so it's easier to run from non-Maven contexts


jmillikin:
I'm using Chicory in a project that builds its external dependencies from source using [Bazel](https://bazel.build/), instead of fetching pre-compiled artifacts from Maven.

The existing `OpCodeGenMojo.java` is a little awkward to use from non-Maven build systems because all of the codegen logic is embedded within a Maven plugin. It would be more convenient for non-Maven builds if that file was split into two parts:

* `OpCodeGen.java` which contains the codegen logic as a plain function, with a signature along the lines of `class OpCodeGen { void generate(File instructionsFile, File sourceDestinationFolder) throws IOException { ... } }`
* `OpCodeGenMojo.java` which wraps `OpCodeGen` in the Maven stuff.

This would let me shim in a Bazel-compatible `main()` codegen tool without having to patch any upstream files.

andreaTP:
Hi! Thanks for reaching out and for your interest in this project!

In Red Hat we have similar concerns with rebuilding internally stuff, so I'm sympathetic to this request.
I don't have any concern with separating the Maven Plugin implementation (is it enough to be in a separate file? or does it requires a separate module as well?).
Since adding new opcodes becomes less and less frequent those days, one option is to get rid of the `.tsv` file and simply continue to maintain the resulting java file as a normal source file (it will also simplify the build); I'll file a PR to discuss this with the other maintainers.

The next question is:
are you going to run the tests internally with Bazel?

andreaTP:
@jmillikin I fixed the issues with your PR and opened an alternative one with the proposed alternative approach.

jmillikin:
> I don't have any concern with separating the Maven Plugin implementation (is it enough to be in a separate file? or does it requires a separate module as well?).

A separate file is fine. Bazel is file-based (it's like a sandboxed `make`), so it'll ignore any files that I haven't specified in its build targets. I figured putting both split files in the same directory would keep Maven happy.

> are you going to run the tests internally with Bazel?

I'm planning to use Chicory as a dependency in a project that uses Bazel as a build system, and have no plans to run the Chicory test suite in Bazel -- I just need the stuff in `wasm/src/main/java/`.

If that work motivates any contributions to Chicory itself, I would use Maven for builds + testing. Trying to contribute to a project using a different build system than upstream doesn't seem like a great idea, especially since Chicory doesn't seem to have automated tests configured in GitHub yet.

[...]

PR #515, opened 2024-09-05
Remove wasm-support-plugin and deal with the code

andreaTP:
Alternative ( to #514 ) fix for #513 

I think this is simpler and, at the end of the day, more easily maintainable (fun fact: I have seen multiple ppl stumbling on the `.tsv` as IntelliJ was not respecting the tabs :sweat_smile: ).


jmillikin:
(author of #513 / #514 here)

I don't have a strong preference either way -- if this PR gets merged I'll close mine.

If you expect to add more instructions over time (for example when adding support for SIMD or multi-memory) then keeping the table might be better. Java is not generally considered a terse language, so having a bit of codegen can be useful for making the programmer's intent clear. 

> fun fact: I have seen multiple ppl stumbling on the `.tsv` as IntelliJ was not respecting the tabs 😅 

I was definitely surprised to see TSV used here. Similar tables I've seen in other projects treat tabs and spaces as just plain whitespace, which tends to survive editor round-trips better. For an example see the Linux kernel's [`arch/arm/tools/syscall.tbl`](https://github.com/torvalds/linux/blob/v6.10/arch/arm/tools/syscall.tbl). I've also seen people use structured formats such as lines of JSON, which can be nice when items are heterogenous.

Regardless of which approach is used, thank you for taking the time. I'll answer your questions in the other PR in case that helps with the decision.

wasm/src/main/java/com/dylibso/chicory/wasm/types/OpCode.java
```diff
@@ -0,0 +1,650 @@
...
+
+    private static final Map<OpCode, WasmEncoding[]> signature = new HashMap<>();
+
+    public static WasmEncoding[] getSignature(OpCode o) {
+        return signature.get(o);
+    }
+
+    static {
+        for (OpCode e : OpCode.values()) {
+            byOpCode.put(e.opcode(), e);
```
electrum, line 437: Might want a validation here
```java
if (byOpCode.put(e.opcode(), e) != null) {
    throw new AssertionError("duplicate opcode: " + e.opcode());
}

**Review by electrum — APPROVED** [review]
I like getting rid of the code generation for this. Adding new instructions here is very minor compared to the rest of the implementation.

File: wasm/src/main/java/com/dylibso/chicory/wasm/types/OpCode.java [hunk 2]
```diff
@@ -0,0 +1,650 @@
...
+
+    private static final Map<Integer, OpCode> byOpCode = new HashMap<>();
+
+    public static OpCode byOpCode(int opcode) {
+        return byOpCode.get(opcode);
+    }
+
+    private static final Map<OpCode, WasmEncoding[]> signature = new HashMap<>();
+
+    public static WasmEncoding[] getSignature(OpCode o) {
```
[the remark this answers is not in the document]
andreaTP: done

wasm/src/main/java/com/dylibso/chicory/wasm/types/OpCode.java
andreaTP: Moved the Map to be an array, I see the value of the check, but is, ultimately, only for us developers of the engine.
I left it out for now.

andreaTP:
@electrum moved to a fully hand rolled implementation, let me know what you think about this!

electrum (approved):
Looks great!

I don't know if an array is needed here. With 8 byte references, each array is 512kB, so this adds a 1MB memory overhead. Does this provide a measurable improvement when parsing WASM files?

If there is no improvement and the memory overhead is a problem for someone, we can switch back to a hash map later.

wasm/src/main/java/com/dylibso/chicory/wasm/types/OpCode.java
electrum: Agreed, and you would find the error immediately anyway, so it's not needed.

wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java
```diff
@@ -906,7 +906,7 @@ private static Instruction parseInstruction(ByteBuffer buffer) {
         }
 
         // System.out.println("b: " + b + " op: " + op);
-        var signature = OpCode.getSignature(op);
+        var signature = OpCode.signature(op);
         // TODO: Encode this in instructions.tsv ?
```
electrum, line 910: Remove outdated comment

andreaTP: done

PR #517, opened 2024-09-07
Make Instruction class immutable

electrum:
This keeps the `long[]` to avoid boxing, which might have a performance impact for the interpreter. It also adds a constant for the common case of an empty array, which avoids allocations and reduces memory usage.

The extra clone for instructions created during parsing could be eliminated by adding a trusted package-private constructor, but I hope we don't need to optimize to that level.

### DIFF
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/InterpreterMachine.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/InterpreterMachine.java
@@ -125,8 +125,8 @@ class InterpreterMachine implements Machine {
             //                                + " stack="
             //                                + stack);
             var opcode = instruction.opcode();
-            var operands = instruction.operands();
-            instance.onExecution(instruction, operands, stack);
+            Operands operands = instruction::operand;
+            instance.onExecution(instruction, stack);
             switch (opcode) {
                 case UNREACHABLE:
                     throw new TrapException("Trapped on unreachable instruction", callStack);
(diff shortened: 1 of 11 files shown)


electrum:
I missed a critical place in the interpreter that calls `operands()` and passes the array to the static methods.

electrum:
This should be fixed now. I added a private `Operands` interface in `InterpreterMachine` that is functionally identical to `IntToLongFunction`, but makes the code read better (the latter looked ugly).

electrum:
Screenshot 2024-09-09 at 9 27 41 PM

bhelx:
good catch @andreaTP :)

runtime/src/main/java/com/dylibso/chicory/runtime/InterpreterMachine.java
```diff
@@ -125,8 +125,8 @@ static void eval(MStack stack, Instance instance, Deque<StackFrame> callStack)
             //                                + " stack="
             //                                + stack);
             var opcode = instruction.opcode();
-            var operands = instruction.operands();
-            instance.onExecution(instruction, operands, stack);
+            Operands operands = instruction::operand;
+            instance.onExecution(instruction, stack);
```
andreaTP, line 129: nice clean up!
`operands` was redundant here.

runtime/src/main/java/com/dylibso/chicory/runtime/InterpreterMachine.java
```diff
@@ -1982,4 +1982,9 @@ private static void checkInterruption() {
             throw new ChicoryException("Thread interrupted");
         }
     }
+
+    @FunctionalInterface
+    private interface Operands {
```
andreaTP, line 1987: I like this trick!

andreaTP (approved):
Thanks for the double check and the quick fix @electrum

PR #518, opened 2024-09-08
Add README section for host modules

electrum:


andreaTP (changes requested):
Thanks for the docs!

Since we are not really done with `CString` and `Buffer` I would leave those out from the documentation a little longer.

README.md
```diff
@@ -291,6 +291,97 @@ writeResultFile("hostFunction.result", hostFunctionResult);
...
+system you are using. This is how to do it with Maven:
+
+```xml
+<plugin>
+  <groupId>org.apache.maven.plugins</groupId>
+  <artifactId>maven-compiler-plugin</artifactId>
+  <configuration>
+    <annotationProcessorPaths>
+      <path>
+        <groupId>com.dylibso.chicory</groupId>
```
andreaTP, line 376: This is good and correct, but, what I normally see is to just drop the dependency in the `provided` scope and expect that everything will be picked up automatically:
https://github.com/dylibso/chicory/pull/506/files#diff-97b0981589bcc6a36fc5ef766650d723817a031ccdad83102f28585cfdd23e41R26-R31

README.md
```diff
@@ -291,6 +291,97 @@ writeResultFile("hostFunction.result", hostFunctionResult);
...
+        <artifactId>function-processor</artifactId>
+        <version>0.0.12</version>
```
andreaTP, line 378: What about using a placeholder for the version number?
Alternatively, it needs to be updated by the release process by this script:
https://github.com/dylibso/chicory/blob/main/scripts/update-readme-version.sh

README.md
electrum: Using annotation processors on the compiler class path is legacy behavior. That's disabled by default in JDK 22+. I think older versions warn about this.

It's also bad because it "infects" with class path with code that ie only needed for annotation processing. So it's better to configure them separately.

README.md
electrum: Looks like the script will already update it, as it's the same format as a normal dependency?

README.md
andreaTP: > That's disabled by default

Good to hear!
Thanks for the explanation.

andreaTP:
This needs to target the new `docs` site.

Maybe a good first issue for @kanatohodets  that demonstrated interest in working on the documentation side?

andreaTP:
Superseded by #628

PR #519, opened 2024-09-09
Add annotation framework for Wasm interfaces

electrum:
This is an alternative approach to #506. See `Demo` and `DemoGenerated` for what the interface looks like.

It still needs actual runnable tests, maybe the OPA code plus some examples from `wasm-corpus`, but I wanted to put this up now to get some feedback. Also, needs tests for error handling.

### DIFF
--- a/function-processor/src/test/resources/wasm/Demo.java
+++ b/function-processor/src/test/resources/wasm/Demo.java
@@ -1,0 +1,35 @@
+package chicory.testing;
+
+import com.dylibso.chicory.function.annotations.Allocate;
+import com.dylibso.chicory.function.annotations.Buffer;
+import com.dylibso.chicory.function.annotations.CString;
+import com.dylibso.chicory.function.annotations.Free;
+import com.dylibso.chicory.function.annotations.WasmExport;
+import com.dylibso.chicory.function.annotations.WasmModule;
+import com.dylibso.chicory.runtime.Instance;
+import com.dylibso.chicory.wasm.types.Value;
+
+@WasmModule
+public interface Demo {
+
+    @Allocate
+    @WasmExport
+    int malloc(int size);
+
+    @Free
+    @WasmExport
+    void free(int ptr);
+
+    @WasmExport
+    void print(@CString String data);
+
+    @WasmExport
+    int length(@Buffer String data);
+
+    @WasmExport
+    Value[] multiReturn(float x);
+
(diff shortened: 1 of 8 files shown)


andreaTP (changes requested):
Thanks a lot @electrum for putting this up to show where you are going, super useful!

Before jumping on a solution that is going this "high level" I would love to share some context regarding (my limited understanding of) CM, integration frameworks(extistm/wasm-bindgen/etc.) and the landscape we know about.

As a full disclosure, I'm comfortable moving forward and merging PRs up to:
- my proposal in #506 moving to an annotation processor
- this PR, without handling of Alloc / Free/ String/ Buffer etc.

(probably the final result will look pretty similar)

Would you be able to join tomorrow's call?
Alternatively, we can setup something ad-hoc, possibly, I'd like to have @evacchi and Ben's experience on the table.

electrum:
Let's talk about it tomorrow on the calle.

function-processor/src/test/resources/wasm/Demo.java
```diff
@@ -0,0 +1,35 @@
...
+import com.dylibso.chicory.function.annotations.Allocate;
+import com.dylibso.chicory.function.annotations.Buffer;
+import com.dylibso.chicory.function.annotations.CString;
+import com.dylibso.chicory.function.annotations.Free;
+import com.dylibso.chicory.function.annotations.WasmExport;
+import com.dylibso.chicory.function.annotations.WasmModule;
+import com.dylibso.chicory.runtime.Instance;
+import com.dylibso.chicory.wasm.types.Value;
+
+@WasmModule
```
bhelx, line 12: This is more inline with what I was thinking. I like the annotation approach and providing explicit c types. What I think we're trying to accomplish is more akin to what JNA does today. Not necessarily write the high-level bindings for you yet, but give you some types and code to keep you from having to explicitly encode the ABI level details when we can just borrow from some known ABI and types (usually C). This reduces repetition and mistakes as @andreaTP pointed out. 

I can't make the call today but I'd vote for this approach. Feel free to make a decision without me though I trust y'all to figure out the right approach.

andreaTP:
@electrum I'd like to close this one as I'm going to work on those stuff anyway sometime soon.
Would you agree?

Thanks a lot for the inspiration and help in shaping this out!

andreaTP:
Agreed on call, I'll take over this effort :+1:

PR #523, opened 2024-09-11
No jgit repo downloaders

andreaTP:
JGit [dropped support for Java 11](https://projects.eclipse.org/projects/technology.jgit/releases/7.0.0) with version 7.

It's convenient, but we don't really need it, here I'm replacing it using [this feature](https://docs.github.com/en/repositories/working-with-files/using-files/downloading-source-code-archives#source-code-archive-urls) of GitHub.

Tested on Windows in CI [here](https://github.com/andreaTP/chicory/actions/runs/10809074688)

Supersede #522

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -86,7 +86,7 @@
     <commons-lang.version>3.17.0</commons-lang.version>
     <jackson.version>2.17.2</jackson.version>
     <javaparser.version>3.26.2</javaparser.version>
-    <jgit.version>6.10.0.202406032230-r</jgit.version>
+    <jgit.version>7.0.0.202409031743-r</jgit.version>
     <jmh.version>1.37</jmh.version>
     <junit.version>5.11.0</junit.version>
     <maven-plugin-annotations.version>3.15.0</maven-plugin-annotations.version>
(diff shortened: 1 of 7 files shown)


evacchi (approved):

pom.xml
```diff
@@ -86,7 +86,7 @@
     <commons-lang.version>3.17.0</commons-lang.version>
     <jackson.version>2.17.2</jackson.version>
     <javaparser.version>3.26.2</javaparser.version>
-    <jgit.version>6.10.0.202406032230-r</jgit.version>
+    <zip4j.version>2.11.5</zip4j.version>
```
electrum, line 89: What features is `java.util.zip.ZipFile` missing?

andreaTP: None afaik, it was just much more convenient to use a higher level library (less LOCs, maintenance etc.).

andreaTP: @electrum the dependency is not exposed to the end user and only used internally in Maven plugins, do you think we should go the extra mile to remove it?

electrum (approved):

pom.xml
electrum: Sorry, I should have clarified that this was just a question and not a blocking concern about the PR.

PR #527, opened 2024-09-16
Rename HostImports to ExternalValues

evacchi:
a.k.a. **The Big Rename™️**. This follows an offline discussion with @andreaTP. The idea would be to finalize naming before the future 1.0 release.

The class `HostImports` and the related dependencies `HostFunction`, `HostTable`, `HostGlobal`, `HostMemory` are currently being used with two different purposes:

1. allowing a **host**, that is, the application that embeds the Wasm runtime, to define such values programmatically using the **host** language (especially true for functions)
2. allowing **the runtime** to link together two different module instances, connecting their "import" requirements to the "exports" of other instances.

However, name "Host<something>" makes most sense for the first case, while it feels slightly off and ambiguous in the second case: are those Host-defined values, or are those just values belonging to another instance?

The spec in this regard is not ambiguous, and generally refers to [**External Types**](https://webassembly.github.io/spec/core/syntax/types.html#external-types) and [**External Values**](https://webassembly.github.io/spec/core/exec/runtime.html#syntax-externval). Specifically:

- **External Types** classify **imports** and **external values**. Notice [we are already using the term `ExternalType`](https://github.com/dylibso/chicory/blob/dc6959dbf20544517374d93e1f7c255d316ecdaa/wasm/src/main/java/com/dylibso/chicory/wasm/types/ExternalType.java#L13) under package `com.dylibso.chicory.wasm.types`.
- I think **External Values** identifies more correctly what we are currently calling **HostImports**; i.e. instances of the respective types (instances of functions, of globals, of memories or of tables),  capturing the fact that these values are just "externally defined", not necessarily by **the host**, but possibly by an instance of another module.

Thus, with this change, I propose the following renames:

- `FromHost` -> `ExternalValue`
- `FromHost.FromHostType` -> `ExternalValue.Type` notice that we could also reuse `com.dylibso.chicory.wasm.types. ExternalType`, but this would require exposing to the end users package `com.dylibso.chicory.wasm.types` from package `com.dylibso.chicory.runtime`
- `HostImports` -> `ExternalValues`
- `HostGlobal` -> `ExternalGlobal`
- `HostTable` -> `ExternalTable`
- `HostMemory` -> `ExternalMemory`
- `HostFunction` still exists, as a *subtype* of `ExternalFunction`: this allows to keep a distinction that might be useful in the future; in fact, the constructor to `ExternalFunction` could be even package-local. 

Notice that, I have not re-defined `Host{Global,Table,Memory}` as subtypes of `External{Global,Table,Memory}`, because I feel like there is no meaningful distinction:
- the implementation for the "host" version does not conceptually differs from the "wasm" version
- on the other hand, `HostFunction`s implement logic using the host language, while `ExternalFunction`s are otherwise just the exported image of an instance of a `Function` defined in a Wasm module. So, even though their interface might be the same, they are different concepts.

Even if we decided to drop the concrete type `HostFunction` altogether, it might still make sense to expose a factory `ExternalFunction.createHostFunction() for an `ExternalFunction` because the term "host function" is generally known to refer to a host-defined function. 

It might be now feel slightly more awkward to import host functions directly through an `Instance.Builder`; e.g.:

```
var wasi = new WasiPreview1(logger, wasiOpts);
var hostFunctions = new ExternalValues(wasi.toHostFunctions());
var inst = Instance.builder(Parser.parse(new File("greet-wasi.wasm")))
              .withExternalValues(hostFunctions).build();
```

but consider that, on the other hand, the `Instance.Builder` interface is a lower-level  interface, so it might be acceptable for it to expose "lower-level" naming; in the future, I would expect users to interact with the runtime using higher level interfaces, such as the `Store`. In this case, users would write:

```
var wasi = new WasiPreview1(logger, wasiOpts);
var store = new Store();
store.addFunctions(wasi.toHostFunctions());
var inst = store.instantiate("main", Parser.parse(new File("greet-wasi.wasm")));
```

### Notes

Will follow-up with another, conceptually unrelated rename, but with the same goal of improved clarity; i.e.:

- `fieldName()` -> `symbolName()`

reason being `fieldName()` to me sounds too much like "name of a property", that is, it makes me think "Global" but we also use it for "Function"s. "Symbol" is more generic and should avoid ambiguity.

Signed-off-by: Edoardo Vacchi <evacchi@users.noreply.github.com>


runtime/src/main/java/com/dylibso/chicory/runtime/HostFunction.java
```diff
@@ -3,50 +3,16 @@
...
 
-public class HostFunction implements FromHost {
-    private final WasmFunctionHandle handle;
-    private final String moduleName;
-    private final String fieldName;
-    private final List<ValueType> paramTypes;
-    private final List<ValueType> returnTypes;
-
+/**
+ * A HostFunction is an ExternalFunction that has been defined by the host.
```
andreaTP, line 7: If we keep the distinction, we should have also, at least `HostMemory` too.

In general, I would go for the simplest solution and just use `ExternalFunction` everywhere.

runtime/src/main/java/com/dylibso/chicory/runtime/ExternalFunction.java
```diff
@@ -0,0 +1,52 @@
...
+public class ExternalFunction implements ExternalValue {
+    private final WasmFunctionHandle handle;
+    private final String moduleName;
+    private final String fieldName;
+    private final List<ValueType> paramTypes;
+    private final List<ValueType> returnTypes;
+
+    public ExternalFunction(
+            WasmFunctionHandle handle,
+            String moduleName,
```
andreaTP, line 15: nitpick I always forget to do: since you are at it, I would get this constructor signature aligned with the rest of `External`Things with `moduleName` and `fieldName` in first and second position.

andreaTP (commented):
Thanks for the proposal, I like it overall!

runtime/src/main/java/com/dylibso/chicory/runtime/HostFunction.java
evacchi: I feel like it is worth keeping the distinction. 

IMO the two might diverge at some point i.e. while we want HostFunction to be user-facing, the internals/mechanics of a generic ExternalFunction (e.g. when it deals with exports from other instances) might be different. This is not necessarily (only) a matter of implementation, but also a matter of clarity.

OTOH I don't think an ExternalMemory defined by the Host is different from any other host memory.

Again, I'd rather keep the concept of HostFunction because it's what people are familiar with; but maybe this can be solved with a named factory somewhere (as mentioned in the body of the PR) :)

evacchi: for a more authoritative source, notice [how HostFunctions are separate from FunctionInstances in the spec](https://webassembly.github.io/spec/core/exec/runtime.html#function-instances), while there is no distinction between host-supplied and "plain" [table instances](https://webassembly.github.io/spec/core/exec/runtime.html#table-instances), [memory instances](https://webassembly.github.io/spec/core/exec/runtime.html#memory-instances) or [global instances](https://webassembly.github.io/spec/core/exec/runtime.html#global-instances).

pom.xml
```diff
@@ -546,6 +546,7 @@
               -Xep:OperatorPrecedence:OFF \
               -Xep:ReferenceEquality:OFF \
               -Xep:UnnecessaryParentheses:OFF \
+              -Xep:InlineMeSuggester:OFF \
```
evacchi, line 549: I am disabling here the `InlineMeSuggester` which causes the build to fail if the annotation target can be inlined; in this case

https://errorprone.info/bugpattern/InlineMeSuggester. 

```
[WARNING] /Users/evacchi/Devel/java/chicory/runtime/src/main/java/com/dylibso/chicory/runtime/HostFunction.java:[15,12] [InlineMeSuggester] This deprecated API looks inlineable. If you'd like the body of the API to be automatically inlined to its callers, please annotate it with @InlineMe. NOTE: the suggested fix makes the method final if it was not already.
    (see https://errorprone.info/bugpattern/InlineMeSuggester)
  Did you mean '@InlineMe(replacement = "this(moduleName, fieldName, handle, paramTypes, returnTypes)")'?
```

while this is useful, it requires to add a run-time dependency on error-prone, which I assume we want to avoid.

evacchi:
@bhelx I think the docs look good, from what I can see the only place where we still used "host" _"improperly"_ (so to speak) was a line of the WASI readme:

https://github.com/dylibso/chicory/commit/e774e2661e3215fcc0f31932afa3793b8646ef65#diff-2a0f73c29b5710acc7ebe0330361345e74a05b4e51a027012ccbf9862ed6779a

runtime/src/main/java/com/dylibso/chicory/runtime/HostFunction.java
andreaTP: I'm convinced, thanks for explaining and finding the sources! Appreciated :+1:

pom.xml
andreaTP: > it requires adding a run-time dependency on error-prone

this should not be accurate, error-prone is used only in this repository and is not required as a dependency as it's declared in the `annotationProcessorPaths` section.

Instead of fully disabling the check I suggest (if applicable):
- a local warning suppression
- to fix the warning

[example reference](https://github.com/dylibso/chicory/blob/9594ed738d7a64976d47def85e0f917c2644c339/log/src/main/java/com/dylibso/chicory/log/Logger.java#L78) cc. @electrum

evacchi: no, I mean that adding `@InlineMe` as suggested requires adding the run-time dependency to the artifact that contains that annotation. 

I can indeed add a local SuppressWarning as I originally did in this commit https://github.com/dylibso/chicory/pull/527/commits/0d5e01d713d68cd56290a8ecd7013e96bfc6a569. I can revert the global disable if you prefer. 

TBH I can also remove the `@Deprecated` method altogether; I felt it was useful to give a short grace period to end users; but OTOH we are indeed renaming quite a few other symbols...

andreaTP: > I can also remove the `@Deprecated` method altogether

Strongly agree, we don't want to have `@Deprecated` code in a 1.0 and, since the last release, we are already beyond any possible "grace period" for the user.
At the moment, we are releasing `0.0.x` versions, and API breaking changes are expected.

runtime/src/main/java/com/dylibso/chicory/runtime/ExternalFunction.java
andreaTP: thanks :+1:

evacchi:
FYI I have pushed the other refactor I have mentioned in the top post `ExternalValue#fieldName() -> ExternalValue#symbolName()` to this branch https://github.com/evacchi/chicory/commit/60026a0edde8857ec8c7563e026ccd7be769e89f

Using `fieldName` might cause confusion now (a "field" in a module to me sounds like a "global" value, not a function) and in the future; e.g. [fields](https://webassembly.github.io/gc/core/syntax/types.html#syntax-fieldtype) are the fields in a `Structure` (see WasmGC);

I can create another PR once this gets merged or just push it here. The changes are related, but they don't need to be shipped together; but if we agree on this rename too, then it makes sense to ship them in the same release.

evacchi:
rebased due to conflict

andreaTP (approved):
LGTM, thanks!

PR #528, opened 2024-09-16
[WIP] Initial SIMD support

andreaTP:
I wanted to relax a bit, and took this yak shaving task.
I realized that the SIMD support needs a few little tweaks to the public API, and, the initial support, requires quite a bit of wiring rising the bar for possible contributions.

Notes:

- I discussed with @evacchi and the usage of the Stack with double push and double pop comes straight from `wazero` I think it's a good approach for further optimizations!
- the vector API, at the moment requires copy of arrays, I should resurrect and finalize #447 to improve here
- there is only one operation actually implemented as a Vector instruction `V128_NOT` but demonstrates that the entire wiring works
- I'll clean up the PR in the next days and try to finalize it (with CI setup etc.)


evacchi:
I am not 100% sold on the approach `Map<Opcode, OpImpl>`, especially because `OpImpl` causes the indirection of a method reference. All SIMD opcodes are [prefixed with 0xFD](https://pengowray.github.io/wasm-ops/#simd) so one approach could be

- in `InterpreterMachine`, add a new `case` with `SIMD_PREFIX(0xFD)`; 
- if SIMD support is enabled, then delegate to another class/method that implements the _switch table_ for that set of opcodes

while this requires adding `SIMD_PREFIX(0xFD)` to the `Opcode` enum, it also allows to raise a proper error message if the feature is not enabled.

andreaTP:
Thanks for your proposal @evacchi , I have no strong opinions here, happy to go down with it if there is agreement, let's @bhelx or @electrum break the tie :slightly_smiling_face:

evacchi:
BTW the comment on the approach is general, and I am not exceedingly attached to it, in general the infra works and we can refine further at another interation

bhelx:
I have no issue with it because i understand what it's doing and it seems reasonable. But I am notoriously unpicky about these things :)

I'd say just merge it, all the code is always up for being refactored. If you're feeling uneasy about something I generally leave a comment or make a follow-up ticket.

wasm/src/main/java/com/dylibso/chicory/wasm/types/OpCode.java
```diff
@@ -211,7 +211,24 @@ public enum OpCode {
     TABLE_COPY(0xFC0E, List.of(VARUINT, VARUINT)),
     TABLE_GROW(0xFC0F, List.of(VARUINT)),
     TABLE_SIZE(0xFC10, List.of(VARUINT)),
-    TABLE_FILL(0xFC11, List.of(VARUINT));
+    TABLE_FILL(0xFC11, List.of(VARUINT)),
+    V128_LOAD(0xFD00, List.of(VARUINT, VARUINT)),
+    I8x16_SWIZZLE(0xFD0E),
+    V128_CONST(0xFD00 + 13, List.of(VEC_VARUINT)),
```
kanatohodets, line 217: off by one I think:
```bash
$ perl -E 'say 0xFD00 + 13; say 0xFD0C'
64781
64780
```

(some Javy code failed to parse on 0xFD0C which led me to [the table](https://pengowray.github.io/wasm-ops/#simd))

andreaTP:
Superseded by #611

PR #531, opened 2024-09-18
Primitive, non boxing, long array based MStack

andreaTP:
Supersede #447 this is a complete implementation, already passing all the tests up to the interpreter, can be tested with something like:

```
 mvn clean install -pl runtime-tests -am
```

I started, but got a bit confused about the changes necessary in "aot", @electrum or @danielperano I would love some help in finalizing this PR :pray: and learning a bit more how `asm` works ...

This is a prerequisite to avoid arrays construction [here](https://github.com/dylibso/chicory/pull/528/files#diff-f9f476fefd928e6770830c090018da2b95a054834c70bf472460d568fd1a9d2bR184), the plan is to open up the underlying array so that we can perform operations just using the offset :crossed_fingers:

### DIFF
--- a/aot/src/main/java/com/dylibso/chicory/aot/AotMachine.java
+++ b/aot/src/main/java/com/dylibso/chicory/aot/AotMachine.java
@@ -354,9 +354,9 @@ public final class AotMachine implements Machine {
     }
 
     @Override
-    public Value[] call(int funcId, Value[] args) throws ChicoryException {
+    public long[] call(int funcId, long[] args) throws ChicoryException {
         try {
-            return (Value[]) compiledFunctions[funcId].invokeExact(args);
+            return (long[]) compiledFunctions[funcId].invokeExact(args);
         } catch (ChicoryException e) {
             // propagate ChicoryExceptions
             throw e;
@@ -567,7 +567,11 @@ public final class AotMachine implements Machine {
         int slot = type.params().stream().mapToInt(AotUtil::slotCount).sum();
 
         // parameters: arguments, funcTableIdx, tableIdx, instance
-        emitBoxArguments(asm, type.params());
+        // emitBoxArguments(asm, type.params());
+        // asm.visitLdcInsn(LONG);
+        for (int i = 0; i < type.params().size(); i++) {
+            asm.visitVarInsn(Opcodes.LLOAD, i);
+        }
         asm.visitLdcInsn(typeId);
         asm.visitVarInsn(Opcodes.ILOAD, slot); // funcTableIdx
         asm.visitVarInsn(Opcodes.ILOAD, slot + 1); // tableIdx
@@ -575,7 +579,7 @@ public final class AotMachine implements Machine {
 
         emitInvokeStatic(asm, AotMethods.CALL_INDIRECT);
 
-        emitUnboxResult(type, asm);
+        asm.visitInsn(returnTypeOpcode(type));
     }
 
     private static void compileHostFunction(int funcId, FunctionType type, MethodVisitor asm) {
--- a/aot/src/main/java/com/dylibso/chicory/aot/AotMachine.java
+++ b/aot/src/main/java/com/dylibso/chicory/aot/AotMachine.java
@@ -73,6 +73,7 @@ import org.objectweb.asm.MethodVisitor;
 import org.objectweb.asm.Opcodes;
 import org.objectweb.asm.commons.InstructionAdapter;
 import org.objectweb.asm.util.CheckClassAdapter;
+import org.objectweb.asm.util.TraceClassVisitor;
 
 /**
  * Simple Machine implementation that AOT compiles function bodies and runtime-links them
@@ -511,8 +512,11 @@ public final class AotMachine implements Machine {
         } catch (VerifyError e) {
             // run ASM verifier to help with debugging
             try {
+                PrintWriter pw = new PrintWriter(System.out, true, UTF_8);
                 ClassReader reader = new ClassReader(classBytes);
-                CheckClassAdapter.verify(reader, true, new PrintWriter(System.out, false, UTF_8));
+                CheckClassAdapter.verify(reader, true, pw);
+
+                reader.accept(new TraceClassVisitor(pw), 0);
             } catch (NoClassDefFoundError ignored) {
                 // the ASM verifier is an optional dependency
             } catch (Throwable t) {
@@ -589,9 +593,10 @@ public final class AotMachine implements Machine {
     }
 
     private static void emitBoxArguments(MethodVisitor asm, List<ValueType> types) {
+        int slotCount = types.stream().mapToInt(AotUtil::slotCount).sum();
         int slot = 0;
         // box the arguments into long[]
-        asm.visitLdcInsn(types.size());
+        asm.visitLdcInsn(slotCount);
         asm.visitIntInsn(Opcodes.NEWARRAY, Opcodes.T_LONG); // long
         for (int i = 0; i < types.size(); i++) {
             asm.visitInsn(Opcodes.DUP);
--- a/aot/src/main/java/com/dylibso/chicory/aot/AotMachine.java
+++ b/aot/src/main/java/com/dylibso/chicory/aot/AotMachine.java
@@ -3,10 +3,12 @@ package com.dylibso.chicory.aot;
 import static com.dylibso.chicory.aot.AotMethods.CHECK_INTERRUPTION;
 import static com.dylibso.chicory.aot.AotMethods.INSTANCE_CALL_HOST_FUNCTION;
 import static com.dylibso.chicory.aot.AotMethods.THROW_TRAP_EXCEPTION;
-import static com.dylibso.chicory.aot.AotUtil.boxer;
-import static com.dylibso.chicory.aot.AotUtil.boxerHandle;
 import static com.dylibso.chicory.aot.AotUtil.callIndirectMethodName;
 import static com.dylibso.chicory.aot.AotUtil.callIndirectMethodType;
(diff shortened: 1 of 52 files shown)


andreaTP:
update: I benchmarked the interpreter and got good improvements on this branch, I think it's in the right direction!
Regarding the AOT, the changes seem to be a bit more invasive than expected, it would be great to have some help.

aot/src/main/java/com/dylibso/chicory/aot/AotMachine.java
```diff
@@ -567,15 +567,19 @@ private static void compileCallIndirect(MethodVisitor asm, int typeId, FunctionT
         int slot = type.params().stream().mapToInt(AotUtil::slotCount).sum();
 
         // parameters: arguments, funcTableIdx, tableIdx, instance
-        emitBoxArguments(asm, type.params());
+        // emitBoxArguments(asm, type.params());
+        // asm.visitLdcInsn(LONG);
+        for (int i = 0; i < type.params().size(); i++) {
```
electrum, line 572: This needs to use `slotCount()` to compute the slot parameter correctly, as long/double use two slots in JVM bytecode. This annoyance means you always need to know the type you are dealing with.

The incoming parameters to this method are a native Java value: int, long, float, double

aot/src/main/java/com/dylibso/chicory/aot/AotMethods.java
```diff
@@ -183,6 +183,9 @@ public static Value[] callIndirect(
         }
 
         checkInterruption();
+        // TODO: verify
+        // here we should not pass through the external "call" method
+        // but directly emit the invocation of the underlying function
```
electrum, line 188: Agreed, this could be improved, but we still need to put the lookup table somewhere. This should all be replaced with invoke dynamic.

aot/src/main/java/com/dylibso/chicory/aot/AotUtil.java
```diff
@@ -202,7 +202,7 @@ public static Class<?> jvmReturnType(FunctionType type) {
             case 1:
                 return jvmType(type.returns().get(0));
             default:
-                return Value[].class;
+                return long[].class;
```
electrum, line 205: This will break as it's also used by host functions. Or do you plan to change `WasmFunctionHandle` as well?

andreaTP: My plan was to keep the user API layer with the typed `Value[]` for the moment, do you think that we should re-evaluate?

I know this is, anyhow, a low level API, but asking ppl to deal with `long[]` seems a bit of a stretch.

aot/src/main/java/com/dylibso/chicory/aot/AotMachine.java
```diff
@@ -567,15 +567,19 @@ private static void compileCallIndirect(MethodVisitor asm, int typeId, FunctionT
...
+            asm.visitVarInsn(Opcodes.LLOAD, i);
+        }
         asm.visitLdcInsn(typeId);
         asm.visitVarInsn(Opcodes.ILOAD, slot); // funcTableIdx
         asm.visitVarInsn(Opcodes.ILOAD, slot + 1); // tableIdx
         asm.visitVarInsn(Opcodes.ALOAD, slot + 2); // instance
 
         emitInvokeStatic(asm, AotMethods.CALL_INDIRECT);
 
-        emitUnboxResult(type, asm);
```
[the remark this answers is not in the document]
andreaTP: I'm a bit tired now, but I leave this comment anyhow to see the reaction :-)

What if we change all the signatures of the methods `@UsedByGeneratedCode` and the intrinsics to deal with solely `long` types as they are returned from the stack?

Pros:
- having to deal only with the arity we can simplify the code of the aot
- we keep the conversions in plain Java code
- it's easier to reason about
Cons:
- performance? I have no idea of the impact, it would be great to reason on numbers but I would listen to gut feelings

aot/src/main/java/com/dylibso/chicory/aot/AotUtil.java
electrum: I’m suggesting to change it. Rather, if it is not changed, then this code will need to handle the two different styles of return type arrays.

aot/src/main/java/com/dylibso/chicory/aot/AotMachine.java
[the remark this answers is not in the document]
electrum: I don’t think that’s feasible. You would need to convert to and from native types for every operation. This would likely be slower and would bloat the byte code. It’s actually more complicated in many ways than using native types.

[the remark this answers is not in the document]
andreaTP: thanks for the feedback!

aot/src/main/java/com/dylibso/chicory/aot/AotUtil.java
andreaTP: I'm going down the path to change everything to `long[]` in the lower level API.

The plan is to bridge the gap with the code-generation effort right after.

andreaTP: @electrum I think I'm getting there, still a ~1300 are failing, if you have time to take a look, any hint or help is always appreciated :pray:

andreaTP:
I did a little benchmarking on my super noisy machine(take the results with a lot of salt):

```
Chicory 0.0.12:

With GraalJIT:
Benchmark                                      Mode  Cnt      Score        Error  Units
ChicoryAotTest.chicoryAotTest                  avgt    3   1282.401 ±    51.518  ns/op
ChicoryInterpreterTest.chicoryInterpreterTest  avgt    3  63086.906 ± 18526.425  ns/op

Without GraalJIT:
Benchmark                                      Mode  Cnt      Score        Error  Units
ChicoryAotTest.chicoryAotTest                  avgt    3   1222.858 ±    326.915  ns/op
ChicoryInterpreterTest.chicoryInterpreterTest  avgt    3  87630.869 ± 138021.856  ns/op

---

Current main:

With GraalJIT:
Benchmark                                      Mode  Cnt      Score       Error  Units
ChicoryAotTest.chicoryAotTest                  avgt    3   1237.301 ±   423.694  ns/op
ChicoryInterpreterTest.chicoryInterpreterTest  avgt    3  71274.055 ± 11410.525  ns/op

Without GraalJIT:
Benchmark                                      Mode  Cnt       Score       Error  Units
ChicoryAotTest.chicoryAotTest                  avgt    3    1349.806 ±   945.091  ns/op
ChicoryInterpreterTest.chicoryInterpreterTest  avgt    3  101900.141 ± 38884.734  ns/op

---

Primitive longs (this branch):

With GraalJIT:
Benchmark                                      Mode  Cnt      Score       Error  Units
ChicoryAotTest.chicoryAotTest                  avgt    3   1153.859 ±   984.977  ns/op
ChicoryInterpreterTest.chicoryInterpreterTest  avgt    3  41791.372 ± 43360.065  ns/op

Without GraalJIT:
Benchmark                                      Mode  Cnt      Score       Error  Units
ChicoryAotTest.chicoryAotTest                  avgt    3    899.077 ±   483.678  ns/op
ChicoryInterpreterTest.chicoryInterpreterTest  avgt    3  49629.772 ± 88844.628  ns/op

```

aot/src/main/java/com/dylibso/chicory/aot/AotMethods.java
andreaTP: tracked in #540

andreaTP:
CI is green, ready for final review!

evacchi:
this is exciting!

andreaTP:
Begging for review here :pray: cc. @bhelx @electrum @danielperano

evacchi (approved):
I think this should be fine, but I think it would be better to have another pair of eyes on it :)

aot/src/main/java/com/dylibso/chicory/aot/AotMachine.java
```diff
@@ -528,19 +526,19 @@ private static MethodHandle adaptSignature(FunctionType type, MethodHandle handl
...
 
         if (type.returns().isEmpty()) {
-            return result.asType(result.type().changeReturnType(Value[].class));
+            return result.asType(result.type().changeReturnType(long[].class));
         }
         if (type.returns().size() > 1) {
             return result;
         }
-        result = filterReturnValue(result, boxerHandle(type.returns().get(0)));
-        return filterReturnValue(result, ValueWrapper.HANDLE);
```
electrum, line 543: I think `ValueWrapper` is unused now

wasm/src/main/java/com/dylibso/chicory/wasm/types/Value.java
```diff
@@ -50,176 +89,52 @@ public static Value f64(long data) {
...
-                return Value.EXTREF_NULL;
+                return REF_NULL_VALUE;
             default:
                 throw new IllegalArgumentException(
                         "Can't create a zero value for type " + valueType);
         }
     }
 
-    // TODO memoize these
-    public int asInt() {
```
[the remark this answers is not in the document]
andreaTP: in this case the re-ordering is intentional, I know that it makes a worst diff, but the result looks better to me.

aot/src/main/java/com/dylibso/chicory/aot/AotMachine.java
andreaTP: now it's a `LongArrayWrapper` to: `long -> long[]`

andreaTP:
ok, let's merge this, any follow up comment will be fixed on top.

PR #536, opened 2024-09-20
CString annotation uses a single argument

andreaTP:
This is a trivia, but it will spare a couple of lines in the following PRs


electrum (approved):
Good catch

PR #537, opened 2024-09-23
Rename `moduleName()` -> `module()`, `fieldName()` -> `name()` 

evacchi:
Follow up to #527.

Using `fieldName` might cause confusion now (a "field" in a module to me sounds like a "global" value, not a function) and in the future; e.g. [fields](https://webassembly.github.io/gc/core/syntax/types.html#syntax-fieldtype) are the fields in a `Structure` (see WasmGC);

Signed-off-by: Edoardo Vacchi <evacchi@users.noreply.github.com>

### DIFF
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/ExternalFunction.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/ExternalFunction.java
@@ -6,19 +6,19 @@ import java.util.List;
 public class ExternalFunction implements ExternalValue {
     private final WasmFunctionHandle handle;
     private final String moduleName;
-    private final String fieldName;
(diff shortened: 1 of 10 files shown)


bhelx (approved):

andreaTP:
Re-reading [the spec](https://webassembly.github.io/spec/core/syntax/modules.html#imports) it says:

> Each import is labeled by a two-level name space, consisting of a module name and a name for an entity within that module.

Since we are at it, it might be worth to change `name()` with `entity()` to map, more closely, the spec.

andreaTP:
Clarified with @evacchi `entity` is not used anywhere else in the spec. LGTM.

andreaTP (approved):

PR #543, opened 2024-09-27
Fix memory initialization for dynamically linked modules

andreaTP:
An imported memory should not be zeroed during initialization.


wasi/src/test/java/com/dylibso/chicory/wasi/WasiPreview1Test.java
```diff
@@ -80,6 +82,82 @@ public void shouldRunWasiDemoJavyModule() {
         assertEquals(fakeStdout.output(), "{\"foo\":3,\"newBar\":\"baz!\"}");
     }
 
+    @Test
+    public void shouldUseQuickJsProvider() {
```
andreaTP, line 86: Not really relevant to this PR, but it's a good reference for the future on how to programmatically use the provider QuickJS module.

evacchi (approved):
SHIP IT 🚀

bhelx:
Good catch!

PR #550, opened 2024-10-01
WASI improvements

electrum:
Each commit is intended to be reviewed separately.


wasi/README.md
```diff
@@ -31,60 +31,54 @@ If your module calls a wasi function that we don't support, or uses a feature th
...
-| sock_accept             | ❌         |                                                                                                                                                                                        |
-| sock_recv               | ❌         |                                                                                                                                                                                        |
-| sock_send               | ❌         |                                                                                                                                                                                        |
-| sock_shutdown           | ❌         |                                                                                                                                                                                        |
-
-
-> **Note**: 💀 means the function is no longer part of WASI.
+| WASI Function           | Supported | Notes                                                                     |
```
andreaTP, line 34: nice :+1:

wasi/src/main/java/com/dylibso/chicory/wasi/WasiFstFlags.java
```diff
@@ -0,0 +1,14 @@
+package com.dylibso.chicory.wasi;
+
+final class WasiFstFlags {
+    private WasiFstFlags() {}
+
+    public static final int ATIM = bit(0);
```
andreaTP, line 6: can you add a comment with a link to lookup those definitions?

wasi/src/main/java/com/dylibso/chicory/wasi/WasiPreview1.java
```diff
@@ -1124,7 +1281,12 @@ public int sockSend(int sock, int siDataPtr, int siDataLen, int siFlags, int ret
     @WasmExport
     public int sockShutdown(int sock, int how) {
         logger.infof("sock_shutdown: [%s, %s]", sock, how);
-        throw new WASMRuntimeException("We don't yet support this WASI call: sock_shutdown");
+        Descriptor descriptor = descriptors.get(sock);
+        if (descriptor == null) {
+            return wasiResult(WasiErrno.EBADF);
+        }
+        // sockets are not supported, so this cannot be a socket
+        return wasiResult(WasiErrno.ENOTSOCK);
```
andreaTP, line 1289: Does this implementation improve the execution of some programs?
If not, I would keep the Exception for now as it's a cleaner for the user what is going on.

electrum: This passes the tests and is a technically complete implementation. In practice, it shouldn't matter either way, as this won't be called as no program can obtain a valid socket. But if it is called with an invalid descriptor, the program will receive the expected error rather than crashing.

wasi/src/main/java/com/dylibso/chicory/wasi/WasiFstFlags.java
electrum: I added another commit with doc links for all the classes

wasi/src/main/java/com/dylibso/chicory/wasi/WasiPreview1.java
andreaTP: There are questions that shouldn't be asked if you are not ready for the answer, thanks for explaining!

wasi/src/main/java/com/dylibso/chicory/wasi/WasiFstFlags.java
andreaTP: Can you, please, use permalinks?

Linking to `main/blob` tend to go out of date pretty fast in my experience.

electrum: I can do that if you feel strongly, but I think it's better to link to the current documentation and update the link if it breaks, rather than linking to an outdated version (without noticing that it's old).

Given that preview1 is basically stable, I'm not sure it matters either way.

andreaTP: I do prefer, almost always, permalinks over moving tags:

- they don't get outdated 
- they show the link the author used to do the implementation, including errors etc.

That said, the comment is close to a nitpick, not going to block this great effort for it. Thanks 👍

andreaTP (approved):
Thanks for taking the time to improve this aspect! Looks awesome!

wasi/src/main/java/com/dylibso/chicory/wasi/WasiFstFlags.java
electrum: For code links, especially with line numbers, I complete agree.

Whenever I see a documentation link to a commit, my first thought is "What's this random commit? Let me go find the latest."

So that's my aversion to doing it. But I do see your point view. Ideally, they would publish the spec to a website with stable URLs.

andreaTP: following up only to challenge my stand point and because I enjoy this discussion 😊:
when you get the latest you will have a hard time mixing and matching the date on the git commit and the actual status of the repo at that point in time, to understand what the author followed.

I usually find the inverse process much easier to follow (e.g. forwarding changes to see what changed) as it gives you a stable and fixed starting point to reason about.

That said, yes, stable URLs will be best...

electrum: Alright, you convinced me. I'll update it to the latest release tag. Also, the README was pointing to `main` and the link was broken 😆

andreaTP: hahah :+1:

PR #552, opened 2024-10-02
Move static code used at runtime by aot compiled modules to a separate package

andreaTP:
Minor but "diff noisy" PR to get to #494


aot/src/test/resources/com/dylibso/chicory/approvals/approve-all.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
andreaTP, line 1: Sneaked this in, can remove if requested.

aot-runtime/src/main/java/com/dylibso/chicory/aot/runtime/UsedByGeneratedCode.java
electrum, line 1: We could drop this annotation now. Previously, we had referenced methods in multiple places, and referenced them directly in the generated bytecode. Now that we're using Java reflection `getMethod()` to reference them, IntelliJ understands that the methods are used.

It was useful when I first added it, but isn't needed now.

aot/src/main/java/com/dylibso/chicory/aot/AotEmitters.java
```diff
@@ -104,7 +73,7 @@ public static void ELEM_DROP(AotContext ctx, AnnotatedInstruction ins, MethodVis
         asm.visitVarInsn(Opcodes.ALOAD, ctx.instanceSlot());
         asm.visitLdcInsn(index);
         asm.visitInsn(Opcodes.ACONST_NULL);
-        emitInvokeVirtual(asm, INSTANCE_SET_ELEMENT);
+        emitInvokeVirtual(asm, AotMethodsRefs.INSTANCE_SET_ELEMENT);
```
electrum, line 76: Let's keep these static imported. It looks cleaner and makes the diff smaller.

aot/src/main/java/com/dylibso/chicory/aot/AotMethodsRefs.java
```diff
@@ -1,20 +1,13 @@
...
-import com.dylibso.chicory.runtime.TrapException;
-import com.dylibso.chicory.runtime.exceptions.WASMRuntimeException;
-import com.dylibso.chicory.wasm.exceptions.ChicoryException;
 import com.dylibso.chicory.wasm.types.Element;
-import com.dylibso.chicory.wasm.types.FunctionType;
 import java.lang.reflect.Method;
-import java.util.List;
 
-public final class AotMethods {
+public final class AotMethodsRefs {
```
electrum, line 10: Let's name this `AotMethodRefs` to avoid the double plural. It also directly matches the English:

* AOT methods
* AOT method references

aot/src/test/resources/com/dylibso/chicory/approvals/approve-all.sh
electrum: I like this. Is there a way to run the tests from Maven without the diffs popping up individually in IntelliJ?

electrum (approved):
Please remember to rebase as there were changes to `AotMethods` in #555

aot/src/test/resources/com/dylibso/chicory/approvals/approve-all.sh
andreaTP: Thanks for looking into it, I remove the script in this case.
But I need to add this note somewhere in the docs, otherwise I'm going to forget pretty quickly :sweat_smile:

aot/src/main/java/com/dylibso/chicory/aot/AotMethodsRefs.java
andreaTP: thanks :sweat_smile:

aot/src/main/java/com/dylibso/chicory/aot/AotEmitters.java
andreaTP: In all honesty, I personally prefer this notation, but I'm not fussy on it.

aot/src/test/resources/com/dylibso/chicory/approvals/approve-all.sh
andreaTP: > Is there a way to run the tests from Maven without the diffs popping up individually in IntelliJ?

It was happening to me on a Mac, now on Linux is I don't have this horrific behavior anymore, but I can't recall if I did something about it ...

PR #553, opened 2024-10-02
Re-enable jitpack releases

andreaTP:
It's easy enough and we can have "pre-releases" easily sharable.

Example from this branch:
https://jitpack.io/#andreaTP/chicory/re-enable-jitpack-6b3e236206-1


electrum (approved):
This is cool!

PR #555, opened 2024-10-02
Optimize AOT CALL_INDIRECT for calls in the same module

electrum:


andreaTP (approved):
Great work, easy to read and straightforward.

LGTM, thanks a lot as always!

PR #556, opened 2024-10-03
Aot pre-compiled machine

andreaTP:
This builds on top of #552 , so review only the second commit here.

Supersede and finalize the effort started in #494

Notes:
- I couldn't get rid of the generated `Machine` implementation to get the interoperability working
- this can be ported to an annotation processor if required, but there are a few limitations in reading files from the disk there

### DIFF
--- a/aot-maven-plugin/pom.xml
+++ b/aot-maven-plugin/pom.xml
@@ -1,0 +1,77 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>com.dylibso.chicory</groupId>
+    <artifactId>chicory</artifactId>
+    <version>999-SNAPSHOT</version>
+    <relativePath>../pom.xml</relativePath>
+  </parent>
+  <artifactId>aot-maven-plugin</artifactId>
+  <packaging>maven-plugin</packaging>
+  <name>Chicory - AOT Maven Plugin</name>
+  <description>A Maven Plugin to generate a package for running WASM</description>
+
+  <dependencies>
+    <dependency>
+      <groupId>com.dylibso.chicory</groupId>
+      <artifactId>aot</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.dylibso.chicory</groupId>
+      <artifactId>runtime</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.dylibso.chicory</groupId>
+      <artifactId>wasm</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.github.javaparser</groupId>
+      <artifactId>javaparser-core</artifactId>
+      <version>${javaparser.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.maven</groupId>
+      <artifactId>maven-core</artifactId>
+      <version>${maven-plugin-api.version}</version>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.maven</groupId>
+      <artifactId>maven-model</artifactId>
+      <version>${maven-plugin-api.version}</version>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.maven</groupId>
+      <artifactId>maven-plugin-api</artifactId>
+      <version>${maven-plugin-api.version}</version>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.maven.plugin-tools</groupId>
+      <artifactId>maven-plugin-annotations</artifactId>
+      <version>${maven-plugin-annotations.version}</version>
+      <scope>provided</scope>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-plugin-plugin</artifactId>
+        <configuration>
+          <goalPrefix>chicory</goalPrefix>
+        </configuration>
+        <executions>
+          <execution>
+            <id>default-descriptor</id>
+            <phase>process-classes</phase>
+          </execution>
+        </executions>
+      </plugin>
+    </plugins>
+  </build>
+</project>
--- a/aot-maven-plugin/src/main/java/com/dylibso/chicory/maven/aot/AotGenMojo.java
+++ b/aot-maven-plugin/src/main/java/com/dylibso/chicory/maven/aot/AotGenMojo.java
@@ -1,0 +1,134 @@
+package com.dylibso.chicory.maven.aot;
+
+import static com.github.javaparser.StaticJavaParser.parseClassOrInterfaceType;
+import static com.github.javaparser.StaticJavaParser.parseType;
+
+import com.dylibso.chicory.aot.AotCompiler;
+import com.dylibso.chicory.runtime.Machine;
+import com.dylibso.chicory.wasm.Parser;
+import com.github.javaparser.ast.CompilationUnit;
+import com.github.javaparser.ast.Modifier;
+import com.github.javaparser.ast.NodeList;
+import com.github.javaparser.ast.expr.NameExpr;
+import com.github.javaparser.ast.expr.ObjectCreationExpr;
+import com.github.javaparser.ast.stmt.ReturnStmt;
+import com.github.javaparser.utils.SourceRoot;
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.util.Map;
+import org.apache.maven.model.Resource;
+import org.apache.maven.plugin.AbstractMojo;
+import org.apache.maven.plugin.MojoExecutionException;
+import org.apache.maven.plugins.annotations.LifecyclePhase;
+import org.apache.maven.plugins.annotations.Mojo;
+import org.apache.maven.plugins.annotations.Parameter;
+import org.apache.maven.project.MavenProject;
+
+/**
+ * This plugin generates an invokable library from the compiled Wasm
+ */
+@Mojo(name = "wasm-aot-gen", defaultPhase = LifecyclePhase.GENERATE_SOURCES, threadSafe = true)
+public class AotGenMojo extends AbstractMojo {
+
+    /**
+     * the wasm module to be used
+     */
+    @Parameter(required = true)
+    private File wasmFile;
+
+    /**
+     * the name to be used by the generated class
+     */
+    @Parameter(required = true)
+    private String name;
+
+    /**
+     * the target folder to generate classes
+     */
+    @Parameter(
+            required = true,
+            defaultValue = "${project.basedir}/target/generated-resources/chicory-aot")
+    private File targetClassFolder;
+
+    /**
+     * the target source folder to generate the Machine implementation
+     */
+    @Parameter(
+            required = true,
+            defaultValue = "${project.basedir}/target/generated-sources/chicory-aot")
+    private File targetSourceFolder;
+
+    /**
+     * The current Maven project.
+     */
+    @Parameter(property = "project", required = true, readonly = true)
+    private MavenProject project;
+
+    @Override
+    @SuppressWarnings({"StringSplitter", "deprecation"})
+    public void execute() throws MojoExecutionException {
+        var module = Parser.parse(wasmFile);
+        var result = AotCompiler.compileModule(module, name);
+        var split = name.split("\\.");
+        var finalFolder = targetClassFolder.toPath();
+        var finalSourceFolder = targetSourceFolder.toPath();
+        for (int i = 0; i < (split.length - 1); i++) {
+            finalFolder = finalFolder.resolve(split[i]);
+            finalSourceFolder = finalSourceFolder.resolve(split[i]);
+        }
+        finalFolder.toFile().mkdirs();
+        var packageName = split[0];
+        for (int i = 1; i < (split.length - 1); i++) {
+            packageName += "." + split[i];
+        }
+
+        // Generate static Machine implementation
+        finalSourceFolder.toFile().mkdirs();
+        final SourceRoot dest = new SourceRoot(finalSourceFolder);
+
+        var machineName = split[split.length - 1] + "MachineFactory";
+
+        var cu = new CompilationUnit(packageName);
+        cu.setPackageDeclaration(packageName);
+        cu.setStorage(finalSourceFolder.resolve(machineName + ".java"));
+
+        cu.addImport("com.dylibso.chicory.runtime.Instance");
+        cu.addImport("com.dylibso.chicory.runtime.Machine");
+
+        var clazz = cu.addClass(machineName, Modifier.Keyword.PUBLIC, Modifier.Keyword.FINAL);
+
+        var constr = clazz.addConstructor(Modifier.Keyword.PRIVATE);
+        constr.createBody();
+
+        var method = clazz.addMethod("create", Modifier.Keyword.PUBLIC, Modifier.Keyword.STATIC);
+        method.addParameter(parseType("Instance"), "instance");
+        method.setType(Machine.class);
+        var methodBody = method.createBody();
+
+        var constructorInvocation =
+                new ObjectCreationExpr(
+                        null,
+                        parseClassOrInterfaceType(name),
+                        NodeList.nodeList(new NameExpr("instance")));
+        methodBody.addStatement(new ReturnStmt(constructorInvocation));
+
+        dest.add(cu);
+        dest.saveAll();
+
+        for (Map.Entry<String, byte[]> entry : result.classBytes().entrySet()) {
+            var binaryName = entry.getKey().replace('.', '/') + ".class";
+            var targetFile = targetClassFolder.toPath().resolve(binaryName);
+            try {
+                Files.write(targetFile, entry.getValue());
+            } catch (IOException e) {
+                throw new MojoExecutionException("Failed to write " + targetFile, e);
+            }
+        }
+
+        Resource resource = new Resource();
+        resource.setDirectory(targetClassFolder.getPath());
+        project.addResource(resource);
+        project.addCompileSourceRoot(targetSourceFolder.getPath());
+    }
+}
--- a/machine-tests/pom.xml
+++ b/machine-tests/pom.xml
@@ -7,32 +7,44 @@
     <artifactId>chicory</artifactId>
     <version>999-SNAPSHOT</version>
   </parent>
-  <artifactId>wabt</artifactId>
+  <artifactId>machine-tests</artifactId>
   <packaging>jar</packaging>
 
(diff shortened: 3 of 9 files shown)


wabt-aot/pom.xml
```diff
@@ -39,10 +43,37 @@
...
+        <artifactId>aot-maven-plugin</artifactId>
+        <version>${project.version}</version>
+        <executions>
+          <execution>
+            <id>wast2json</id>
+            <goals>
+              <goal>wasm-aot-gen</goal>
+            </goals>
+            <configuration>
+              <name>chicory.gen.Wast2JsonModule</name>
```
electrum, line 61: Generate into the same package as the user code
```xml
<name>com.dylibso.chicory.wabt.Wast2JsonModule</name>

File: wabt-aot/pom.xml [hunk 2]
```diff
@@ -39,10 +43,37 @@
...
+            </configuration>
+          </execution>
+          <execution>
+            <id>wat2wasm</id>
+            <goals>
+              <goal>wasm-aot-gen</goal>
+            </goals>
+            <configuration>
+              <name>chicory.gen.Wat2WasmModule</name>
+              <wasmFile>${project.basedir}/../wabt-aot/src/main/resources/wat2wasm</wasmFile>
```
electrum, line 72: Same as above

runtime/src/test/java/com/dylibso/chicory/runtime/ModuleTest.java
```diff
@@ -361,4 +362,15 @@ public void shouldConsumeStackLoopOperations() {
         // IIUC: 3 values returning from last CALL + 1 result
         assertTrue(finalStackSize.get() == 4L);
     }
+
+    @Test
+    public void shouldIgnoreMissingImports() {
+        assertDoesNotThrow(
```
electrum, line 368: This seems redundant over simply executing the code?

runtime/src/main/java/com/dylibso/chicory/runtime/Instance.java
```diff
@@ -684,10 +690,16 @@ public Instance build() {
...
             var mappedHostImports =
-                    mapHostImports(
-                            imports,
-                            (externalValues == null) ? new ExternalValues() : externalValues,
-                            module.memorySection().map(MemorySection::memoryCount).orElse(0));
+                    (skipImportMapping)
+                            ? null
+                            : mapHostImports(
+                                    imports,
+                                    (externalValues == null)
```
electrum, line 697: See if this looks nicer:
```java
requireNonNullElseGet(externalValues, ExternalValues::new)

File: runtime/src/main/java/com/dylibso/chicory/runtime/Instance.java [hunk 5]
```diff
@@ -684,10 +690,16 @@ public Instance build() {
             }
 
             var mappedHostImports =
-                    mapHostImports(
-                            imports,
-                            (externalValues == null) ? new ExternalValues() : externalValues,
-                            module.memorySection().map(MemorySection::memoryCount).orElse(0));
+                    (skipImportMapping)
```
electrum, line 693: Nit: no need for parens

pom.xml
```diff
@@ -138,6 +140,11 @@
         <artifactId>aot</artifactId>
         <version>${project.version}</version>
       </dependency>
+      <dependency>
```
electrum, line 143: Is this needed? I don't think it's ever used as a dependency

wabt-aot/pom.xml
```diff
@@ -7,13 +7,17 @@
     <artifactId>chicory</artifactId>
     <version>999-SNAPSHOT</version>
   </parent>
-  <artifactId>wabt</artifactId>
+  <artifactId>wabt-aot</artifactId>
```
electrum, line 10: Why rename the module? The AOT part seems like an implementation detail.

pom.xml
```diff
@@ -44,10 +45,11 @@
     <module>fuzz</module>
     <module>jmh</module>
     <module>log</module>
+    <module>machines-tests</module>
```
electrum, line 48: Nit: `machine-tests` sounds nicer. The double plural is a bit off.

machines-tests/src/test/java/com/dylibso/chicory/testing/MachinesTest.java
```diff
@@ -0,0 +1,218 @@
...
+
+    private Instance.Builder moduleInstanceBuilder() {
+        return Instance.builder(
+                Parser.parse(
+                        MachinesTest.class.getResourceAsStream(
+                                "/compiled/hello-world.js.javy-dynamic.wasm")));
+    }
+
+    private WasiPreview1 setupWasi(ByteArrayOutputStream stderr) {
+        ByteArrayInputStream stdin = new ByteArrayInputStream("".getBytes(UTF_8));
```
electrum, line 49: `InputStream.nullInputStream()`

machines-tests/src/test/java/com/dylibso/chicory/testing/MachinesTest.java
```diff
@@ -0,0 +1,218 @@
...
+                assertEquals(18, startFunctionIndex.get());
+
+                Instance.builder(Parser.parse(new File("../wabt-aot/src/main/resources/wat2wasm")))
+                        .withMachineFactory(
+                                (inst) ->
+                                        (funcId, args) -> {
+                                            assertEquals(startFunctionIndex.get(), funcId);
+                                            // The IDE breaks here, but this is not something the
+                                            // user is supposed to do
+                                            chicory.gen.Wat2WasmModule.func_18(inst.memory(), inst);
```
electrum, line 205: What if we do
```java
Machine machine = new Wat2WasmModuleMachine();

.withMachineFactory(inst -> (funcId, args) -> {
    assertEquals(startFunctionIndex.get(), funcId);
    return machine.call(funcId, args);
})

File: machines-tests/pom.xml [hunk 11]
```diff
@@ -0,0 +1,100 @@
...
+        <artifactId>aot-maven-plugin</artifactId>
+        <version>${project.version}</version>
+        <executions>
+          <execution>
+            <id>compile-quickjs</id>
+            <goals>
+              <goal>wasm-aot-gen</goal>
+            </goals>
+            <configuration>
+              <name>chicory.gen.QuickJSModule</name>
```
electrum, line 72:
```xml
<name>com.dylibso.chicory.testing.gen.QuickJSModule</name>
```
Same for below

aot-maven-plugin/src/main/java/com/dylibso/chicory/maven/aot/AotGenMojo.java
```diff
@@ -0,0 +1,275 @@
...
+    /**
+     * the wasm module to be used
+     */
+    @Parameter(required = true)
+    private File wasmFile;
+
+    /**
+     * the name to be used by the generated class
+     */
+    @Parameter(required = true, defaultValue = "chicory.gen.CompiledModule")
```
electrum, line 53: Let's remove the default value. Using any default value would cause duplicate class conflicts.

aot-maven-plugin/src/main/java/com/dylibso/chicory/maven/aot/AotGenMojo.java
```diff
@@ -0,0 +1,275 @@
...
+import org.apache.maven.plugins.annotations.LifecyclePhase;
+import org.apache.maven.plugins.annotations.Mojo;
+import org.apache.maven.plugins.annotations.Parameter;
+import org.apache.maven.project.MavenProject;
+
+/**
+ * This plugin generates an invokable library from the compiled Wasm
+ */
+@Mojo(name = "wasm-aot-gen", defaultPhase = LifecyclePhase.GENERATE_RESOURCES, threadSafe = true)
+public class AotGenMojo extends AbstractMojo {
```
electrum, line 42: Not required, but it would be nice to have a test against an example generated class source, so that we can see what the generated code looks like.

aot-maven-plugin/src/main/java/com/dylibso/chicory/maven/aot/AotGenMojo.java
```diff
@@ -0,0 +1,275 @@
...
+import org.apache.maven.plugin.MojoExecutionException;
+import org.apache.maven.plugins.annotations.LifecyclePhase;
+import org.apache.maven.plugins.annotations.Mojo;
+import org.apache.maven.plugins.annotations.Parameter;
+import org.apache.maven.project.MavenProject;
+
+/**
+ * This plugin generates an invokable library from the compiled Wasm
+ */
+@Mojo(name = "wasm-aot-gen", defaultPhase = LifecyclePhase.GENERATE_RESOURCES, threadSafe = true)
```
electrum, line 41: `GENERATE_SOURCES` is probably more appropriate, since we're generating sources and that phase runs earlier

electrum (approved):
This looks great!

electrum:
Let me see if I can change the generated module so that we don't need to generate the machine.

electrum:
It's mostly done, but I'll need to finish it tomorrow: https://github.com/electrum/chicory/commits/aot-machine

andreaTP:
> Let me see if I can change the generated module so that we don't need to generate the machine.

That would be really cool, we still need to generate a wrapper(or something) usable by the user as directly referencing the generated `.class` horribly breaks the IDE.

wabt-aot/src/main/java/com/dylibso/chicory/aot/wabt/Files.java
```diff
@@ -1,4 +1,4 @@
-package com.dylibso.chicory.wabt;
+package com.dylibso.chicory.aot.wabt;
```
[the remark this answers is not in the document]
andreaTP: agreed, I changed the package name during development to be able to compare stuff and make sure I was using the correct thing.

pom.xml
andreaTP: right, moved to `pluginManagement`

pom.xml
andreaTP: I couldn't decide for an appropriate name for this module testing the interoperability of the implementations.
Let's go with `machine-tests` for now :+1:

aot-maven-plugin/src/main/java/com/dylibso/chicory/maven/aot/AotGenMojo.java
andreaTP: Do you mean a test "in isolation" other than using `wabt` and `machine-tests`?

aot-maven-plugin/src/main/java/com/dylibso/chicory/maven/aot/AotGenMojo.java
andreaTP: ok, agreed, thinking about this line twice, I'm not sure if this is `thread-safe`, I change it to `false` to stay on the safe side for now.

andreaTP:
Things are coming together pretty nicely!
Thanks a lot @electrum for supporting the effort.

aot-maven-plugin/src/main/java/com/dylibso/chicory/maven/aot/AotGenMojo.java
electrum: Yeah, I'm thinking something similar to the annotation processor tests that have the expected generated Java files, or the approval tests for AOT with the expected bytecode. But hopefully with the AOT changes we don't need to generate any code now?

aot-maven-plugin/src/main/java/com/dylibso/chicory/maven/aot/AotGenMojo.java
electrum: It should be thread safe. According to [this post](https://stackoverflow.com/questions/50247168/how-to-make-sure-that-a-maven-plugin-is-thread-safe):

> a mojo is considered `threadSafe` as long as there's no mutable state that is *shared across different instances* of the mojo -- e.g. a static reference to an object with mutable contents

The expectation now for Maven is that every Mojo should be thread safe. You have to violate general Java best practices to make it not thread safe.

machine-tests/pom.xml
```diff
@@ -0,0 +1,100 @@
...
+  </parent>
+  <artifactId>machine-tests</artifactId>
+  <packaging>jar</packaging>
+
+  <name>Chicory - Machines tests</name>
+
+  <dependencies>
+    <dependency>
+      <groupId>com.dylibso.chicory</groupId>
+      <artifactId>aot-runtime</artifactId>
```
electrum, line 18: We missed this earlier, but `aot-runtime` should be added to the BOM, as user projects will transitively depend on this and we want to ensure the version is consistent with the other Chicory libraries.

machine-tests/src/test/java/com/dylibso/chicory/testing/MachinesTest.java
```diff
@@ -0,0 +1,217 @@
...
+        assertEquals(expectedOutput, stderr.toString(UTF_8));
+
+        // and now the interpreter
+        moduleInstanceBuilder().withExternalValues(store.toExternalValues()).build();
+
+        assertEquals(expectedOutput + expectedOutput, stderr.toString(UTF_8));
+    }
+
+    @Test
+    public void shouldUseMachineCallOnlyForExport() throws Exception {
```
electrum, line 162: Should we move this test to the AOT module? It's not dependent on precompilation

aot-maven-plugin/pom.xml
```diff
@@ -0,0 +1,77 @@
...
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-plugin-plugin</artifactId>
+        <configuration>
+          <goalPrefix>prefix</goalPrefix>
```
electrum, line 66: Should this be `chicory`?

aot-maven-plugin/src/main/java/com/dylibso/chicory/maven/aot/AotGenMojo.java
```diff
@@ -0,0 +1,137 @@
...
+    @SuppressWarnings({"StringSplitter", "deprecation"})
+    public void execute() throws MojoExecutionException {
+        var instance =
+                Instance.builder(Parser.parse(wasmFile))
+                        .withMachineFactory(inst -> new AotMachine(name, inst))
+                        .withStart(false)
+                        .withInitialize(false)
+                        .withSkipImportMapping(true)
+                        .build();
+        var compiled = ((AotMachine) instance.getMachine()).compiledClass();
```
electrum, line 78: With #565, this can become
```java
var module = Parser.parse(wasmFile);
var compiler = new AotCompiler(module, name);
var compiled = compiler.compileClassToBytes();
// verify that compiled class is valid
compiler.loadClassBytes(compiled);

File: runtime/src/main/java/com/dylibso/chicory/runtime/Instance.java [hunk 20]
```diff
@@ -385,6 +387,11 @@ public Builder withStart(boolean s) {
             return this;
         }
 
+        public Builder withSkipImportMapping(boolean s) {
```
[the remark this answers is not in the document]
andreaTP: I like this approach and I'll follow up, let me have a little more slack time :sweat_smile:

wabt/pom.xml
```diff
@@ -11,9 +11,13 @@
   <packaging>jar</packaging>
 
   <name>Chicory - wabt</name>
-  <description>wabt tools running in pure Java with Chicory</description>
+  <description>wabt tools compiled to pure Java with Chicory</description>
 
   <dependencies>
+    <dependency>
+      <groupId>com.dylibso.chicory</groupId>
+      <artifactId>aot-runtime</artifactId>
```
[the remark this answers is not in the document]
andreaTP: My understanding here is that we want to add to the `bom` only modules that are actually going to make the "supported" `1.0`.
The `aot` is still out of this category imho(needs a bit more investment), I aim for it to be next, and I'm happy to help whenever possible, but I'm a bit on the fence of adding it to the `bom` now as it might create misaligned expectations.

machine-tests/src/test/java/com/dylibso/chicory/testing/MachinesTest.java
andreaTP: good suggestion :+1: it's just a sanity check from the prev implementation of `call_indirect` at the end of the day.

aot-maven-plugin/src/main/java/com/dylibso/chicory/maven/aot/AotGenMojo.java
andreaTP: Sounds like an ideal API, just give me a little to review 565 :+1: thanks a lot for the support!

wabt/pom.xml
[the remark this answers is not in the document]
electrum: With the latest changes in #565, `aot-runtime` is no longer needed at all!

aot-maven-plugin/src/main/java/com/dylibso/chicory/maven/aot/AotGenMojo.java
electrum: I updated the API to allow returning multiple classes, which is needed for inlining `AotMethods`, method splitting, etc. With that change, this code can become something like
```java
for (Entry<String, byte[]> entry : result.classBytes()) {
    var targetFile = targetClassFolder.toPath().resolve(entry.getKey() + ".class");
    try {
        Files.write(targetFile, entry.getValue());
    } catch (IOException e) {
        throw new MojoExecutionException("Failed to write " + targetFile, e);
    }
}

File: aot-maven-plugin/src/main/java/com/dylibso/chicory/maven/aot/AotGenMojo.java (hunk 13 above)
Comment (andreaTP) [reply aot-maven-plugin/src/main/java/com/dylibso/chicory/maven/aot/AotGenMojo.java:32]: Understood, let's keep it for a follow up if you don't mind.
To avoid me forgetting about it, feel free to open an issue, and I'll eventually get there.

File: aot-maven-plugin/src/main/java/com/dylibso/chicory/maven/aot/AotGenMojo.java (hunk 14 above)
Comment (andreaTP) [reply aot-maven-plugin/src/main/java/com/dylibso/chicory/maven/aot/AotGenMojo.java]: got it, done, thanks for looking it up :+1:

File: machine-tests/src/test/java/com/dylibso/chicory/testing/MachinesTest.java (hunk 17 above)
Comment (andreaTP) [reply machine-tests/src/test/java/com/dylibso/chicory/testing/MachinesTest.java:164]: Turns out that is not dependent on precompilation, but it uses `wabt` and moving it to `aot` introduces a circular dependency.
Let's keep it as a follow up.

File: aot-maven-plugin/src/main/java/com/dylibso/chicory/maven/aot/AotGenMojo.java [hunk 22]
```diff
@@ -0,0 +1,134 @@
...
+
+    @Override
+    @SuppressWarnings({"StringSplitter", "deprecation"})
+    public void execute() throws MojoExecutionException {
+        var module = Parser.parse(wasmFile);
+        var result = AotCompiler.compileModule(module, name);
+        var split = name.split("\\.");
+        var finalFolder = targetClassFolder.toPath();
+        var finalSourceFolder = targetSourceFolder.toPath();
+        for (int i = 0; i < (split.length - 1); i++) {
```
andreaTP, line 76: this code is a little messy, but should be easy to improve over it.

andreaTP (commented):
I updated everything on top of the great latest improvement done by @electrum .
I go ahead and merge this, we can fix on top any leftover.

PR #558, opened 2024-10-03
Move read bytes utilities to Encoding

### LINKED ISSUE #521: Clean Parser API
The current `Parser` API consist of the expected `parse` methods and some `read...` methods.
I think the `read...` methods should not be public (at least not available from the `Parser` class).

andreaTP:
Fix #521

### DIFF
--- a/wasm/src/main/java/com/dylibso/chicory/wasm/Encoding.java
+++ b/wasm/src/main/java/com/dylibso/chicory/wasm/Encoding.java
@@ -2,6 +2,7 @@ package com.dylibso.chicory.wasm;
 
 import com.dylibso.chicory.wasm.exceptions.MalformedException;
 import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
 
 public final class Encoding {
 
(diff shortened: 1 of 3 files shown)


wasm/src/main/java/com/dylibso/chicory/wasm/Encoding.java
```diff
@@ -11,6 +13,126 @@ public final class Encoding {
...
+    public static String readName(ByteBuffer buffer) {
+        return readName(buffer, true);
+    }
+
+    public static String readName(ByteBuffer buffer, boolean checkMalformed) {
+        var length = (int) readVarUInt32(buffer);
+        byte[] bytes = new byte[length];
+        try {
+            readBytes(buffer, bytes);
+        } catch (BufferUnderflowException e) {
```
electrum, line 122: This catch shouldn't be needed as `readBytes` already checks the length.

Or we could replace the checks in those `read` methods with handling for `BufferUnderflowException`.

wasm/src/main/java/com/dylibso/chicory/wasm/Encoding.java
```diff
@@ -11,6 +13,126 @@ public final class Encoding {
...
+     * See <a href="https://www.w3.org/TR/wasm-core-1/#names%E2%91%A0">2.2.4. Names</a> of the WebAssembly Core Specification.
+     *
+     * @param buffer the byte buffer
+     * @return the resulting long
+     */
+    public static String readName(ByteBuffer buffer) {
+        return readName(buffer, true);
+    }
+
+    public static String readName(ByteBuffer buffer, boolean checkMalformed) {
```
electrum, line 117: Nit: copy Javadoc from above or make it `@see` the above

wasm/src/main/java/com/dylibso/chicory/wasm/Encoding.java
```diff
@@ -11,6 +13,126 @@ public final class Encoding {
...
+     * @param buffer the byte buffer
+     * @return the resulting long
+     */
+    public static long readFloat32(ByteBuffer buffer) {
+        return readInt(buffer);
+    }
+
+    /**
+     * Read a symbol name from the buffer as UTF-8 String.
+     * See <a href="https://www.w3.org/TR/wasm-core-1/#names%E2%91%A0">2.2.4. Names</a> of the WebAssembly Core Specification.
```
electrum, line 108: The link is broken: the extra `%E2%91%A0` at the end of the URL shouldn't be there.

wasm/src/main/java/com/dylibso/chicory/wasm/Encoding.java
```diff
@@ -11,6 +13,126 @@ public final class Encoding {
 
     private Encoding() {}
 
+    static int readInt(ByteBuffer buffer) {
```
electrum, line 16: Do these need to be public? Could they be needed for parsing custom sections?

This is just a question to consider. It's fine to leave them private for now.

electrum (approved):

wasm/src/main/java/com/dylibso/chicory/wasm/Encoding.java
andreaTP: removed

wasm/src/main/java/com/dylibso/chicory/wasm/Encoding.java
andreaTP: done

wasm/src/main/java/com/dylibso/chicory/wasm/Encoding.java
andreaTP: good catch, fixed!

wasm/src/main/java/com/dylibso/chicory/wasm/Encoding.java
andreaTP: In the current status I think that there are quite a few changes necessary to offer a proper, usable and generic API to parse custom sections ...
I would not try to predict the future, and welcome effort in this direction when someone needs it.

Opening visibility is not going to be a breaking change.

PR #561, opened 2024-10-04
Make AOT compiled classes implement Machine

electrum:
The first two commits are from #559


andreaTP:
Well, what should I say:
Screenshot from 2024-10-04 13-29-45
:rocket: :rocket: :rocket:

aot-runtime/src/main/java/com/dylibso/chicory/aot/runtime/AotMethods.java
```diff
@@ -10,34 +9,14 @@
...
 import com.dylibso.chicory.wasm.exceptions.ChicoryException;
+import com.dylibso.chicory.wasm.exceptions.InvalidException;
 import com.dylibso.chicory.wasm.types.FunctionType;
 import java.util.List;
 
 public final class AotMethods {
 
     private AotMethods() {}
 
-    public static long[] callIndirect(
```
andreaTP, line 20: ouch, this was probably a leftover due to a rebase, sorry :sweat_smile:

aot/src/main/java/com/dylibso/chicory/aot/AotUtil.java
```diff
@@ -29,38 +27,19 @@ public enum StackSize {
...
-            F32_TO_LONG = ValueConversions.class.getMethod("f32ToLong", float.class);
-            F64_TO_LONG = ValueConversions.class.getMethod("f64ToLong", double.class);
-
-            LONG_TO_I32_MH = publicLookup().unreflect(LONG_TO_I32);
-            LONG_TO_F32_MH = publicLookup().unreflect(LONG_TO_F32);
-            LONG_TO_F64_MH = publicLookup().unreflect(LONG_TO_F64);
-            I32_TO_LONG_MH = publicLookup().unreflect(I32_TO_LONG);
-            F32_TO_LONG_MH = publicLookup().unreflect(F32_TO_LONG);
-            F64_TO_LONG_MH = publicLookup().unreflect(F64_TO_LONG);
+            LONG_TO_F32 = Value.class.getMethod("longToFloat", long.class);
```
andreaTP, line 37: Nice cleanup, thanks :+1:

andreaTP (approved):
This change is increasing a bit the complexity, but is probably pretty much "complete".
Amazing work @electrum :+1: thanks!

electrum:
It does add complexity, but also removes the `MethodHandle` adaptation code, which wasn't easy to understand (or write). But the primary motivation is for your pre-compiled AOT work. Now we don't need the Java code generator and it uses the same code in both modes.

andreaTP:
Agreed, thanks a lot for the help.

Much appreciated, and the result is outstanding!

Next is going to be `Instance.initialize` fully embedded in AOT, but let's settle things a bit first 😅

PR #562, opened 2024-10-04
Add a note on the env var to be used with Approval tests

andreaTP:
Follow up from:
https://github.com/dylibso/chicory/pull/552#discussion_r1787241933

I haven't found a better place to drop this note ...


bhelx (approved):

electrum (approved):
Nice! You could add this into the `ApprovalTest.java` file as well.

PR #563, opened 2024-10-04
Down-level logging in the WasiPreview1

### LINKED ISSUE #534: Down-level logging in the WasiPreview1 module
https://github.com/dylibso/chicory/blob/b69237737c6f05f74d4c3fafbb1e32a350953e9a/wasi/src/main/java/com/dylibso/chicory/wasi/WasiPreview1.java#L161

I think it would be appropriate to set the log level on all of these to trace or at most debug. It can be quite noisy for the average user.

andreaTP:
Fix #534

### DIFF
--- a/wasi/src/main/java/com/dylibso/chicory/wasi/WasiPreview1.java
+++ b/wasi/src/main/java/com/dylibso/chicory/wasi/WasiPreview1.java
@@ -139,19 +139,19 @@ public final class WasiPreview1 implements Closeable {
 
     @WasmExport
     public int adapterCloseBadfd(int fd) {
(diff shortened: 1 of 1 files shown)


electrum (approved):
As a follow up, we could generate these automatically in the annotation processor. I’d like to have output like [`strace`](https://strace.io/) where it prints the call and result all on one line. That would be ugly to do in the code, but easy with a generated wrapper. It needs try/catch so that we can log the call without the result when an exception is thrown.

andreaTP:
@electrum agreed, let me work a little more on the annotation processors before proposing this change if you don't mind :pray:

electrum:
Sure, I'm in no hurry to work on this, currently focused on AOT.

PR #564, opened 2024-10-04
Remove function-annotations from the BOM

andreaTP:
We need to work a bit more on the annotation processing story, I noticed that we have this in the BOM and I think that we should leave it out to avoid setting wrong expectations when we release the next version.

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -16,11 +16,6 @@
     <dependencies>
       <dependency>
         <groupId>com.dylibso.chicory</groupId>
-        <artifactId>function-annotations</artifactId>
-        <version>${project.version}</version>
-      </dependency>
-      <dependency>
-        <groupId>com.dylibso.chicory</groupId>
         <artifactId>log</artifactId>
         <version>${project.version}</version>
       </dependency>


electrum:
These are already used by WASI successfully for implementing host functions. Shouldn’t we make it easy for people to try out to get more feedback?

electrum:
Maybe best to discuss this in the meeting next week?

andreaTP:
> Maybe best to discuss this in the meeting next week?

I might be missing this one, the idea behind this removal was to release those annotations as experimental for the 1.0 release.
I have no problem in closing this PR when we have a better defined "release plan" though, it's mostly a call to "let's think about this" than anything.

electrum:
Having it in the BOM makes it easier to use. We can still mark them as experimental. Unless we split it out into a separate project, the artifact would still be versioned as `1.0`, so I don't see how removing it from the BOM changes anything, other than making it more difficult to use.

andreaTP:
They'd likely be renamed etc., but your reasoning is fine for me, we have to decide how to mark experimental modules though ...

PR #570, opened 2024-10-09
corpus(tinygo): add panic reproducer for tinygo with dependency

evacchi:
While testing the Chicory SDK I found an issue with initializing a TinyGo module with a dependency. 
The issue is in the initializer, and it looks like it's an index out of bound (I tested with other runtimes and they initialize successfully).

Because this relies on an external dependency, I modified `wasm-corpus` so that the TinyGo files are organized as modules.

PR in draft, since this is only the reproducer, without the solution.


evacchi:
interesting, so compiling with `-gc=leaking` and adding a small tracing breakpoint logging the function name, we get:

```
_start
runtime.alloc
runtime.alloc
runtime.runqueuePushBack
tinygo_launch
runtime.run$1$gowrapper
runtime.run$1
__wasm_call_ctors
__wasilibc_populate_preopens
__wasi_fd_prestat_get
__imported_wasi_snapshot_preview1_fd_prestat_get
runtime.args_sizes_get
time.now
runtime.clock_time_get
errors.New
runtime.alloc
errors.New
runtime.alloc
errors.New
runtime.alloc
runtime.lookupPanic
runtime.runtimePanicAt
runtime.printstring
runtime.putchar
runtime.putchar
runtime.putchar
runtime.putchar
runtime.putchar
runtime.putchar
runtime.putchar
runtime.putchar
runtime.putchar
runtime.putchar
runtime.putchar
runtime.putchar
runtime.putchar
runtime.putchar
runtime.putchar
runtime.putchar
runtime.putchar
runtime.putchar
runtime.putchar
runtime.putchar
runtime.putchar
runtime.putchar
runtime.printstring
runtime.putchar
runtime.putchar
runtime.putchar
runtime.putchar
runtime.putchar
runtime.putchar
runtime.putchar
runtime.putchar
runtime.putchar
runtime.putchar
runtime.putchar
runtime.putchar
runtime.putchar
runtime.putchar
runtime.putchar
runtime.putchar
runtime.putchar
runtime.putchar
runtime.printnl
runtime.putchar
runtime.fd_write
```

evacchi:
oughh. So, the issue was that the default WasiOptions.Arguments is an empty list, but the C/UNIX convention is that `args[0]` should be defined as the executable. So, if you _don't_ check the args you are fine, but if anyone by chance does 💥. I suppose that my import by transitive import ends up calling some initializer that peeks at the args.

Ok technically this is not a bug in Chicory, so I suppose we might close it.

PR #571, opened 2024-10-11
Implement WASI fd_pread and fd_pwrite

electrum:


wasi/src/test/java/com/dylibso/chicory/wasi/WasiPreview1Test.java
```diff
@@ -226,4 +230,27 @@ public void wasiRandom() {
...
+        var exitCode = 0;
+        Optional<String> stderr = Optional.empty();
+        Optional<String> stdout = Optional.empty();
+        Executable executable =
+                () -> WasiTestRunner.execute(test, args, dirs, env, exitCode, stderr, stdout);
+
+        var exception = assertThrows(RuntimeException.class, executable);
+        assertEquals(
+                "Failed to execute test: "
+                        + "../wasi-testsuite/tests/c/testsuite/pwrite-with-append.wasm\n"
```
andreaTP, line 249: The CI is failing on Win because of the path separator, i.e. `..\wasi-testsuite\tests\c\testsuite\pwrite-with-append.wasm`.

That said, I'm puzzled, the test is passing here: https://github.com/dylibso/chicory/pull/571/files#diff-291aacbde1fda307c1231bc99f7e351d279573c7f250ce5ae98f4087ad3ce384R79

I assume you are testing this note:
```
Not supported for files opened in append mode.
```
but it's not clear to me how this happens, would you mind adding a comment or something that better clarifies the difference in the setup?

electrum: That test is still excluded. According to POSIX, for files opened in append mode, `pwrite` should ignore append and perform a positional write. On Linux, it ignores the position and treats it the same as a regular appending write. Jimfs emulates this behavior.

I’ll add a comment to clarify that the line of the assertion failure is after the write call, which shows that the write failed.

Or perhaps I could just test the WASI code directly?

andreaTP:
@electrum I rebased your branch and added a commit on top to have the failure kind of "unit tested".
It's a much more "white box" test than I usually like, but, in this case, the integration test seemed a bit artificial too ...

Let me know if you agree with this way forward!

wasi/src/test/java/com/dylibso/chicory/wasi/WasiPreview1Test.java
```diff
@@ -232,25 +235,25 @@ public void wasiRandom() {
...
+
+        var options = WasiOptions.builder();
+
+        var dir = "fs-tests.dir";
+        Path source = new File("../wasi-testsuite/tests/c/testsuite").toPath().resolve(dir);
+        Path target = fs.getPath(dir);
+        copyDirectory(source, target);
+        options.withDirectory(target.toString(), target);
+
+        var wasip1 = WasiPreview1.builder().withOpts(options.build()).build();
```
electrum, line 251: Nit: could name this `wasi`

wasi/src/test/java/com/dylibso/chicory/wasi/WasiPreview1Test.java
```diff
@@ -232,25 +235,25 @@ public void wasiRandom() {
...
+
+        var memory = new Memory(new MemoryLimits(1));
+        wasip1.pathOpen(memory, 3, 1, "pwrite.cleanup", 1, 0, 0, 1, 0);
+        var result = wasip1.fdPwrite(memory, 4, 0, 0, 0, 0);
```
electrum, line 255: Let's change to
```java
int fd = wasi.pathOpen(memory, 3, WasiLookupFlags.SYMLINK_FOLLOW, "pwrite.cleanup", WasiOpenFlags.CREAT, 0, 0, WasiFdFlags.APPEND, 0);
int result = wasi.fdPwrite(memory, fd, 0, 0, 0, 0);
```

wasi/src/test/java/com/dylibso/chicory/wasi/WasiPreview1Test.java
```diff
@@ -232,25 +235,25 @@ public void wasiRandom() {
...
-        assertEquals(
-                "Failed to execute test: "
-                        + "../wasi-testsuite/tests/c/testsuite/pwrite-with-append.wasm\n"
-                        + "<<<<<\n"
-                        + "Assertion failed: size == sizeof(buf) "
-                        + "(testsuite/pwrite-with-append.c: main: 22)\n"
-                        + ">>>>>",
-                exception.getMessage());
+    public void wasiPositionedWriteWithAppendShouldFail() throws IOException {
+        FileSystem fs =
```
electrum, line 239: Nit: this should be in try-with-resources

electrum (commented):
Thanks for updating the test. We should squash it into the commit that adds the pread/pwrite methods.

electrum:
@andreaTP Thanks for updating the test. I cleaned it up a bit and squashed it into the commit that adds the methods. I also rebased the branch and added another commit to cleanup the `withOptions()` builder method name.

wasi/src/test/java/com/dylibso/chicory/wasi/WasiPreview1Test.java
```diff
@@ -226,4 +233,43 @@ public void wasiRandom() {
...
+                                "pwrite.cleanup",
+                                WasiOpenFlags.CREAT,
+                                0,
+                                0,
+                                WasiFdFlags.APPEND,
+                                fdPtr);
+                assertEquals(WasiErrno.ESUCCESS.value(), result);
+
+                int fd = memory.readInt(fdPtr);
+                assertEquals(4, fd);
```
andreaTP, line 268: I'm going to forget what is this 4 in a couple of hours :sweat_smile: 
But it's an improvement over my version of it :+1:

andreaTP (approved):
LGTM, good to go!
Thanks!

wasi/src/test/java/com/dylibso/chicory/wasi/WasiPreview1Test.java
electrum: It's the next open descriptor (0-2 are stdin/out/err, 3 is the pre-opened directory). We don't actually need to assert this, but doesn't hurt to validate anyway.

andreaTP: Yeah, I know because I wrote the test with the debugger on my side :sweat_smile: but I'm going to forget about it quickly.

PR #572, opened 2024-10-11
Pinning github actions

mfirry:
Specifically

`peter-evans/create-pull-request` and `EnricoMi/publish-unit-test-result-action`


andreaTP (approved):
Thanks, LGTM!
It would be perfect if you add the reference version in a comment, but is likely to be fixed automatically by dependabot.

PR #574, opened 2024-10-12
aot suggested fixes

andreaTP:
At Devoxx I had the pleasure to meet @forax , he is one of the maintainer of the ASM library we are using in the AOT compiler and suggested those fixes.

### DIFF
--- a/aot/src/main/java/com/dylibso/chicory/aot/AotCompiler.java
+++ b/aot/src/main/java/com/dylibso/chicory/aot/AotCompiler.java
@@ -209,7 +209,7 @@ public final class AotCompiler {
 
         classWriter.visit(
                 Opcodes.V11,
-                Opcodes.ACC_PUBLIC | Opcodes.ACC_FINAL,
+                Opcodes.ACC_PUBLIC | Opcodes.ACC_FINAL | Opcodes.ACC_SUPER,
                 internalClassName,
                 null,
                 getInternalName(Object.class),


aot/src/main/java/com/dylibso/chicory/aot/AotMachine.java
```diff
@@ -367,13 +367,13 @@ private byte[] compileClass(String className, FunctionSection functions) {
         var classWriter = new ClassWriter(ClassWriter.COMPUTE_FRAMES);
         classWriter.visit(
                 Opcodes.V11,
-                Opcodes.ACC_PUBLIC | Opcodes.ACC_FINAL,
+                Opcodes.ACC_PUBLIC | Opcodes.ACC_FINAL | Opcodes.ACC_SUPER,
```
andreaTP, line 370: Apparently, compatibility with project Leyden.

aot/src/main/java/com/dylibso/chicory/aot/AotMachine.java
```diff
@@ -367,13 +367,13 @@ private byte[] compileClass(String className, FunctionSection functions) {
...
                 internalClassName,
                 null,
                 getInternalName(Object.class),
                 new String[] {getInternalName(Machine.class)});
 
-        classWriter.visitSource("wasm", "wasm");
+        classWriter.visitSource(null, null);
```
andreaTP, line 376: Better to leave those fields empty than using dummy values.

aot/src/main/java/com/dylibso/chicory/aot/AotMachine.java
forax: Not Leyden, Valhalla. In Valhalla, if you have a class without ACC_SUPER (renamed ACC_IDENTITY), you are creating a value class. See https://bugs.openjdk.org/browse/JDK-8317278

andreaTP: Many thanks for the explanation!

electrum: I'm guessing this doesn't matter was we're currently generating `V11` classes, but good to fix for the future!

electrum (approved):

aot/src/main/java/com/dylibso/chicory/aot/AotMachine.java
electrum: The second argument seems wrong for sure. The first one causes stack traces to include `wasm`, which seems useful, but I don't know if there are downsides. If @forax says we should remove it, then let's remove it!

electrum, line 376: We can simply remove this call, as calling with null won't do anything.

forax: As electrum said, for me, it should be
```java
classWriter.visitSource("wasm", null);
```
so you get "wasm" printed in the stacktraces.

andreaTP: Ok, I have nothing against, my reasoning was:
"better to not provide anything than a wrong `fileName` that the IDE will try to resolve".

But the IDE is still broken in either cases, so I just opened #576 to track it as a separate issue.

Thanks everyone for the feedback, much appreciated!

PR #578, opened 2024-10-16
Better split and structure the annotation processing stuff

andreaTP:
Moving the annotation processing stuff under a `host-module` sub-module.
Including the renames etc.


andreaTP:
I'm not attached to the package name, I'll happily apply any proposed alternative :slightly_smiling_face:

pom.xml
```diff
@@ -186,6 +186,25 @@
         <version>${project.version}</version>
       </dependency>
 
+      <!-- Jackson -->
```
electrum, line 189: Thanks for moving the dependency versions here 👍

electrum (approved):
I also dislike the naming. Using a single lowercase word `hostmodule` seems to match common practice. This is really just a compound word, not a hierarchy.

What do you expect to call the WASM side of this? `wasm-module`?

andreaTP:
> What do you expect to call the WASM side of this? `wasm-module`?

I have an ongoing discussion with @evacchi on it, and progress on it already got stuck on naming a few times :sweat_smile:

PR #581, opened 2024-10-16
Force docusaurus project to use npm

bhelx:
* We were missing package-lock.json which i think cloudflare checks the presence of to decide to use npm over yarn
* Added yarn.lock to .gitignore. I think trying to support yarn will just bring us heartache. we use this strategy in our other docusaurus deployments.
* Updated the CI to use npm over yarn


andreaTP (approved):
LGTM, we can try and see if this setup works for us.

PR #583, opened 2024-10-18
More initial docs

andreaTP:
This PR improve the structure of the docs and bring all the current content there.
It's not in great shape, but I moved all of it and it's a starting point to have more people contributing to it.


bhelx:
Unclear why cloudflare is not dropping the link here. You can view these feature branches by name though: https://more-initial-docs.chicory.pages.dev/

evacchi (approved):

evacchi:
I am approving, then I'll do another round of review over the text

PR #589, opened 2024-10-22
Rename `ExternalValues` to `ImportValues`; order of params in `ImportFunction`,`HostFunction`.

evacchi:
Follow up to https://github.com/dylibso/chicory/pull/527

Hopefully last big rename (related to this, at least) before the RC 😅 

After chatting with @andreaTP, we agreed that `ExternalValues`, from the perspective of an end user, was confusing naming ("external to what?"). We came up with `ImportedValues`. 

To be honest, my original objection with `HostImports` was about them being `Host`, not `Imports" :) 

- In all our usages, these values are _indeed_ used _by an instance_. 
- They represent a _named value_ assigned to an _import_ (i.e. a  named pair `<module,name>`). 
- The name is generic, it does not imply that value belongs to a Host-thing, nor a Wasm-thing, thus it can be used in both cases.
- There is a specular opposite, at least for functions (`ExportFunction`)

we do "lose" the parallel with the spec, but this is less of an issue, because we only use these `ImportValues` as, duh, values imported by `Instance`s. We never return `ImportValue` from an `export*()`-kind API, so everything should work

### Order of Parameters for External/HostFunction

In this PR I am also changing, again, the order of the parameters for `ImportFunction` because the old constructor "broke" the signature into two pieces, i.e. the handle was _between_ the names and the types:

- module, name, **handle**, paramTypes, returnTypes

vs:

- module, name, paramTypes, returnTypes, **handle**

which I think is clearer. This change is in the last 2 commits, for easier review.

### DIFF
--- a/aot-tests/src/test/java/com/dylibso/chicory/testing/Spectest.java
+++ b/aot-tests/src/test/java/com/dylibso/chicory/testing/Spectest.java
@@ -26,35 +26,35 @@ public final class Spectest {
     public static ImportValues toImportValues() {
         return new ImportValues(
                 new HostFunction[] {
-                    new HostFunction("spectest", "print", noop, List.of(), List.of()),
+                    new HostFunction("spectest", "print", List.of(), List.of(), noop),
                     new HostFunction(
-                            "spectest", "print_i32", noop, List.of(ValueType.I32), List.of()),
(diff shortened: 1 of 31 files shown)


andreaTP (approved):
LGTM! Thanks!

PR #590, opened 2024-10-22
docs: some cleanup, rewording

evacchi:
Reworded a few sections, this will need changes if #589 gets merged. Notably:

- The `HostFunction` section now comes first, so I removed the reference to "using _again_ memory" because this is the first time we are seeing it
- I have also rounded up the section to introduce and clarify the concept of "host", "guest"
- [x] the description of the `HostFunction` should change to reflect #589
- [x] we should also update this same section to use the `Store` from the start
- [x] we might describe later the alternate method (passing in "imports"/"external values" directly) for low-level access

EDIT: I did all the last 3 above as #589 was merged in the meantime


andreaTP (approved):
LGTM

PR #593, opened 2024-10-22
Review runtime checks

### LINKED ISSUE #422: Review Machine runtime checks

andreaTP:
Fix #422

### DIFF
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/InterpreterMachine.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/InterpreterMachine.java
@@ -269,7 +269,7 @@ class InterpreterMachine implements Machine {
                     MEMORY_GROW(stack, instance);
                     break;
                 case MEMORY_FILL:
-                    MEMORY_FILL(stack, instance, operands);
+                    MEMORY_FILL(stack, instance);
                     break;
                 case I32_STORE8:
                 case I64_STORE8:
@@ -718,7 +718,7 @@ class InterpreterMachine implements Machine {
                     DATA_DROP(instance, operands);
                     break;
                 case MEMORY_COPY:
-                    MEMORY_COPY(stack, instance, operands);
+                    MEMORY_COPY(stack, instance);
                     break;
                 case TABLE_COPY:
                     TABLE_COPY(stack, instance, operands);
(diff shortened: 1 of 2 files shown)


runtime/src/main/java/com/dylibso/chicory/runtime/InterpreterMachine.java
```diff
@@ -1696,14 +1682,12 @@ private static void I64_LOAD32_U(MStack stack, Instance instance, Operands opera
     private static void I64_LOAD32_S(MStack stack, Instance instance, Operands operands) {
         var ptr = readMemPtr(stack, operands);
         var val = instance.memory().readI32(ptr);
-        // TODO this is a bit hacky
```
evacchi, line 1699: curious: what was this about? 😬

runtime/src/main/java/com/dylibso/chicory/runtime/InterpreterMachine.java
```diff
@@ -1630,11 +1620,7 @@ private static void F32_NEG(MStack stack) {
         stack.push(Value.floatToLong(-tos));
     }
 
-    private static void MEMORY_FILL(MStack stack, Instance instance, Operands operands) {
-        var memidx = (int) operands.get(0);
-        if (memidx != 0) {
-            throw new WASMRuntimeException("We don't support multiple memories just yet");
-        }
```
evacchi, line 1637: is this handled somewhere else now? e.g. validation?

runtime/src/main/java/com/dylibso/chicory/runtime/InterpreterMachine.java
andreaTP: I don't remember :shrug: an early day comment that got copy pasted around ...

runtime/src/main/java/com/dylibso/chicory/runtime/InterpreterMachine.java
andreaTP: yup: https://github.com/dylibso/chicory/blob/87ecca5f6d40fc15668df16885a5ddbf70b85497/wasm/src/main/java/com/dylibso/chicory/wasm/Validator.java#L583-L593

evacchi (approved):

PR #596, opened 2024-10-23
Add unstructured, typed IR for AOT compiler

electrum:
The IR uses unstructured control flow (i.e. jump/goto) and includes types for `SELECT`, `DROP` and `LOCAL_TEE`. It also directly encodes stack unwinding via the `DROP_KEEP` opcode. This simplifies JVM bytecode generation by extracting the control flow, stack unwinding, and stack/type tracking.

While this simplifies the compiler (at the expense of more code), the goal is to make method splitting simpler to implement, as the IR is easier to analyze than raw WebAssembly.


andreaTP:
I want to do an in-depth review as the PR is on the large side, give me a little time to come to it :pray: 

I just skimmed through and I have a first question, the `TypeStack` and the `Analyzer` are doing a job very similar(in terms of structure) to the [Validator](https://github.com/dylibso/chicory/blob/main/wasm/src/main/java/com/dylibso/chicory/wasm/Validator.java) now, wondering if we can re-use some code there, wdyt?
e.g. using a single implementation of the `analyzeSimple` methods seems doable, this will help keeping the changes isolated when implementing new opcodes.

A note that we are reviewing the public API, and it's possible that we will create git conflicts, sorry for the noise.

electrum:
I agree they are similar. We could possibly move the type annotations to `AnnotatedInstruction` and have the validator attach those. But that might be mixing concerns and make the code more complicated.

aot/src/main/java/com/dylibso/chicory/aot/AotAnalyzer.java
```diff
@@ -0,0 +1,731 @@
...
+                labels.add(ins.labelTrue());
+            }
+            if (ins.labelFalse() != AnnotatedInstruction.UNDEFINED_LABEL) {
+                labels.add(ins.labelFalse());
+            }
+            labels.addAll(ins.labelTable());
+        }
+
+        // implicit block for the function
+        stack.enterScope(FUNCTION_SCOPE, FunctionType.of(List.of(), functionType.returns()));
```
andreaTP, line 78: Are we discarding `functionType.params()`?

This code seems to make use of them:
https://github.com/dylibso/chicory/pull/596/files#diff-89b97bafcb107230b3864b1fc861e71f0eef8784e78fed7987356111d89f632cR634

aot/src/main/java/com/dylibso/chicory/aot/AotAnalyzer.java
```diff
@@ -0,0 +1,731 @@
...
+
+        List<ValueType> dropKeepTypes =
+                stack.types().stream().limit(drop + keep).collect(toCollection(ArrayList::new));
+        reverse(dropKeepTypes);
+        dropKeepTypes.stream().mapToLong(ValueType::id).forEach(operands::add);
+
+        return Optional.of(new AotInstruction(AotOpCode.DROP_KEEP, operands.build().toArray()));
+    }
+
+    private FunctionType blockType(Instruction ins) {
```
andreaTP, line 661: `blockType` is used also for `LOOP`s which is using the params instead of the returns, am I missing something?

aot/src/main/java/com/dylibso/chicory/aot/AotCompiler.java
```diff
@@ -636,272 +597,73 @@ private void compileFunction(
...
                     break;
-                case ELSE:
-                    asm.visitJumpInsn(Opcodes.GOTO, labels.get(ins.labelTrue()));
-                    ctx.popStackSizesStack();
-                    break;
-                case BR:
-                    exitBlockDepth = ins.depth();
-                    if (ins.labelTrue() < idx) {
+                case GOTO:
+                    if (visitedTargets.contains(ins.operand(0))) {
```
andreaTP, line 621: Why are we skipping the interruption check here?
Can we document what's the rationale?

aot/src/main/java/com/dylibso/chicory/aot/AotOpCode.java
```diff
@@ -0,0 +1,237 @@
+package com.dylibso.chicory.aot;
+
+import com.dylibso.chicory.wasm.types.OpCode;
+import java.util.EnumMap;
+import java.util.Map;
+import java.util.Optional;
+
+enum AotOpCode {
+    LABEL,
+    DROP_KEEP,
```
andreaTP, line 10: What's `DROP_KEEP`? What's the semantics?

aot/src/main/java/com/dylibso/chicory/aot/AotEmitters.java
```diff
@@ -95,35 +93,74 @@ public static Builder builder() {
...
+        // save result values
+        int slot = ctx.tempSlot();
+        for (int i = ins.operandCount() - 1; i >= keepStart; i--) {
+            var type = ValueType.forId((int) ins.operand(i));
+            asm.visitVarInsn(storeTypeOpcode(type), slot);
+            slot += slotCount(type);
+        }
+
+        // drop intervening values
+        for (int i = keepStart - 1; i >= 1; i--) {
```
andreaTP, line 113: is this a stack `unwind`?

aot/src/test/resources/com/dylibso/chicory/approvals/ApprovalTest.verifyBrTable.approved.txt
```diff
@@ -55,33 +55,24 @@ public final class com/dylibso/chicory/$gen/CompiledMachine implements com/dylib
 
   public static func_0(ILcom/dylibso/chicory/runtime/Memory;Lcom/dylibso/chicory/runtime/Instance;)I
     ILOAD 0
-    INVOKESTATIC com/dylibso/chicory/$gen/CompiledMachine$AotMethods.checkInterruption ()V
```
andreaTP, line 58: This is an interesting change, do you think it's not needed?

andreaTP (commented):
Sorry for the delay, I finally took the time to go through the code.
It's really great work @electrum :+1: 

I like the direction of those changes, everything becomes a bit easier to understand.
At the same time, there are a couple of details that might be cleaner with a little "design doc", such as:
- what's `GOTO` and the expected semantics
- etc.

I see that this code structure bring us closer to fix the issues we have in the AOT, but I would feel more relaxed about it looking at a "target" PR, where the dots are connected and we can actually test out the those fixes.
This approach will confirm that we are not missing necessary semantics.

aot/src/main/java/com/dylibso/chicory/aot/AotAnalyzer.java
electrum: Good observation, but this is actually a "block type" rather than a "function type". The parameters for a block are on the stack when the block is entered. The function's implicit block is different, as function arguments are in locals rather than on the stack.

aot/src/main/java/com/dylibso/chicory/aot/AotAnalyzer.java
electrum: This is how block types are encoded per the [specification](https://webassembly.github.io/spec/core/syntax/instructions.html#syntax-instr-control):

> A structured instruction can consume input and produce output on the operand stack according to its annotated block type. It is given either as a *type index* that refers to a suitable *function type*, or as an optional *value type* inline, which is a shorthand for the function type `[] -> [valuetype?]`.

aot/src/main/java/com/dylibso/chicory/aot/AotCompiler.java
electrum: There's a comment above that I hoped would explain it
```java
// track targets to detect backward jumps
Set<Long> visitedTargets = new HashSet<>();
```
If the target label has already been visited, then this is a backward jump. Otherwise, it is a forward jump. We document this in `InterpreterMachine.checkinterruption()`:

> Terminate WASM execution if requested. This is called at the start of each call and at any potentially backwards branches. Forward branches and other non-branch instructions are not checked, as the execution will run until it eventually reaches a termination point.

Should we reference this comment in `AotMethods.checkInterruption()`?

aot/src/main/java/com/dylibso/chicory/aot/AotOpCode.java
electrum: The WABT IR has the same instruction. I've been thinking that `UNWIND` might be a better name as that describes the purpose and usage rather. The semantics can be seen in `AotEmitters.DROP_KEEP()`. What it does is to drop `x` number of intermediate stack values, keeping the last `y` values on the stack. This is how unwinds work.

aot/src/main/java/com/dylibso/chicory/aot/AotEmitters.java
electrum: Yes, exactly, this is why I've been thinking to name the opcode `UNWIND` instead.

aot/src/test/resources/com/dylibso/chicory/approvals/ApprovalTest.verifyBrTable.approved.txt
electrum: Yes, it's not needed because none of the switch targets are a backward jump.

electrum:
If you think this existing structure makes the compiler easier to understand and would like to merge it now, then I can create a design doc to help with understanding it. Otherwise, we can wait until the method splitting work is ready, which will make some changes to it.

I looked more at sharing the stack analyzer code with the validator, but I don't think it's worthwhile. The instruction handling is based on the specification and won't ever change, so it's better to have two simple implementations than a shared one that's more complicated.

aot/src/main/java/com/dylibso/chicory/aot/AotCompiler.java
andreaTP: > Should we reference this comment in `AotMethods.checkInterruption()`?

yes please :pray:

andreaTP:
> If you think this existing structure makes the compiler easier to understand and would like to merge it now

Ok, agreed, otherwise it will generate a ton of rebasing work.

> I can create a design doc to help with understanding it

Please, :pray: if possible, I'll prioritize the design doc over the full method splitting implementation, I'll be happy to have more people able to review and reason about the aot compiler.

aot/src/test/resources/com/dylibso/chicory/approvals/ApprovalTest.verifyBrTable.approved.txt
andreaTP: The reason why I'm a bit unsure here is that we are not running the `InterruptionTest` on the aot compiled code.
But I'd prefer to keep the ball rolling other than fussing over a detail that can be reverted at any time.

PR #619, opened 2024-10-28
[chore] Remove callHostFunc from public API of Instance

andreaTP:
Noticed reviewing the public API, `callHostFunction` is an internal implementation and we should not provide the user a helper method to user to invoke it.

### DIFF
--- a/aot/src/main/java/com/dylibso/chicory/aot/AotMethods.java
+++ b/aot/src/main/java/com/dylibso/chicory/aot/AotMethods.java
@@ -26,6 +26,11 @@ public final class AotMethods {
         return instance.getMachine().call(funcId, args);
     }
 
+    public static long[] callHostFunction(Instance instance, int funcId, long[] args) {
+        var imprt = instance.imports().function(funcId);
+        return imprt.handle().apply(instance, args);
+    }
+
     public static boolean isRefNull(int ref) {
         return ref == REF_NULL_VALUE;
     }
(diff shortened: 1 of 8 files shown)


aot/src/main/java/com/dylibso/chicory/aot/AotMethods.java
```diff
@@ -26,6 +26,14 @@ public static long[] callIndirect(long[] args, int typeId, int funcId, Instance
         return instance.getMachine().call(funcId, args);
     }
 
+    public static long[] callHostFunction(Instance instance, int funcId, long[] args) {
+        var imprt = instance.imports().function(funcId);
+        if (imprt == null) {
```
electrum, line 31: This check shouldn't be needed, as we validate the imports in `com.dylibso.chicory.runtime.Instance.Builder#mapHostImports`

So this could be inlined
```java
return instance.imports().function(funcId).handle().apply(instance, args);

File: runtime/src/main/java/com/dylibso/chicory/runtime/InterpreterMachine.java [hunk 2]
```diff
@@ -70,7 +70,11 @@ private static long[] call(
             stackFrame.pushCtrl(OpCode.CALL, 0, type.returns().size(), stack.size());
             callStack.push(stackFrame);
 
-            var results = instance.callHostFunction(funcId, args);
+            var imprt = instance.imports().function(funcId);
+            if (imprt == null) {
```
electrum, line 74: Same here

aot/src/main/java/com/dylibso/chicory/aot/AotMethods.java
andreaTP: good catch!

PR #620, opened 2024-10-28
[docs] improve home page

andreaTP:
This way the home looks a bit nicer :slightly_smiling_face: 
I'm not great at graphics, if anyone want to take over just let me know.


docs/src/components/HomepageFeatures/index.tsx
```diff
@@ -11,7 +11,7 @@ type FeatureItem = {
 const FeatureList: FeatureItem[] = [
   {
     title: 'Pure Java',
-    Svg: require('@site/static/img/pure-java.svg').default,
```
electrum, line 14: I don't feel strongly, but the original with just a `0` looks better to me than the `0%` in the new one.

electrum (approved):
Looks nice!

docs/src/components/HomepageFeatures/index.tsx
andreaTP: done :+1:

PR #621, opened 2024-10-28
Cleanup exceptions in public API

electrum:
* Rename `WASMRuntimeException` to `WasmRuntimeException` to match naming elsewhere
* Should we move `ChicoryException` to `com.dylibso.chicory` instead of `com.dylibso.chicory.wasm`?


andreaTP (approved):
Squashed everything in a single commit to make it easier to resolve conflicts etc.
Hope you don't mind :sweat_smile: 

LGTM, thanks!

PR #623, opened 2024-10-28
Remove empty host-module-parent module

electrum:
We could also consider moving these to the top level.


andreaTP (approved):
I'm good with those changes, I'd like to keep the annotation processing stuff in a separate folder to ease moving them out of the main repo if/when the time comes.

PR #628, opened 2024-10-29
host func module docs

kanatohodets:
Following up on #518, this moves @electrum's Host Module docs into the docs site. I omitted mentions of `@Buffer` and `@CString` since Andrea's feedback on #518 mentioned that those are still in flux.

Also tossed off a script to replace `scripts/update-readme-version.sh`. Same thing but a Perl script which crawls the `docs/docs` dir for any markdown files and swaps `DOCS_PLACEHOLDER_VERSION` for the given value. 

I did take a quick look into Docusaurus plugins to handle this, but I think I'd need a custom remark plugin and I didn't have capacity to go down that road.


andreaTP (commented):
Thanks for following up @kanatohodets !
And, of course, I see you hitting the rough edges :sweat_smile: of the new docs, sorry for that.

- `versioning` we should make sure we have a setup compatible with this: https://docusaurus.io/docs/versioning
- `@Buffer` and `@CString` are fine to be mentioned, but we should add a note regarding the usage limited to function parameters

docs/docs/usage/host-functions.md
```diff
@@ -114,3 +114,79 @@ logIt.apply();
...
+```xml
+<plugin>
+  <groupId>org.apache.maven.plugins</groupId>
+  <artifactId>maven-compiler-plugin</artifactId>
+  <configuration>
+    <annotationProcessorPaths>
+      <path>
+        <groupId>com.dylibso.chicory</groupId>
+        <artifactId>function-processor</artifactId>
+        <version>DOCS_PLACEHOLDER_VERSION</version>
```
andreaTP, line 187: We should be able to remove this version and instead, document the usage of the `BOM`

electrum (approved):

docs/docs/usage/host-functions.md
andreaTP: This can be done as a follow up, @kanatohodets let me know if you are interested in documenting this "super Java specific" aspect of the project, [ref](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#bill-of-materials-bom-poms).

This is an example usage:
- you include the BOM in `dependencyManagement`: https://github.com/andreaTP/opa-chicory/blob/9931be414b6d5f9faca25fa07445b868ab443b25/pom.xml#L66-L76
- you declare all artifact dependencies without the version: https://github.com/andreaTP/opa-chicory/blob/9931be414b6d5f9faca25fa07445b868ab443b25/core/pom.xml#L12-L15

PR #631, opened 2024-10-31
WIP - start implementing EH

andreaTP:
To stress out the stability of the API against new development, I started exploring the implementation of the Exception Handling proposal.

This is still in a very initial state, I'd share only for tracking purposes.

Takeaways so far:

- `ImportValues` overloaded constructors are going to age badly - fixed in #629 
- we need to improve the `test-gen` plugin to be able to enable/disable proposals depending on the folder - e.g. `binary.wast` in the root of testsuite fails with the wrong exception when enabling `exception-handling` in parsing
- last and current blocker: exception handling testsuite requires `tail-call` - I'm probably going to tackle that first


andreaTP:
Most of the rough edges to pave the way to this implementation have been merged, closing and we'll come up with a revamp at a later point in time.

PR #635, opened 2024-11-04
Generate sources in correct package directory

electrum:
These were previously generated in the root of the destination directory.


andreaTP (approved):
LGTM, thanks for the fix!

electrum:
This should fix the issue of IntelliJ not finding these. I have no idea why compilation works. Maybe `javac` only cares about the declared package name?

andreaTP:
That's correct, I discovered it not long time ago:
- package folder structure matching the package name is optional for `javac`
- if there is a mismatch, the IDE breaks

PR #640, opened 2024-11-04
Make WASI Files class public

electrum:


andreaTP:
the visibility is already fixed on `main` (I needed it to split `wasi-tests`), feel free to update the comment :+1:

electrum:
Updated to just add Javadoc

PR #641, opened 2024-11-04
Implement WASI path_readlink

electrum:
Python requires this when running `python foo.py`

### DIFF
--- a/docs/docs/usage/wasi.md
+++ b/docs/docs/usage/wasi.md
@@ -163,7 +163,7 @@ For the most up-to-date info, and to see what specific functions we support, see
 | path_filestat_set_times | ✅         |                                                                           |
 | path_link               | ❌         |                                                                           |
 | path_open               | ✅         |                                                                           |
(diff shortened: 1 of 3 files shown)


andreaTP:
The typo should be fixed on `main`.

andreaTP (approved):
I'm not super happy that there are no tests for this, would it be possible to enable something like: `tests/rust/testsuite/path_link.wasm` ? and maybe share some code path?
Alternatively, is it complex to do a test in isolation to cover the "happy path"?

Not going to oppose to merge though, as it looks a straight improvement anyhow, thanks.

electrum:
It has a test in `WasiPreview1Test`. We can't run `path_link.rs` because requires `path_link`. I'm hesitant to implement that as symlinks are currently a security hole if using the normal file system.

andreaTP:
> It has a test 

Right, I thought from the `wasi-testsuite`, but that's good enough :+1:  thanks!

PR #645, opened 2024-11-05
Generate separate AOT class for Machine call method

electrum:
This is needed to avoid exhausting the constant pool for modules with thousands of functions (e.g., Python).

### DIFF
--- a/aot/src/main/java/com/dylibso/chicory/experimental/aot/AotMethodInliner.java
+++ b/aot/src/main/java/com/dylibso/chicory/experimental/aot/AotMethodInliner.java
@@ -1,5 +1,6 @@
 package com.dylibso.chicory.experimental.aot;
 
+import static com.dylibso.chicory.experimental.aot.AotUtil.internalClassName;
 import static org.objectweb.asm.Type.getInternalName;
 
 import com.dylibso.chicory.wasm.ChicoryException;
@@ -32,7 +33,7 @@ final class AotMethodInliner {
                         super.visit(
                                 version,
                                 Opcodes.ACC_PRIVATE | Opcodes.ACC_FINAL,
-                                AotUtil.internalClassName(className + "$AotMethods"),
+                                internalClassName(className + "$AotMethods"),
                                 null,
                                 superName,
                                 null);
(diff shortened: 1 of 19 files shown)


andreaTP (commented):
side note:
I'm not sure why the CI is not being triggered ...

regarding the PR:
I like the separation that it's introduced here, but IIRC, the indirection(e.g. going through the `Machine.call` invocation), is needed only for `CALL_INDIRECT`, and `CALL_HOST`, or am I missing an invariant?

.gitignore
```diff
@@ -15,3 +15,5 @@ scripts/jmh-tmp
 
 /*.wasm
 readmes/*/current
+
+.approval_tests_temp/
```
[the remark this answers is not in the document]
electrum: Fixed

electrum:
`CALL_INDIRECT` is handled via the `call_indirect_xxx` methods, and host function bridges are compiled into `func_xxx` just like native WASM methods.

andreaTP:
Thanks for the further explanation :+1:

andreaTP (approved):

electrum:
I'm going to update this to move the `call_xxx` bridges to the separate class.

electrum:
@andreaTP Do you want to take another look?

aot/src/main/java/com/dylibso/chicory/experimental/aot/AotMethodInliner.java
```diff
@@ -31,16 +32,16 @@ public void visit(
...
+                                Opcodes.ACC_FINAL | Opcodes.ACC_SUPER,
+                                internalClassName(className + "$AotMethods"),
                                 null,
                                 superName,
                                 null);
                     }
 
                     @Override
                     public void visitEnd() {
-                        visitNestHost(AotUtil.internalClassName(className));
```
andreaTP, line 43: nit: do you have an IDE setting that does this change?
is it possible to enforce it, project wise, using spotless?
otherwise we are going to do forth and back when we edit the same files ...

andreaTP (approved):
LGTM, looks great, thanks!

aot/src/main/java/com/dylibso/chicory/experimental/aot/AotMethodInliner.java
electrum: No, I do it manually.

andreaTP: If you really care about it, please write a guideline somewhere, otherwise I think we are going to have a lot of changes in this detail... personally I don't care much ...

PR #648, opened 2024-11-06
Simplify AOT code using InstructionAdapter

electrum:


andreaTP (commented):
Where can I read more about the different tradeoffs of using `MethodVisitor` as opposed to `InstructionAdapter`?

I think I found various codebases on the internet using `MethodVisitor`s to build bytecode, is there any reason other than a slightly better readability?

If @danielperano has an opinion and finds the time to share, I'm interested in it.

electrum:
I would be happy to try switching to the Class-File API. We could use @dmlloyd's https://github.com/dmlloyd/jdk-classfile-backport if we are willing to move to JDK 17.

andreaTP:
Let's re-evaluate after the class files API gets released(out of preview) in March 2025 (JDK 24 target release date).

PR #649, opened 2024-11-06
Include function debug name when too large

electrum:


andreaTP (approved):
LGTM, thanks!
I noticed that this mechanism for returning the function names is used even in other Wasm proposals.
In the future, we can even bake this logic in the `WasmModule` as it will be re-used(e.g. EH).

electrum:
This is using `nameSection()`. All the logic here is to extract the function ID from the Java method name, as this exception is thrown when the class is being written, so all we have is the Java method name to know which WASM function is too large.

electrum:
I was thinking we could extend this to look at the exports to see if there is an export name for the function ID in the case where the debug name is not available.

andreaTP:
Yeah, looking at the exports would be an additional improvement.
Maybe move down most of the logic(without the "string extraction" part) to a helper method in `WasmModule` ?

This way we can re-use the information in more places.

PR #652, opened 2024-11-06
Scaffolding to get started with SIMD on the Vector API - again

andreaTP:
Supersede #611 , opening a new one to enable comparing the `additionalOpcodes` approach to this `EvalDefault`.
I also moved away from using reflection in the tests, and rely on `java-templates` to do the switch.

Everything looks good in my IDE, but a second pair of eyes is welcome.


andreaTP:
I'm not 100% sold on the `EvalDefault` mechanism I implemented, maybe I misinterpreted your comment @electrum , but referencing from the static context has been a challenge to get all the pieces together.
Feel free to review and eventually, propose an alternative ...

simd/src/main/java/com/dylibso/chicory/simd/SimdInterpreterMachine.java
```diff
@@ -0,0 +1,399 @@
...
+                                                            (int)
+                                                                    ((val1Low >> shift)
+                                                                            & 0xFFFFFFFFL))
+                                                    * Float.intBitsToFloat(
+                                                            (int)
+                                                                    ((val2Low >> shift)
+                                                                            & 0xFFFFFFFFL)))
+                                    & 0xFFFFFFFFL)
+                            << shift;
+        }
```
evacchi, line 264: I guess this is spotless applying some weird formatting, can you maybe assign some values to variables to avoid this weird spacing?

evacchi (approved):
I think it's fine overall, we can refine further as we go.

runtime-tests/pom.xml
```diff
@@ -218,6 +217,258 @@
           </execution>
         </executions>
       </plugin>
+
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>templating-maven-plugin</artifactId>
```
electrum, line 223: This seems complicated. We should be able to instantiate conditionally. The class will be initialized lazily, so something like this should work:
```java
return Runtime.version().comareTo(Version.parse("21")) >= 0
    ? new SimdInterpreterMachine(instance)
    : new InterpreterMachine(instance);
```
The class won't be loaded on the older JDK version.

machine-tests/src/test/java/com/dylibso/chicory/testing/MachinesTest.java
```diff
@@ -204,9 +206,13 @@ public void shouldUseMachineCallOnlyForExport() throws Exception {
                         .withMachineFactory(
                                 (inst) -> {
                                     var machine = Wat2WasmMachineFactory.create(inst);
-                                    return (funcId, args) -> {
-                                        assertEquals(startFunctionIndex.get(), funcId);
-                                        return machine.call(funcId, args);
+                                    return new Machine() {
```
electrum, line 209: This change to use an anonymous class shouldn't be needed anymore. We could make `Machine` have `@FunctionalInterface`.

runtime/src/main/java/com/dylibso/chicory/runtime/Machine.java
```diff
@@ -5,4 +5,17 @@
 public interface Machine {
 
     long[] call(int funcId, long[] args) throws ChicoryException;
+
+    @FunctionalInterface
+    interface Operands {
+        long get(int index);
+    }
+
+    static int readMemPtr(MStack stack, Operands operands) {
```
electrum, line 14: Make this a static protected method on `InterpreterMachine`. It doesn't belong on `Machine` as it's only useful for the interpreter (since it uses `MStack`).

simd/src/main/java/com/dylibso/chicory/simd/SimdInterpreterMachine.java
electrum: Yeah, this the downside of an automatic formatter

electrum (approved):

simd/src/main/java/com/dylibso/chicory/simd/SimdInterpreterMachine.java
andreaTP: I'm improving a little, I'm not too worried, as in this case those implementations are just "stubbed" waiting to be fully replaced with the usage of Vector API.

runtime/src/main/java/com/dylibso/chicory/runtime/Machine.java
andreaTP: good catch, I'll move it back

runtime/src/main/java/com/dylibso/chicory/runtime/InterpreterMachine.java
```diff
@@ -95,15 +122,16 @@ private static long[] call(
...
+        var totalResults = sizeOf(type.returns());
         var results = new long[totalResults];
         for (var i = totalResults - 1; i >= 0; i--) {
             results[i] = stack.pop();
         }
         return results;
     }
 
-    static void eval(MStack stack, Instance instance, Deque<StackFrame> callStack)
+    static void eval(
```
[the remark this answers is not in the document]
andreaTP: I'm afraid of the consequences on performance of such change, I'll try to measure first.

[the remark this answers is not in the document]
andreaTP: I did it here:
https://github.com/dylibso/chicory/commit/22e2d98d3528efa63bcb125a00247b05e3fc6988

let's keep this change separated as a follow-up, as it showed performance impact from the first early tests.

PR #655, opened 2024-11-07
Use inheritance for eval

andreaTP:
Follow up from [this comment](https://github.com/dylibso/chicory/pull/652#discussion_r1832392005).

I noticed performance differences locally, but my machine was pretty noisy.

### DIFF
--- a/simd/src/main/java/com/dylibso/chicory/simd/SimdInterpreterMachine.java
+++ b/simd/src/main/java/com/dylibso/chicory/simd/SimdInterpreterMachine.java
@@ -15,10 +15,11 @@ import jdk.incubator.vector.LongVector;
 public class SimdInterpreterMachine extends InterpreterMachine {
 
     public SimdInterpreterMachine(Instance instance) {
-        super(instance, SimdInterpreterMachine::evalDefault);
+        super(instance);
     }
 
-    protected static void evalDefault(
+    @Override
+    protected void evalDefault(
             MStack stack,
             Instance instance,
             Deque<StackFrame> callStack,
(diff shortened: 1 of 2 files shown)


andreaTP:
It's not really bad, but looks a slight bit worst:
Screenshot from 2024-11-07 15-17-46

The code is slightly cleaner, but I would pass and not merge this one.

electrum (approved):

electrum:
It might not be worse, as all the results are within the error bounds. I'll run again and see what it looks like.

andreaTP:
> I'll run again and see what it looks like.

:+1: let me know how it goes, I have had constantly, slightly worst results, but I'll test on a dedicated machine if you get something different.

electrum:
I think this shows that the benchmarks are not very accurate.

Screenshot 2024-11-11 at 10 07 12 AM

simd/src/main/java/com/dylibso/chicory/simd/SimdInterpreterMachine.java
```diff
@@ -15,10 +15,11 @@
 public class SimdInterpreterMachine extends InterpreterMachine {
```
electrum, line 15: Can you try making this `final`? I doubt it will have any impact on performance, but it doesn't hurt, and we'd expect the class to be final anyway.

andreaTP:
The benchmarks are not accurate, the CI machines are just giving us an indication, that's accurate.
I'll try to run on a clean machine and I get back :+1:

andreaTP:
I have been running the benchmarks on a quiet machine and you have the results attached [bench.txt](https://github.com/user-attachments/files/17733321/bench.txt)

Now I'm convinced this is not making performance much worse.
Going to rebase and finalize the PR :+1:

PR #659, opened 2024-11-09
Generate self-contained modules for pre-compiled AOT classes

electrum:


andreaTP:
I like the approach, it is simple and achieves the result without much fuss.

Thanks a lot!

I'll do a full review sometime soon (e.g. do not change the blog post but have docs about the feature etc.)

aot-maven-plugin/src/main/java/com/dylibso/chicory/experimental/maven/aot/AotGenMojo.java
```diff
@@ -123,12 +129,126 @@ public void execute() throws MojoExecutionException {
...
+                wasmBytes,
+                section -> {
+                    if (section.sectionId() == SectionId.CODE) {
+                        var out = new ByteArrayOutputStream();
+                        int count = module.codeSection().functionBodyCount();
+                        writeVarUInt32(out, count);
+                        for (int i = 0; i < count; i++) {
+                            writeVarUInt32(out, 3); // function size in bytes
+                            writeVarUInt32(out, 0); // locals count
+                            out.write(OpCode.UNREACHABLE.opcode());
```
andreaTP, line 241: This looks a bit strange, instead of just removing the function section we are substituting the functions with "unreachable" code.
I understand that the validation will fail but I would rather disable the validation(since it has already been running on the original module) and remove an entire section instead of changing its content.

wasm/src/main/java/com/dylibso/chicory/wasm/RawParserListener.java
```diff
@@ -0,0 +1,9 @@
+package com.dylibso.chicory.wasm;
+
+import com.dylibso.chicory.wasm.types.RawSection;
+
+@FunctionalInterface
+public interface RawParserListener {
```
andreaTP, line 6: Genuine question, we use the same patter with a Listener for normal Parsing, what's the advantage we get against a more direct approach?

wasm/src/main/java/com/dylibso/chicory/wasm/WasmWriter.java
```diff
@@ -0,0 +1,44 @@
+package com.dylibso.chicory.wasm;
+
+import static java.lang.Integer.toUnsignedLong;
+
+import com.dylibso.chicory.wasm.types.RawSection;
+import java.io.ByteArrayOutputStream;
+
+public final class WasmWriter {
+    private static final byte[] MAGIC = {0x00, 0x61, 0x73, 0x6D};
```
andreaTP, line 9: I think we should re-use the same constants as the Parser?
https://github.com/dylibso/chicory/blob/cad19fef9b96179e76bc0f795e390a6f6e8aa662/wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java#L80

or have a separate `Constants` file in the module

wasm/src/test/java/com/dylibso/chicory/wasm/WasmWriterTest.java
```diff
@@ -0,0 +1,31 @@
...
+import static com.dylibso.chicory.wasm.ParserTest.wasmCorpusFiles;
+import static org.junit.jupiter.api.Assertions.assertArrayEquals;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import org.junit.jupiter.api.MethodOrderer;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestMethodOrder;
+
+@TestMethodOrder(MethodOrderer.OrderAnnotation.class)
```
andreaTP, line 13: copy paste?

wasm/src/test/java/com/dylibso/chicory/wasm/WasmWriterTest.java
```diff
@@ -0,0 +1,31 @@
...
+public class WasmWriterTest {
+
+    @Test
+    public void shouldRoundTrip() throws IOException {
+        for (var file : wasmCorpusFiles()) {
+            // uses non-canonical size encodings
```
andreaTP, line 19: What does this means? Can you elaborate?
Sounds more like a bug ...

aot-maven-plugin/src/main/java/com/dylibso/chicory/experimental/maven/aot/AotGenMojo.java
```diff
@@ -123,12 +129,126 @@ public void execute() throws MojoExecutionException {
...
+                new ObjectCreationExpr()
+                        .setType(parseClassOrInterfaceType("UncheckedIOException"))
+                        .addArgument(new StringLiteralExpr("Failed to load AOT WASM module"))
+                        .addArgument(new NameExpr("e"));
+        var catchIoException =
+                new CatchClause()
+                        .setParameter(
+                                new com.github.javaparser.ast.body.Parameter(
+                                        parseClassOrInterfaceType("IOException"), "e"))
+                        .setBody(new BlockStmt(new NodeList<>(new ThrowStmt(newException))));
```
andreaTP, line 218: Generated code should not catch Exception unless in specific cases.
I would just add the Exception to the signature.

docs/blog/2024-11-08-chicory-1.0.0-M1.md
```diff
@@ -285,8 +285,8 @@ calling a native executable.
           <configuration>
             <!-- Translate the Wasm binary `wat2wasm` into bytecode -->
             <wasmFile>src/main/resources/wat2wasm</wasmFile>
-            <!-- Generate the following class file as a result -->
-            <name>com.dylibso.chicory.wabt.Wat2WasmModule</name>
+            <!-- Generate the result classes with this prefix -->
+            <name>com.dylibso.chicory.wabt.Wat2Wasm</name>
```
andreaTP, line 289: Instead of generating two classes with one static method each, can we create 1 class with 2 static methods?

docs/blog/2024-11-08-chicory-1.0.0-M1.md
```diff
@@ -285,8 +285,8 @@ calling a native executable.
           <configuration>
             <!-- Translate the Wasm binary `wat2wasm` into bytecode -->
             <wasmFile>src/main/resources/wat2wasm</wasmFile>
-            <!-- Generate the following class file as a result -->
```
andreaTP, line 288: We should add a documentation section in the docs instead of changing the content of the blog.

andreaTP (commented):
I like the direction overall!

Added a few comments around, I'm not 100% convinced around the "function bodies mangling", and I would prefer to just strip the section at all.
If we really really need to keep the current approach it should be documented appropriately, as I can see people getting extremely surprised in finding a mangled version of the wasm module on the classpath.

aot-maven-plugin/src/main/java/com/dylibso/chicory/experimental/maven/aot/AotGenMojo.java
```diff
@@ -123,12 +129,126 @@ public void execute() throws MojoExecutionException {
...
+        methodBody.addStatement(
+                new TryStmt()
+                        .setResources(new NodeList<>(resourceVar))
+                        .setTryBlock(new BlockStmt(new NodeList<>(returnStmt)))
+                        .setCatchClauses(new NodeList<>(catchIoException)));
+
+        dest.add(packageName, moduleName + ".java", cu);
+    }
+
+    private static byte[] rewriteWasm(byte[] wasmBytes, WasmModule module) {
```
andreaTP, line 229: Instead of rewriting in-place the existing wasm file, we can place the original `.wasm` files outside `resources` and produce the result in `target/generated-resources`.

andreaTP: already doing it

andreaTP:
I did a little exploration [here](https://github.com/electrum/chicory/pull/3), I don't mind about attributions, feel free to take from it without keeping the commit :slightly_smiling_face:

aot-maven-plugin/src/main/java/com/dylibso/chicory/experimental/maven/aot/AotGenMojo.java
electrum: I agree it's strange, but having valid modules seems better. Making invalid modules work requires the following:

* Add a new API to `Parser` to skip validation
* Change two places in `Instance` to use `functionTypes.length` instead of `functions.length`
* Change `WasmModule` to skip calling `validateModule()` when validation is disabled

It also prevents using tools such as `wasm2wat` to view the modules. The `--no-check` flag doesn't help.

So if you really feel strongly about using invalid modules, I can change it, but having valid modules seems like the overall cleaner solution.

electrum: Skipping structural validation of the module as opposed to contents seems like a dangerous path and makes the APIs more complicated. There's a difference between *this function body has an invalid instruction stack height* and *these two linked data structures are mismatched*.

wasm/src/main/java/com/dylibso/chicory/wasm/WasmWriter.java
electrum: If you're fine to change the parser to use a byte array. I copied the hex format from the specification.

wasm/src/test/java/com/dylibso/chicory/wasm/WasmWriterTest.java
electrum: Good catch. I copied `ParserTest` but it's also not needed there, either.

aot-maven-plugin/src/main/java/com/dylibso/chicory/experimental/maven/aot/AotGenMojo.java
andreaTP: What I feel strongly about is:
we are packing in the Jar a wasm binary with the same name as the original, but that the user won't be able to use.

I can see users wasting a lot of time before figuring out an "hidden" function body substitution happened without them realizing.

With this context I'd prefer a fully invalid wasm module that will clearly indicate that "something happened" and it's not the original module.

That said I'm fully open to hearing more alternatives:

- keep the original wasm file(how much do we save?)
- have a mechanism the user should explicitly enable to mangle the file
- have an external tool that will explicitly perform the operation when needed
- add yours :-)

wasm/src/main/java/com/dylibso/chicory/wasm/WasmWriter.java
andreaTP: Totally fine with it :+1: any downside I haven't considered?

wasm/src/test/java/com/dylibso/chicory/wasm/WasmWriterTest.java
electrum: The `main.go.wasm` file encodes a section size as `F2 80 80 80 00` rather than the canonical `72`. When writing the output, we write the size using the canonical encoding, which means the round-trip test will fail as the bytes are different between the original and what we wrote.

aot-maven-plugin/src/main/java/com/dylibso/chicory/experimental/maven/aot/AotGenMojo.java
electrum: This is try-with-resources, which is needed to close the `Class.getResourceAsStream()` stream. This follows the model of `Parser.parse()` in not throwing a checked exception.

From a practical standpoint, making `load()` throw `IOException` would be annoying, as it would prevent this type of usage, requiring a static initializer block:
```java
private static final WasmModule MODULE = Wast2JsonModule.load();
```

Changing `Parser` makes more sense, as that can be used with arbitrary resources. This code is used with a resource in the same JAR, so any failure would be quite unexpected, and not something that user code is expected to handle (which is the typical reasoning for checked exceptions).

docs/blog/2024-11-08-chicory-1.0.0-M1.md
electrum: Agreed, I forgot that this blog post is about the published `-M1` artifact which won't be affected by this change.

aot-maven-plugin/src/main/java/com/dylibso/chicory/experimental/maven/aot/AotGenMojo.java
andreaTP: Thanks for the explanation, convinced :+1:

wasm/src/main/java/com/dylibso/chicory/wasm/RawParserListener.java
electrum: It seemed a bit cleaner to avoid the cast in the listener implementation. It's more type-safe, in that the method is guaranteed to always have `RawSection`.

But I can remove this as we only use it one place.

docs/blog/2024-11-08-chicory-1.0.0-M1.md
electrum: That's a good idea. I'll update it.

wasm/src/main/java/com/dylibso/chicory/wasm/WasmWriter.java
electrum: Done

aot-maven-plugin/src/main/java/com/dylibso/chicory/experimental/maven/aot/AotGenMojo.java
electrum: The savings are both size and load time (no need to parse or validate the code). Here's some size comparisons with the amount saved:

```
wast2json 1_523_335 ->   634_206 (2.4x)
wat2wasm  1_482_029 ->   631_428 (2.3x)
python   26_549_480 -> 3_064_154 (8.7x)
```

We can mention the mangling in the documentation. I'm doubtful that people will try to use it, since they had to go through the effort of using the plugin to generate the AOT code, which is the whole point.

What if we name it differently, maybe call it `.meta` instead of `.wasm`? Then people won't be tempted to use it.

electrum:
BTW, in my original implementation I simply removed the code section. But then I was annoyed that `wasm2wat` didn't work.

aot-maven-plugin/src/main/java/com/dylibso/chicory/experimental/maven/aot/AotGenMojo.java
andreaTP: > I'm doubtful that people will try to use it

I'm not that much :slightly_smiling_face: , apart from the chances, narrowing down this issue without knowing about what is happening behind the scenes is going to be difficult.

> What if we name it differently

Ok, this might be a fair compromise, what about `.aot-data` or similar?

wasm/src/test/java/com/dylibso/chicory/wasm/WasmWriterTest.java
andreaTP: I didn't know about this and found:
https://github.com/WebAssembly/design/issues/533

That's surprising, but ok.

andreaTP (approved):
Thanks for making the changes.

Good to go after we change the name of the "mangled" wasm module:
https://github.com/dylibso/chicory/pull/659/files#diff-d8f5741894c8ce3634be19159b74d9d949676e184c8d18e96e331f284dcdc907R115

I assume we can follow up with the documentation.

PR #663, opened 2024-11-10
Using NIO instead of IO [#661]

### LINKED ISSUE #661: Please don't use java.io.File
The current code base is using the rather obsolete `java.io.File` class in many places, in particular as part of the public API. As the code base uses the modern (and faster, and safer, and correcter) NIO API internally, conversion between `File` and `Path` happens frequently. This makes the code more complex than needed, and it makes runtime slower than needed, as conversion takes considerable time. Also, chicory forces applications to adopt an obsolete API, which is not nice. As typically no existing software will depend on chicory but solely newly written software, this API choice is incomprehensible. All modern software should always use NIO API solely but never need to fall back to `java.io.File`, so it makes no sense to enforce the use of a slow an incorrect API.

Please replace `java.io.File` by `Path` everywhere in your code base. Thanks.

*Disclaimer: [I am a contributor to OpenJDK](https://openjdk.org/census#mkarg). In particular I am working on performance enhancements in the area of IO and NIO. My intention here is to evangelize about modern Java IO.*

mkarg:
This PR aims to fix #661 by replacing `java.io.File` by `java.nio.file.Path` everywhere besides Unit Tests.

Closes #661


aot-maven-plugin/src/main/java/com/dylibso/chicory/experimental/maven/aot/AotGenMojo.java
```diff
@@ -69,74 +68,78 @@ public class AotGenMojo extends AbstractMojo {
 
     @Override
     public void execute() throws MojoExecutionException {
-        var module = Parser.parse(wasmFile);
-        var result = AotCompiler.compileModule(module, name);
-        var split = name.split("\\.");
+        try {
```
electrum, line 71: Can we change the API without changing the try/catch structure? That will make the diff easier to review.

mkarg: done :-)

mkarg:
Due to https://github.com/dylibso/chicory/issues/705 I have no chance to test locally. 🙁

mkarg:
> Quick heads up @mkarg if you are still motivated in getting breaking changes merged, you have approx 2 weeks. Those changes will need to be done in a backward compatible way afterwords.

Thank you for the kind reminder. I now have remove `File` from the public API, so the breaking change will be contained. :-)

andreaTP:
> Due to https://github.com/dylibso/chicory/issues/705 I have no chance to test locally.

Answered on the issue, worst case scenario, a viable workaround is, in the root of the project:

```
git clone --branch prod/testsuite-base https://github.com/WebAssembly/wasi-testsuite.git
```

and everything should work after it.

mkarg:
> Thanks for updating the PR! I think you miss a `mvn spotless:apply` to get the CI happy.
> 
> As a general advice, I always run a full `mvn spotless:apply clean install` as a sanity check that the PR will pass the CI.

Okay, so why not simply binding `spotless:apply` to MVN's `verify` phase in the POM then? This would automatically run the spotless checks as part of the test suite.

BTW, `mvn clean install` is really (really!) an anti-pattern in the MVN universe. I assume what you *actually* want contributors to do is `mvn verify`.

andreaTP:
> Okay, so why not simply binding `spotless:apply` to MVN's `verify` phase in the POM then?

We discussed this approach in the past, the "sad" answer is that the current Spotless setup is not "perfect" and automatically running it goes in the way when doing refactorings (e.g.: removing needed imports and more).
I would be more than happy to reconsider this position and running the tool automatically when the situation improves in this sense.

> `mvn clean install` is really (really!) an anti-pattern in the MVN universe

I have heard about this opinion in the past, personally, I'm not a dogmatic person.
Of course, this approach has limitations, but I cannot really find a compelling reason why it would be harmful in any way.
To enable `mvn verify` we would need to publish the Maven plugins from a separate repository, this is really sub-optimal for the project and will need more infrastructure to coordinate releases etc.:
- the maven plugins are used to run the generated tests
- the implementation of the maven plugins is doubly linked with the execution of those: i.e. changes in the maven plugin should be rolled out together with other changes in the repo

I'm contributing to more large upstream repos and asking to run `mvn install` instead of `mvn verify` is a minimal disadvantage compared to the complexity and unnecessary maintenance we would need to sustain.
Happy to revisit when someone provides compelling arguments hat apply to this specific repo and setup(e.g. not theoretical advantages).

andreaTP:
Regarding this PR, the CI is failing with:

> Cannot create instance of interface java.nio.file.Path: java.nio.file.Path

Seems like the Maven Plugins cannot automatically construct `Path` from the configuration(it works with `File`).

mkarg:
> Regarding this PR, the CI is failing with:
> 
> > Cannot create instance of interface java.nio.file.Path: java.nio.file.Path
> 
> Seems like the Maven Plugins cannot automatically construct `Path` from the configuration(it works with `File`).

Sigh... yes, this is what happens if a contributor is forced to push blindly, as the build is failing to even compile HEAD of `main`... 😢 

I will look into this, and hopefully you can fix the build in shared Gitpod workspace meanwhile, so I can actually run `mvn verify` before pushing to Github next time... 🙂

mkarg:
Help needed to finish this PR:
`cd wabt && mvn clean install` works fine...
```
[INFO] ----------------------< com.dylibso.chicory:wabt >----------------------
[INFO] Building Chicory - wabt 999-SNAPSHOT
...
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
```
...but a full `mvn clean install` run *on the root path* stops exactly at that very same `wabt` module:
```
[INFO] Chicory - wabt ..................................... FAILURE [  0.045 s]
...
failed to read WASM file: src/main/resources/wast2json
```
This is paradox...?!

andreaTP:
I'm going to close this for the reasons explained in #661 

For visibility:
- extensive support in terms of reproducing the user environment have been provided (also fixing the mentioned shared GitPod workspace)
- lengthy answers [have been shared](https://github.com/dylibso/chicory/pull/663#issuecomment-2539511946)

The value of this improvement doesn't counterbalance the effort everyone is spending in it.
I decide to accept the sunk cost and move forward without merging.

PR #670, opened 2024-11-13
Only implement AOT Machine call for exported functions

electrum:
The generated switch statement is too large for modules with thousands of functions. For example, Python has almost 10k functions, but we only need the few that are exported.


electrum:
I forgot that we also use `Machine.call()` for linking modules via `CALL_INDIRECT`.

PR #674, opened 2024-11-15
Split AOT Machine call into multiple methods

electrum:


aot/src/main/java/com/dylibso/chicory/experimental/aot/AotCompiler.java
```diff
@@ -84,6 +86,8 @@ public final class AotCompiler {
     private static final MethodType MACHINE_CALL_METHOD_TYPE =
             methodType(long[].class, Instance.class, Memory.class, int.class, long[].class);
 
+    private static final int MAX_MACHINE_CALL_METHODS = 1024; // must be power of two
```
andreaTP, line 89: Is the current limit this low?
This is the reference limit if I understand it correctly:
https://github.com/dylibso/chicory/blob/0c2ed8e2e4581b83351d4796632ca18fccd17142/wasm/src/main/java/com/dylibso/chicory/wasm/WasmLimits.java#L8

aot/src/main/java/com/dylibso/chicory/experimental/aot/AotCompiler.java
```diff
@@ -388,12 +392,23 @@ private byte[] compileMachineCallClass() {
...
-        emitFunction(
-                classWriter,
-                "call",
-                MACHINE_CALL_METHOD_TYPE,
-                true,
-                this::compileMachineCallInvoke);
+        Consumer<InstructionAdapter> callMethod;
+        if (functionTypes.size() < MAX_MACHINE_CALL_METHODS) {
+            callMethod = asm -> compileMachineCallInvoke(asm, 0, functionTypes.size());
+        } else {
```
andreaTP, line 398: ok, so here we are saying that after a fixed number of functions we can dispatch the call, I assume that this is the building block to split.

andreaTP (commented):
I think I understand the direction you are going here, but I would like to have a better understanding of the target implementation:
- here we are still emitting the functions to "trampoline" on the class(without the body iiuc), are we going to hit limits about "number of functions"?
- do we need the trampolines at all?

I'm thinking if we can rely on "assumptions", e.g.:
- we decide that each class contains X function bodies
- `CompiledModule0` contains functions from 0 to X, `CompiledModule1` from X+1 to 2*X, etc. etc.

Do you think that the number to split at can be tweaked by the user? (I see a few possible issues, but maybe can stay as a metadata in the generated code too)

aot/src/main/java/com/dylibso/chicory/experimental/aot/AotCompiler.java
electrum: Correct. We generate an outer dispatch method that calls the correct inner dispatch method.

andreaTP (approved):
I'm good with the explanations, thanks for clarifying.

This change should be documented, though, otherwise I can see myself failing to understand it in a couple of weeks.

If you want to tackle this in the broader scope of the AOT compiler design docs feel free to go for it, alternatively, at least a block of text in a comment is the minimum :pray:

PR #686, opened 2024-11-25
simd: apply some fixes to test generation

evacchi:
This fixes some aspects of the SIMD test codegen, introducing some (debatable) wider changes, so I'm going to push this as a draft, to let you judge :) some of these changes could be also applied to a utility class, but this should be the gist anyway

- I am now passing args through an `MStack` so that we can push V128s onto it: the reason is that codegen might become complicated otherwise: each v128 is a `long[]` so, you would have to "unpack" each long to several pushes. This way you can just push it all at once.
- I have added an `Mstack.pushAll(long[])` method, so that we can easily push vectors all at once.
- I have added an `MStack.slice(start, len)` method returning _a copy_ of the array (thus, only useful for tests), so that 
   a. all places expecting a given arg size (e.g. I have seen this behavior in the new tail call feature) do not get confused by a larger input array
   b. we can pass in a number of values we want to pull; this is important when there are v128 values, since they are size=2
- I have added a `WasmValueType.size()` that is 2 for v128 and 1 otherwise

I have also added a few more missing (?) lanes (i64x2, F64x2) and the reverse of `vecTo32()`, etc...

I will add some other notes inline.

@mfirry this should unblock you! 

folks, feel free to push directly to this branch to apply your fixes on top

### DIFF
--- a/test-gen-plugin/src/main/java/com/dylibso/chicory/maven/JavaTestGen.java
+++ b/test-gen-plugin/src/main/java/com/dylibso/chicory/maven/JavaTestGen.java
@@ -102,7 +102,18 @@ public class JavaTestGen {
         cu.addImport("com.dylibso.chicory.wasm.types.Value.vecTo8", true, false);
         cu.addImport("com.dylibso.chicory.wasm.types.Value.vecTo16", true, false);
         cu.addImport("com.dylibso.chicory.wasm.types.Value.vecTo32", true, false);
+        cu.addImport("com.dylibso.chicory.wasm.types.Value.vecTo64", true, false);
         cu.addImport("com.dylibso.chicory.wasm.types.Value.vecToF32", true, false);
+        cu.addImport("com.dylibso.chicory.wasm.types.Value.vecToF64", true, false);
+
+        cu.addImport("com.dylibso.chicory.wasm.types.Value.i8ToVec", true, false);
+        cu.addImport("com.dylibso.chicory.wasm.types.Value.i16ToVec", true, false);
+        cu.addImport("com.dylibso.chicory.wasm.types.Value.i32ToVec", true, false);
+        cu.addImport("com.dylibso.chicory.wasm.types.Value.i64ToVec", true, false);
+        cu.addImport("com.dylibso.chicory.wasm.types.Value.f64ToVec", true, false);
+        cu.addImport("com.dylibso.chicory.wasm.types.Value.f32ToVec", true, false);
+
+        cu.addImport("com.dylibso.chicory.runtime.MStack", false, false);
 
         // import for Store instance
         cu.addImport("com.dylibso.chicory.runtime.Store");
@@ -372,11 +383,17 @@ public class JavaTestGen {
                 (cmd.action().args() != null)
                         ? Arrays.stream(cmd.action().args())
                                 .map(WasmValue::toArgsValue)
-                                .collect(Collectors.joining(", "))
-                        : "";
+                        .collect(Collectors.toList())
+                        : List.<String>of();
+
+        var argSize =
+                (cmd.action().args() != null)
+                        ? Arrays.stream(cmd.action().args()).mapToInt(a -> a.type().size()).sum()
+                        : 0;
+
         // Function or Global
         var invocationMethod =
-                (cmd.action().type() == INVOKE) ? ".apply(" + args + ")" : ".getValue()";
+                (cmd.action().type() == INVOKE) ? ".apply(args.slice(0, " + argSize + "))" : ".getValue()";
         if (cmd.type() == CommandType.ASSERT_TRAP || cmd.type() == CommandType.ASSERT_EXHAUSTION) {
             var assertDecl =
                     new NameExpr(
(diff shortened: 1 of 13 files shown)


test-gen-plugin/src/main/java/com/dylibso/chicory/maven/wast/WasmValue.java
```diff
@@ -165,11 +186,14 @@ public String toExpectedValue() {
...
+                            case I64:
+                                sb.append("Long.parseUnsignedLong(\"" + v + "\")");
+                                break;
                             case F32:
-                                sb.append(
-                                        "Float.intBitsToFloat(Integer.parseUnsignedInt(\""
-                                                + v
-                                                + "\"))");
+                                sb.append("Integer.parseUnsignedInt(\"" + v + "\")");
+                                break;
```
evacchi, line 194: I think we don't need to parse these into Float bits in this case, the raw integer will do

test-gen-plugin/src/main/java/com/dylibso/chicory/maven/JavaTestGen.java
```diff
@@ -370,10 +381,15 @@ private List<Expression> generateAssert(String varName, Command cmd) {
...
+                                .collect(Collectors.toList())
+                        : List.<String>of();
+
+        var argSize =
+                (cmd.action().args() != null)
+                        ? Arrays.stream(cmd.action().args()).mapToInt(a -> a.type().size()).sum()
+                        : 0;
 
-        var invocationMethod = ".apply(" + args + ")";
+        var invocationMethod = ".apply(" + String.join(",", args) + ")";
```
evacchi, line 392: notice, this is _not_ correct, but works for now. It should be adapted to do same thing as the lines below, but it required a little more refactoring, and I wanted to push this out quickly to unblock @mfirry

evacchi:
here's an example of a test (`simd_bitwise`) after the change:

```java
    @Test()
    @Order(9)
    @DisplayName("simd_bitwise.wast:31")
    public void test9() {
        ExportFunction varAnd = testModule0Instance.export("and");
        var args = new MStack();
        args.pushAll(i32ToVec( new long[] { Long.parseUnsignedLong("0"), Long.parseUnsignedLong("0"), Long.parseUnsignedLong("4294967295"), Long.parseUnsignedLong("4294967295") }));
        args.pushAll(i32ToVec( new long[] { Long.parseUnsignedLong("0"), Long.parseUnsignedLong("4294967295"), Long.parseUnsignedLong("0"), Long.parseUnsignedLong("4294967295") }));
        var results = varAnd.apply(args.slice(0, 4));
        var expected = new long[] {Long.parseUnsignedLong("0"), Long.parseUnsignedLong("0"), Long.parseUnsignedLong("0"), Long.parseUnsignedLong("4294967295") };
        assertArrayEquals(expected, vecTo32(results));
    }
```

vs before:

```java
    @Test()
    @Order(9)
    @DisplayName("simd_bitwise.wast:31")
    public void test9() {
        ExportFunction varAnd = testModule0Instance.export("and");
        var results = varAnd.apply(new long[] { 0, 0, 4294967295, 4294967295 }, new long[] { 0, 4294967295, 0, 4294967295 });
        var expected = new long[] {Long.parseUnsignedLong("0"), Long.parseUnsignedLong("0"), Long.parseUnsignedLong("0"), Long.parseUnsignedLong("4294967295") };
        assertArrayEquals(expected, vecTo32(results));
    }

```

notice:
- `varAnd.apply()` took 2 long arrays (incorrect)
- we wrap the two arg vectors into`i32ToVec()`, because even though it's 4 longs each, they should be treated as i32x4 consts, i.e. 4 integers for a v128 value
- we avoid having to concat the 2 resulting `long[2]` arrs by pushing to a stack instead
- `4294967295` is an integer literal so it errors out because it won't fit (you can add `L`, but it's more correct to treat it as unsigned)

notice that it's not strictly necessary to modify MStack, we can also implement and ad-hoc stack or utility class (e.g. "TestArgs") only for tests

andreaTP:
Updated to use an `ArgsAdapter` instead of `MStack`, we need to progress in arguments and local size tracking, I started here:
https://github.com/dylibso/chicory/commit/fa6bea1ce8c8263f22f84f18bdbca5817bbc26d2

for anyone that want to take over the work.

test-gen-plugin/src/main/java/com/dylibso/chicory/maven/wast/WasmValueType.java
```diff
@@ -25,6 +25,10 @@ public enum WasmValueType {
         this.value = value;
     }
 
+    public int size() {
```
andreaTP, line 28: this is not used anymore

test-gen-plugin/src/main/java/com/dylibso/chicory/maven/wast/WasmValue.java
```diff
@@ -165,11 +186,14 @@ public String toExpectedValue() {
                             case I32:
                                 sb.append("Long.parseUnsignedLong(\"" + v + "\")");
                                 break;
+                            case I64:
```
andreaTP, line 189: nitpick - some of those cases can be collapsed, isn't it?

andreaTP (approved):
A couple of nitpicks but this is ready to go, thanks :+1:

test-gen-plugin/src/main/java/com/dylibso/chicory/maven/TestGenMojo.java
```diff
@@ -179,9 +179,9 @@ public void execute() throws MojoExecutionException {
                     includedExcludedWasts.add(includedWast);
                 }
             }
+            // TODO: this mechanism fails when there are no excluded wast in one profile only
```
andreaTP, line 182: Nothing critical, but FYI @electrum , personally I think it's fine to relax those check as most of the spec have already been included and we will still review the incoming PRs.

andreaTP (approved):
LGTM, this is ready to go, thanks @evacchi !

PR #692, opened 2024-12-05
Add an instance factory to let customize the instantiation on Store

andreaTP:
The current convenience helper method `instantiate` in `Store` doesn't leave room for customizations done with the `Instance.Builder`, such as `withStart` or `withMachineFactory`.

In this PR we add a more generic helper that let the user configure the `Instance` and we base the current helper on it.

### DIFF
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/Store.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/Store.java
@@ -6,6 +6,7 @@ import com.dylibso.chicory.wasm.types.ExportSection;
 import com.dylibso.chicory.wasm.types.FunctionType;
 import java.util.LinkedHashMap;
 import java.util.Objects;
+import java.util.function.Function;
 
 /**
  * The runtime storage for all function, global, memory, table instances.
@@ -129,8 +130,16 @@ public class Store {
      * A shorthand for instantiating a module and registering it in the store.
      */
     public Instance instantiate(String name, WasmModule m) {
+        return this.instantiate(
+                name, imports -> Instance.builder(m).withImportValues(imports).build());
+    }
+
+    /**
+     * A shorthand for instantiating a module and registering it in the store.
+     */
+    public Instance instantiate(String name, Function<ImportValues, Instance> instanceBuilder) {
         ImportValues importValues = this.toImportValues();
-        Instance instance = Instance.builder(m).withImportValues(importValues).build();
+        Instance instance = instanceBuilder.apply(importValues);
         register(name, instance);
         return instance;
     }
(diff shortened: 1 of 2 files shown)


runtime/src/main/java/com/dylibso/chicory/runtime/Store.java
```diff
@@ -129,8 +130,16 @@ public Store register(String name, Instance instance) {
...
      */
     public Instance instantiate(String name, WasmModule m) {
+        return this.instantiate(
+                name, imports -> Instance.builder(m).withImportValues(imports).build());
+    }
+
+    /**
+     * A shorthand for instantiating a module and registering it in the store.
+     */
+    public Instance instantiate(String name, Function<ImportValues, Instance> instanceBuilder) {
```
evacchi, line 140: you could pass directly an `Instance.Builder` though

evacchi (commented):
from the PR and the test case, it is not clear what your final goal is. I assume (from off-line conversations) it might be related to AoT, but you should also update the top post so we won't forget :)

runtime/src/main/java/com/dylibso/chicory/runtime/Store.java
andreaTP: I entertained the idea, and I think that giving max freedom to the user is the way to go to avoid wrapping stuff.

andreaTP:
In general the idea is to avoid having to duplicate the options offered by `Instance.Builder`.

I have been contacted regarding `withStart`, but I see the real value of using it with `withMachineFactory`.

> you should also update the top post

do you have a proposal?

runtime/src/main/java/com/dylibso/chicory/runtime/Store.java
evacchi: maybe let's rename the var from `instanceBuilder` to `instanceFactory` to avoid confusion.

runtime/src/main/java/com/dylibso/chicory/runtime/Store.java
```diff
@@ -129,8 +130,16 @@ public Store register(String name, Instance instance) {
      * A shorthand for instantiating a module and registering it in the store.
      */
     public Instance instantiate(String name, WasmModule m) {
+        return this.instantiate(
+                name, imports -> Instance.builder(m).withImportValues(imports).build());
+    }
+
+    /**
+     * A shorthand for instantiating a module and registering it in the store.
```
evacchi, line 138:
```suggestion
     * Creates an instance with the given factory and registers the result in the store.
```

evacchi (approved):
only left 1 minor suggestion to the javadoc, and another comment to the method signature. Otherwise it's fine, thanks.

runtime/src/main/java/com/dylibso/chicory/runtime/Store.java
andreaTP: done :+1:

PR #696, opened 2024-12-07
Allow overriding memory limits when constructing instance

### LINKED ISSUE #606: MemoryLimits: allow to impose constraints to an instance
Modules are able to define their own memory constraints, but one useful feature of Wasm engines is to allow the host to control the use of resources of the guests.

We should allow end-users to refine the memory constraints of a Wasm module so that 
1) we can impose finer restrictions on modules that have been compiled with default values (i.e. unrestricted limits)
2) rogue instances cannot grow their memory indefinitely

- One solution could be to provide a `.withMemoryLimits()` on the `InstanceBuilder`.
- Another, more general solution would be to provide `.withMemoryFactory()` for other use cases (e.g. experiment with alternative allocation strategies, such as off-heap/MemorySegment)
In this case, however, it would be also useful to define a specific, built-in factory that allows to easily constrain memory limits; e.g. `.withMemoryFactory(MemoryLimits.of(...))`

electrum:
This removes the unused `shared` flag from `MemoryLimits` as that is part of the threads proposal which is not yet standardized. That flag is for validating memory imports, so it's not clear if `MemoryLimits` is the correct place for it. We can revisit that if we implement threads.

Fixes #606

### DIFF
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/Instance.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/Instance.java
@@ -6,6 +6,7 @@ import static com.dylibso.chicory.wasm.types.ExternalType.FUNCTION;
 import static com.dylibso.chicory.wasm.types.ExternalType.GLOBAL;
 import static com.dylibso.chicory.wasm.types.ExternalType.MEMORY;
 import static com.dylibso.chicory.wasm.types.ExternalType.TABLE;
+import static java.util.Objects.requireNonNullElse;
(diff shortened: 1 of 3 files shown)


evacchi (approved):
I am in favor of this approach, but I know @andreaTP wanted a flexible factory-based API. So it's an approve from me, let's see what his opinion is.

andreaTP (approved):
> wanted a flexible factory-based API.

That's accurate, still, I don't have any disagreement with the changes provided here, thanks, LGTM.

PR #698, opened 2024-12-09
Add MemoryFactory and make Memory an interface

andreaTP:
Follow up from #696
This little utility allows for experimenting, fully in user land, with offheap and unsafe memory implementations.

### DIFF
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/ByteBufferMemory.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/ByteBufferMemory.java
@@ -18,7 +18,7 @@ import java.util.function.Function;
  * Represents the linear memory in the Wasm program. Can be shared
  * reference b/w the host and the guest.
  */
-public class ByteBufferMemory implements Memory {
+public final class ByteBufferMemory implements Memory {
 
     private final MemoryLimits limits;
 
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/ByteBufferMemory.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/ByteBufferMemory.java
@@ -1,0 +1,394 @@
+package com.dylibso.chicory.runtime;
+
+import static com.dylibso.chicory.runtime.ConstantEvaluators.computeConstantValue;
+import static java.lang.Math.min;
+
+import com.dylibso.chicory.wasm.ChicoryException;
+import com.dylibso.chicory.wasm.UninstantiableException;
+import com.dylibso.chicory.wasm.types.ActiveDataSegment;
+import com.dylibso.chicory.wasm.types.DataSegment;
+import com.dylibso.chicory.wasm.types.MemoryLimits;
+import com.dylibso.chicory.wasm.types.PassiveDataSegment;
+import java.nio.BufferOverflowException;
+import java.nio.BufferUnderflowException;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+
+/**
+ * Represents the linear memory in the Wasm program. Can be shared
+ * reference b/w the host and the guest.
+ */
+public class ByteBufferMemory implements Memory {
+
+    private final MemoryLimits limits;
+
+    private DataSegment[] dataSegments;
+
+    private ByteBuffer buffer;
+
+    private int nPages;
+
+    public ByteBufferMemory(MemoryLimits limits) {
+        this.limits = limits;
+        this.buffer = allocateByteBuffer(PAGE_SIZE * limits.initialPages());
+        this.nPages = limits.initialPages();
+    }
+
+    private static ByteBuffer allocateByteBuffer(int capacity) {
+        return ByteBuffer.allocate(capacity).order(ByteOrder.LITTLE_ENDIAN);
+    }
+
+    /**
+     * Gets the size of the memory in number of pages
+     */
+    @Override
+    public int pages() {
+        return nPages;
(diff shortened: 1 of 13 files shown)


electrum (approved):
This change looks good, but it's not useful until we convert `Memory` to be an interface.

andreaTP:
> not useful until we convert `Memory` to be an interface

I missed it's a `final` class, let me follow up :+1: thanks!

andreaTP:
I moved `Memory` to be an interface, the default implementation is now `ByteBufferMemory` and is not anymore a `final` class to let people extend it when they know what they are doing ...

electrum:
Why do we mix exception types in `Memory`? It actually seems a bit random which type is used for a given operation.

electrum (approved):
This change looks good to me. But we should follow up on why we use both `WasmRuntimeException` and `UninstantiableException`.

runtime/src/main/java/com/dylibso/chicory/runtime/ByteBufferMemory.java
```diff
@@ -0,0 +1,394 @@
...
+import java.nio.ByteOrder;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+
+/**
+ * Represents the linear memory in the Wasm program. Can be shared
+ * reference b/w the host and the guest.
+ */
+public class ByteBufferMemory implements Memory {
```
electrum, line 24: Make this final?

runtime/src/main/java/com/dylibso/chicory/runtime/Memory.java
```diff
@@ -1,363 +1,106 @@
...
 
     /**
      * A WebAssembly page size is 64KiB = 65,536 bytes.
      */
-    public static final int PAGE_SIZE = 65536;
+    int PAGE_SIZE = 65536;
 
     /**
      * Maximum number of pages allowed by the runtime.
      * WASM supports 2^16 pages, but we must limit based on the maximum JVM array size.
```
electrum, line 15: Is this actually a limit of `ByteBufferMemory`? With a different implementation of `Memory` that is backed by `MemorySegment`, segmented arrays, etc., are there any other places where `Integer.MAX_VALUE` would bite us?

runtime/src/main/java/com/dylibso/chicory/runtime/Memory.java
```diff
@@ -1,363 +1,106 @@
...
+
+    void writeCString(int offset, String str);
+
+    void writeCString(int offset, String str, Charset charSet);
+
+    String readCString(int addr, Charset charSet);
+
+    String readCString(int addr);
+
+    void write(int addr, byte[] data);
```
electrum, line 48: It would be better to make these default methods on the interface, as it's unlikely that other implementations of the interface will need to customize them.

runtime/src/main/java/com/dylibso/chicory/runtime/Memory.java
```diff
@@ -1,363 +1,106 @@
...
+
+    void writeF64(int addr, double data);
+
+    double readDouble(int addr);
+
+    long readF64(int addr);
+
+    void zero();
+
+    void fill(byte value);
```
electrum, line 98: I think these can be removed. The single call to `zero()` in `Instance.initialize()` doesn't seem needed. We can make the contract that `Memory` is filled with zero when it is created (which is what any sane implementation would do anyway). Then we don't need to redundantly fill it with zero again after creating it.

runtime/src/test/java/com/dylibso/chicory/runtime/WasmModuleTest.java
```diff
@@ -206,7 +206,13 @@ public void shouldSupportMemoryFactoryOverride() {
                 .withMemoryFactory(
                         limits -> {
                             memoryCreated.set(true);
-                            return new Memory(limits);
+                            return new ByteBufferMemory(limits) {
```
electrum, line 209: I think it would be better to make `ByteBufferMemory` final

runtime/src/main/java/com/dylibso/chicory/runtime/Memory.java
andreaTP: It's defined in the spec [here](https://webassembly.github.io/spec/core/exec/runtime.html#memory-instances)

Screenshot from 2024-12-17 11-53-18

runtime/src/test/java/com/dylibso/chicory/runtime/WasmModuleTest.java
andreaTP: What's the concern in letting people extend `ByteBufferMemory`?
Exposing too many implementation details?

runtime/src/main/java/com/dylibso/chicory/runtime/Memory.java
andreaTP: removed :+1:

runtime/src/main/java/com/dylibso/chicory/runtime/ByteBufferMemory.java
andreaTP: This should be the only "open" discussion, I'm not 100% convinced we should make it `final` as it seem appealing to just extend instead of having to spin up a full-blown implementation all the time.
But happy to listen to a different POV!

runtime/src/main/java/com/dylibso/chicory/runtime/Memory.java
andreaTP: All the implementations relying on other methods in the interface are now `default`.

evacchi (approved):

PR #699, opened 2024-12-09
Add a note on proposals implementation priority

### LINKED ISSUE #673: Support WASI P2
I noticed support for WASI is P1 now, but P2 has been out for a little while now. It would be awesome to have support for P2 and the component model!

andreaTP:
Fix #673

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -85,6 +85,10 @@ Basic steps:
 
 note: if you're working using a *corporate proxy* (or anything like this), you might need to pass the usual `-Dhttps.proxyHost=...` and `-Dhttps.proxyPort=...` in order to properly instruct Maven about this (this can be required for example for `test-gen-plugin` since it downloads the testsuite).
 
+### Proposals implementation
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -85,6 +85,10 @@ Basic steps:
 
 note: if you're working using a *corporate proxy* (or anything like this), you might need to pass the usual `-Dhttps.proxyHost=...` and `-Dhttps.proxyPort=...` in order to properly instruct Maven about this (this can be required for example for `test-gen-plugin` since it downloads the testsuite).
 
+### Proposals implementation
+
+Our priority is to focus on implementing [proposals](https://github.com/WebAssembly/proposals) that are in the most advanced stages of development. While we wholeheartedly encourage and support explorations, we’ll be dedicating less time to early-stage proposals until we have more comprehensive support for those that are stabilized.
```
bhelx, line 90: Maybe we note that we encourage experimentation in a fork or separate module? i think this is enough though if you want to merge.

andreaTP: I think this is implicitly explicit enough :slightly_smiling_face:  let's iterate according to the feedback we will receive.

PR #703, opened 2024-12-10
Remove BitSet usage from the public API of Parser

### LINKED ISSUE #605: Parser: document or update `BitSet` usage.
`Parser` allows to pass in a `BitSet` instance in its constructor to select the sections that should be parsed. 

However, there is no documentation on what the BitSet represents, and how it should be built. Either document how to build a proper BitSet, or wrap it in an explicit builder.

Besides, there does not seem to be usages in the codebase, so it might be a candidate for removal (or at least, to make it package-private)

e.g.

```java
class ParserOptions {
    BitSet includedSections;
    ParserOptions withImportSection() {
         includedSections.set(SectionId.IMPORT);
         return this;
   }
   // ... etc
}
class Parser {
    public Parser(ParserOptions opts) {
```

andreaTP:
Fix #605

### DIFF
--- a/wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java
+++ b/wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java
@@ -84,17 +84,16 @@ public final class Parser {
     private final Map<String, Function<byte[], CustomSection>> customParsers;
     private final BitSet includeSections;
 
-    public Parser() {
-        this(new BitSet());
+    private static final Map<String, Function<byte[], CustomSection>> DEFAULT_CUSTOM_PARSERS =
+            Map.of("name", NameCustomSection::parse);
+
+    private Parser() {
+        this(null, DEFAULT_CUSTOM_PARSERS);
     }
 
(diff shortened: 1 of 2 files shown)


wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java
```diff
@@ -156,6 +155,41 @@ private static void onSection(WasmModule.Builder module, Section s) {
...
+    public static final class Builder {
+        private Map<String, Function<byte[], CustomSection>> customParsers;
+        private BitSet includeSections;
+
+        private Builder() {}
+
+        /*
+         * @param sectionId : the sectionId to be included while parsing, e.g. SectionId.MEMORY
+         */
+        public Builder addIncludedSectionId(int sectionId) {
```
evacchi, line 171: maybe?

```suggestion
        public Builder includeSectionId(int sectionId) {
```

evacchi (approved):
minor naming suggestion, otherwise fine 👍

Issue #695, opened 2024-12-06
[SIMD]: Use Java Vector API for SIMD operations


mfirry:
This issue is mainly for me to keep track of the current _porting_ of SIMD-related operations to leverage on Java Vector API.
I'll fork, branch out and little by little push my work. Suggestions, advices and ideas are more than welcome.

- [x] I8x16_EQ
- [x] I8x16_ADD
- [x] I8x16_SUB
- [x] I8x16_SHL
- [x] I8x16_ALL_TRUE
- [x] F32x4_MUL
- [x] F32x4_ABS
- [x] F32x4_MIN
- [x] V128_NOT
- [x] V128_BITSELECT
- [x] I8x16_EXTRACT_LANE_S
- [ ] F32x4_CONVERT_I32x4_U
- [ ] I32x4_TRUNC_SAT_F32x4_S
- [ ] I8x16_SWIZZLE

andreaTP:
Yes @wendigo , I'm planning to release a 1.1 by the end of the week.
Please consider that Simd support is only in the interpreter at the moment.

andreaTP:
released in `1.1.0`

PR #706, opened 2024-12-12
[SIMD]: Java Vector API usage

mfirry:
Addresses #695

Thi is the first batch of functionalities leveraging on Java Vector API.

### DIFF
--- a/simd/src/main/java/com/dylibso/chicory/simd/SimdInterpreterMachine.java
+++ b/simd/src/main/java/com/dylibso/chicory/simd/SimdInterpreterMachine.java
@@ -107,76 +107,57 @@ public final class SimdInterpreterMachine extends InterpreterMachine {
         stack.push(result);
     }
 
+    /**
+     * Compares two 128-bit vectors as if they were two vectors of 16 eight-bit integers.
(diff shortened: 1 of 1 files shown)


simd/src/main/java/com/dylibso/chicory/simd/SimdInterpreterMachine.java
```diff
@@ -95,112 +96,84 @@ private static void V128_LOAD(MStack stack, Instance instance, Operands operands
...
-        var val2High = stack.pop();
-        var val2Low = stack.pop();
+        var high = stack.pop();
+        var low = stack.pop();
 
-        long resultLow = 0L;
-        long resultHigh = 0L;
+        var v1 =
+                LongVector.fromArray(
+                                LongVector.SPECIES_128, new long[] {low, high}, stack.size() - 2)
```
andreaTP, line 131: The offset here should be `0`

andreaTP (changes requested):
Thanks @mfirry for bringing this forward!
One single blocker and a just nitpicks we chatted about.

andreaTP (approved):
LGTM, great job @mfirry !

PR #707, opened 2024-12-12
Add `hashCode` and `equals` implementations to the data classes in the `wasm` module

andreaTP:
This allow the data class `WasmModule` to be safely stored in `HashMaps` and similar data structures.

### DIFF
--- a/wasm/src/main/java/com/dylibso/chicory/wasm/WasmModule.java
+++ b/wasm/src/main/java/com/dylibso/chicory/wasm/WasmModule.java
@@ -20,6 +20,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 
 public final class WasmModule {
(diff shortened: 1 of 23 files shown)


evacchi (approved):
looks good, although maybe we could be smarter and just record a digest of the original binary? maybe a future update :)

andreaTP:
> maybe we could be smarter and just record a digest of the original binary?

With the digest it becomes hard to avoid comparing things like `CustomSection`, but let's speak about :+1:

evacchi:
I am not sure you want to avoid that? but sure let's talk about this offline

electrum:
This an issue if you want it to be the key of a map or deduplicated in a set. Would anyone do that? It’s safe to store as the value of a map or in a list (as with any object).

andreaTP:
@electrum that's exactly the use case I want to enable here.
@evacchi is building a cache for "pre compiled"(with aot) modules.

His first "go to" was to use `WasmModule` as the key in an `HashMap`,  since those are all "data classes" I think the expectation is, maybe a bit high, but fair.

Also I don't have any good reason/use case where, keeping the reference, is an advantage.

evacchi:
TBH I don't know if it would be frequent to try to rebuild different instances of the same module for my use case

andreaTP:
Something is going wrong with GH here
Screenshot from 2024-12-17 11-46-40

Let's try to close and re-open ...

PR #714, opened 2024-12-17
[simd]: F32x4_ABS and F32x4_MIN added.

mfirry:
F32x4_ABS now uses Java Vector API

### DIFF
--- a/simd/src/main/java/com/dylibso/chicory/simd/SimdInterpreterMachine.java
+++ b/simd/src/main/java/com/dylibso/chicory/simd/SimdInterpreterMachine.java
@@ -234,15 +234,15 @@ public final class SimdInterpreterMachine extends InterpreterMachine {
     }
 
     private static void F32x4_ABS(MStack stack) {
-        var valHigh = stack.pop();
-        var valLow = stack.pop();
+        var offset = stack.size() - 2;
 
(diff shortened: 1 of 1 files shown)


simd/src/main/java/com/dylibso/chicory/simd/SimdInterpreterMachine.java
```diff
@@ -234,15 +234,15 @@ private static void F32x4_MUL(MStack stack) {
...
-        long resultHigh = valHigh & (1L << 31 | 1L << 63) ^ 0xFFFFFFFFFFFFFFFFL;
+        var v =
+                LongVector.fromArray(LongVector.SPECIES_128, stack.array(), offset)
+                        .reinterpretAsFloats();
 
-        stack.push(resultLow);
-        stack.push(resultHigh);
+        var result = v.abs().reinterpretAsLongs().toArray();
+
+        System.arraycopy(result, 0, stack.array(), 0, 2);
```
andreaTP, line 245: Sadly, this sneaked in with other opcodes too, do you mind fixing them all?

This line should be:

```java
System.arraycopy(result, 0, stack.array(), offset, 2);
```

The tests are passing, probably, because we don't have many elements on the stack.
I'd be in favor of refactoring into a method `overrideLastV128`(not a great name though ...) or similar.

andreaTP (changes requested):
Thanks @mfirry , one correction that I missed noticing in other places and should be good to go :+1:

andreaTP (approved):
LGTM, thanks!

PR #530, opened 2024-09-18
[demo] Run ubuntu on Chicory thanks to c2w

andreaTP:
This is just a demo, I stripped down a bit the needed things, and the CI demonstrates that a container compiled with [container2wasm](https://github.com/ktock/container2wasm) .

Happy to hand over this spike to anyone interested in speaking/blogging/whatever :slightly_smiling_face:

PR #716, opened 2024-12-18
Implement WASI poll_oneoff

### LINKED ISSUE #592: Implement `poll_oneoff` in WASI P1
Zig, the Go standard library, Rust(in some cases) and c2w require `poll_oneoff`.
It would be great to have a stable implementation of it.

Spike ref:
https://github.com/dylibso/chicory/pull/530/files#diff-9037d84891e8043e756bfd63e2101686b67d1cad2cc2c4af3b2e6b48675acdf3R1086

electrum:
Fixes #592

### DIFF
--- a/wasi-tests/src/test/java/com/dylibso/chicory/wasi/WasiPreview1Test.java
+++ b/wasi-tests/src/test/java/com/dylibso/chicory/wasi/WasiPreview1Test.java
@@ -4,9 +4,12 @@ import static com.dylibso.chicory.wasi.Files.copyDirectory;
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static java.nio.file.Files.createDirectory;
 import static java.nio.file.Files.createSymbolicLink;
+import static java.nio.file.Files.writeString;
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import static org.junit.jupiter.api.Assertions.assertArrayEquals;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import com.dylibso.chicory.log.Logger;
 import com.dylibso.chicory.log.SystemLogger;
@@ -28,7 +31,9 @@ import java.nio.file.Path;
 import java.util.List;
 import java.util.Random;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.Timeout;
 
+@Timeout(10)
 public class WasiPreview1Test {
     private final Logger logger = new SystemLogger();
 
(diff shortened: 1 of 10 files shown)


wasi/src/main/java/com/dylibso/chicory/wasi/WasiPreview1.java
```diff
@@ -1339,9 +1338,122 @@ public int pathUnlinkFile(int dirFd, @Buffer String rawPath) {
...
+                break;
+            }
+
+            long elapsed = System.nanoTime() - start;
+            long duration = max(minTimeout, 0) - elapsed;
+            if (!readEvents.isEmpty()) {
+                duration = min(duration, MILLISECONDS.toNanos(100));
+            }
+            try {
+                NANOSECONDS.sleep(duration);
```
andreaTP, line 1436: Looking at the implementation of the method, I think we can use the underlying:

```java
Thread.sleep(0, duration);
```

wasi/src/main/java/com/dylibso/chicory/wasi/WasiPreview1.java
```diff
@@ -1339,9 +1338,122 @@ public int pathUnlinkFile(int dirFd, @Buffer String rawPath) {
...
+                    memory.writeLong(outPtr + 16, available);
+                    outPtr += 32;
+                    nevents++;
+                } catch (IOException e) {
+                    writeEvent(memory, outPtr, userData, WasiEventType.FD_READ, WasiErrno.EIO);
+                    outPtr += 32;
+                    nevents++;
+                }
+            }
+            if (nevents > 0) {
```
andreaTP, line 1426: What's the reference implementation you started from?

I cannot find this condition in the [wazero implementation](https://github.com/tetratelabs/wazero/blob/610c202ec48f3a7c729f2bf11707330127ab3689/imports/wasi_snapshot_preview1/poll.go#L161).

wasi/src/main/java/com/dylibso/chicory/wasi/WasiPreview1.java
```diff
@@ -1339,9 +1338,122 @@ public int pathUnlinkFile(int dirFd, @Buffer String rawPath) {
...
+
+            long elapsed = System.nanoTime() - start;
+            long duration = max(minTimeout, 0) - elapsed;
+            if (!readEvents.isEmpty()) {
+                duration = min(duration, MILLISECONDS.toNanos(100));
+            }
+            try {
+                NANOSECONDS.sleep(duration);
+            } catch (InterruptedException e) {
+                throw new ChicoryException("Thread interrupted");
```
andreaTP, line 1438:
```suggestion
                throw new ChicoryException("Thread interrupted", e);
```

wasi/src/main/java/com/dylibso/chicory/wasi/WasiPreview1.java
```diff
@@ -1339,9 +1338,122 @@ public int pathUnlinkFile(int dirFd, @Buffer String rawPath) {
...
+                    break;
+                default:
+                    return wasiResult(WasiErrno.EINVAL);
+            }
+            inPtr += 32;
+        }
+
+        long minTimeout = clockEvents.stream().mapToLong(Entry::getKey).min().orElse(0);
+        long start = System.nanoTime();
+        while (true) {
```
andreaTP, line 1407: A `while (true)` with so much code afterwords make me slightly uncomfortable.

In this case, if I'm not missing an invariant:

- `nsubscriptions` == 0
- `readEvents` is an empty List
- `clockEvents` is an empty List
- the loop doesn't end

is it the expected behavior?

andreaTP (commented):
Thanks for taking a stab at this!
A few comments to better understand what's going on.

Have you tested this implementation with actual code using it?
Go, Zig and even Rust should produce this call in some circumstances ...

wasi/src/main/java/com/dylibso/chicory/wasi/WasiPreview1.java
evacchi: +1 I think just a few test cases would be useful to clarify the behavior :) 

there's a few here 
- unit tests https://github.com/tetratelabs/wazero/blob/610c202ec48f3a7c729f2bf11707330127ab3689/imports/wasi_snapshot_preview1/poll_test.go
- integration:
  - https://github.com/tetratelabs/wazero/blob/610c202ec48f3a7c729f2bf11707330127ab3689/imports/wasi_snapshot_preview1/wasi_stdlib_test.go#L275
  - https://github.com/tetratelabs/wazero/blob/610c202ec48f3a7c729f2bf11707330127ab3689/imports/wasi_snapshot_preview1/wasi_stdlib_test.go#L501

(it doesn't need to be all, because this is not really covering all of them, just a source of inspiration)

wasi/src/main/java/com/dylibso/chicory/wasi/WasiPreview1.java
electrum: This doesn't work as nanoseconds must be less than one second. Also, it's less clear, as the reader needs to be familiar with the method to know that the second argument is nanoseconds.

wasi/src/main/java/com/dylibso/chicory/wasi/WasiPreview1.java
electrum: Thanks, I missed that

wasi/src/main/java/com/dylibso/chicory/wasi/WasiPreview1.java
electrum: I wrote this myself based on the specification. The wazero implementation seems incorrect as it writes out all clock events, when per the specification it should only be the clocks that reached the timeout:

> The time value of clock `subscription_clock::id` has reached timestamp `subscription_clock::timeout`.

This condition is actually incorrect, as it will prevent writing out clocks if there were any triggered descriptors. Instead, I'll change loop to skip sleeping if it already wrote events.

wasi/src/main/java/com/dylibso/chicory/wasi/WasiPreview1.java
electrum: You're right, I had a typo on the check at the start. It should be

> If `nsubscriptions` is 0, returns `errno::inval`.

So we will never enter the loop in that case.

electrum:
I verified the implementation with Python, which uses this function for some modules such as `socket` (which fail later on but call this function first).

wasi/src/main/java/com/dylibso/chicory/wasi/WasiPreview1.java
andreaTP: > I wrote this myself based on the specification.

Can you add a comment with the relevant (perma-) link?

wasi/src/main/java/com/dylibso/chicory/wasi/WasiPreview1.java
andreaTP: Thanks for updating this, now it looks safer, in all honesty, I still need to squint a lot my eyes to understand what's going on.

Adding unit tests for the implementation is an option to better clarify the expected behavior, does it sound feasible without a major amount of additional work?

andreaTP:
> I verified the implementation with Python

Thanks!
Before merging, for diligence, we should test on one additional language.

On top of my head, those are the easiest to try:
- Go: [`c2w`](https://github.com/dylibso/chicory/pull/530)
- Zig: [testsuite](https://github.com/andreaTP/chicory/tree/zig-testsuite)

wasi/src/main/java/com/dylibso/chicory/wasi/WasiPreview1.java
electrum: I added tests for all cases except for invalid input and `IOException`, and verified this with code coverage in the IDE. I also added comments which hopefully makes the logic easier to follow.

electrum:
Agreed, I'll test both of those above.

wasi/src/main/java/com/dylibso/chicory/wasi/WasiPreview1.java
```diff
@@ -194,12 +197,8 @@ public int clockTimeGet(Memory memory, int clockId, long precision, int resultPt
         logger.tracef("clock_time_get: [%s, %s, %s]", clockId, precision, resultPtr);
         switch (clockId) {
             case WasiClockId.REALTIME:
-                Instant now = clock.instant();
-                long epochNanos = SECONDS.toNanos(now.getEpochSecond()) + now.getNano();
-                memory.writeLong(resultPtr, epochNanos);
-                return wasiResult(WasiErrno.ESUCCESS);
             case WasiClockId.MONOTONIC:
-                memory.writeLong(resultPtr, System.nanoTime());
+                memory.writeLong(resultPtr, clockTime(clockId));
```
andreaTP, line 201: Would it be possible to use `clockTime`(or something similar) everywhere instead of `System.nanotime()`?

wasi-tests/src/test/java/com/dylibso/chicory/wasi/WasiPreview1Test.java
```diff
@@ -294,6 +299,134 @@ public void wasiReadLink() throws IOException {
...
+                assertEquals(WasiEventType.FD_WRITE, memory.read(out + 10));
+            }
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    @Test
+    public void wasiPollOneoffClockAbstime() {
+        long deadline = System.nanoTime() + MILLISECONDS.toNanos(25);
```
andreaTP, line 407: Instead of using `System.nanotime` it would be great to use a "mocked" clock.
i.e.:

- set a clock source that can be controlled from the test
- trigger `pollOneOff` and verify that the deadline has not been reached
- increase the clock to be > than the expected deadline
- verify that the deadline is reached

Do you think it's doable?

andreaTP (commented):
Great improvements here, thanks @electrum for addressing the comments, it reads much better now!

One additional note on the use of `System.nanoTime` and I'll wait for the results of running Go and/or Zig :slightly_smiling_face:

wasi-tests/src/test/java/com/dylibso/chicory/wasi/WasiPreview1Test.java
electrum: We support mocking for the realtime `Clock` via `WasiOptions`. However, I don't think that simply mocking `System.nanoTime()` is sufficient, as it interacts with the way we use `Thread.sleep()`.

Remember that `poll_oneoff()` waits until there is at least one event. So in order to test that a clock event doesn't trigger, we need a different event to trigger. But then we won't actually sleep, unless we mock `stdin` with something that starts off unreadable and becomes readable later.

I can look at this more if you feel strongly, but I think the testing we have is sufficient, especially for something that likely won't be used for more than a basic sleep function.

wasi/src/main/java/com/dylibso/chicory/wasi/WasiPreview1.java
electrum: This method is a helper method that is shared with `clock_time_get()` to get the realtime or monotonic clock. The other usages of `System.nanoTime()` are in `poll_oneoff()` to measure how long we have been sleeping. As I mentioned in the other comment, mocking that seems insufficient as we directly call `Thread.sleep()`. I'd need to think more about how those interact.

wasi/src/main/java/com/dylibso/chicory/wasi/WasiPreview1.java
andreaTP: :+1:

wasi/src/main/java/com/dylibso/chicory/wasi/WasiPreview1.java
electrum: This is just the specification for WASI. We could add links for all the functions as a separate commit if you think that is valuable.

electrum:
The `c2w` test works. The Zig test runs until this point:
```
716/2552 test.GUID... OK
717/2552 test.chdir smoke test... SKIP
718/2552 test.open smoke test... SKIP

com.dylibso.chicory.wasm.UninstantiableException: Trapped on unreachable instruction

	at com.dylibso.chicory.runtime.Instance.initialize(Instance.java:159)
	at com.dylibso.chicory.runtime.Instance.<init>(Instance.java:100)
	at com.dylibso.chicory.runtime.Instance$Builder.build(Instance.java:722)
	at com.dylibso.chicory.wasi.WasiPreview1Test.shouldRunZigStdlibTestsuite(WasiPreview1Test.java:239)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1597)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1597)
Caused by: com.dylibso.chicory.runtime.TrapException: Trapped on unreachable instruction
	at com.dylibso.chicory.runtime.InterpreterMachine.eval(InterpreterMachine.java:150)
	at com.dylibso.chicory.runtime.InterpreterMachine.call(InterpreterMachine.java:81)
	at com.dylibso.chicory.runtime.InterpreterMachine.CALL(InterpreterMachine.java:1637)
	...
```

andreaTP (approved):
> I can look at this more if you feel strongly, but I think the testing we have is sufficient, especially for something that likely won't be used for more than a basic sleep function.

I agree with this evaluation, it would be nice to have more isolated tests not using global resources like `nanoTime` and `sleep`, but the current implementation and testing is definitely sufficient.

LGTM, thanks for the effort and bearing with me.

Additional notes:
- yes please, add the link to the WASI spec you are looking at (I'm always confused about the right link)
- I recall the zig testsuite to progress a bit more, but it can be a number of factors ...

PR #719, opened 2024-12-19
Fix locals handling for V128 values

andreaTP:
Leaving this in draft as I'm not adding tests to actually cover the changes.

@mfirry this is eventually needed to continue with the SIMD work, feel free to take over whenever the times come.


runtime/src/main/java/com/dylibso/chicory/runtime/StackFrame.java
```diff
@@ -25,31 +27,61 @@ public class StackFrame {
     private final int funcId;
     private int pc;
     private final long[] locals;
+    private final List<ValueType> localTypes;
+    private final int[] localIdx;
```
evacchi, line 31: good idea

evacchi (approved):
looks good! ( well, tests aside :) ) @mfirry check this out to go on

andreaTP:
Ready for review/merge as I added a bunch of additional specs relying on the missing bits.

runtime/src/main/java/com/dylibso/chicory/runtime/GlobalInstance.java
```diff
@@ -15,13 +15,23 @@ public GlobalInstance(Value value) {
     }
 
     public GlobalInstance(Value value, MutabilityType mutabilityType) {
-        this.value = value.raw();
+        this.values = new long[] {value.raw()};
```
evacchi, line 18: so, even though I don't want to block, I am tempted to suggest using a pair of long fields instead of an array; this way we avoid some pointer chasing. Even if apparently we "waste" more memory, it's really about the same or less, because the array costs a reference + the array itself (header+n*sizeof(longs)), while in this case we only have 2 longs.

andreaTP: agree, I'll update the PR soon :+1: thanks!

andreaTP: done :+1:

evacchi (approved):

mfirry (approved):

PR #722, opened 2024-12-23
WasmTestsuite from main

### LINKED ISSUE #326: Unpin the testsuite and use `main`
The WebAssembly testsuite, currently, doesn't compile with `wast2json` at version 1.0.34 (still the latest release).
As soon as the next version of `wabt` will be released we should be able to remove those all around:
https://github.com/dylibso/chicory/blob/4c700c4e4c36267d2f18588e977331309841bb94/.github/workflows/ci.yaml#L33

andreaTP:
Finally, this fixes #326 , we are back on track to use the testsuite from `main` with just a sprinkle of validation :slightly_smiling_face:

### DIFF
--- a/aot-tests/pom.xml
+++ b/aot-tests/pom.xml
@@ -139,7 +139,6 @@
             <wast>table_set.wast</wast>
             <wast>table_size.wast</wast>
             <wast>token.wast</wast>
-            <wast>tokens.wast</wast>
             <wast>traps.wast</wast>
             <wast>type.wast</wast>
             <wast>unreachable.wast</wast>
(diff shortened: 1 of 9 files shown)


bhelx (approved):

aot-tests/pom.xml
```diff
@@ -157,6 +156,8 @@
           <excludedUnlinkableWasts/>
           <excludedUninstantiableWasts/>
           <excludedWasts>
+            <!-- obsolete-keywords is only stressing the wat parser -->
```
electrum, line 159: Should we make this part of the Maven plugin?

electrum (approved):
This is great!

aot-tests/pom.xml
andreaTP: It's a single exception at the moment, I'll wait to have more cases to refactor in the Maven Plugin.

PR #732, opened 2025-01-09
More SIMD opcodes

andreaTP:
cc. @mfirry


mfirry:
LGTM! My review can mainly focus on the `SimdInterpreterMachine` class. 
One thought I had: as this file continues to grow (likely reaching 1000+ lines soon), should we consider reorganizing it into smaller modules or splitting the logic across multiple files to improve maintainability?

andreaTP:
> should we consider reorganizing it into smaller modules or splitting the logic across multiple files

In general, I think it's expected that the OpCodes implementation will take some lines of code ...
for example, the current `InterpreterMachine` is > 2000 LOCs but is still manageable and organized well enough.
Splitting in multiple files not always improves readability and maintainability.

That said, @mfirry , if you have a proposal, go ahead and we can evaluate it, alternatively we will face the problem when it hurts.

evacchi (approved):
looks good, as long as the tests are passing 😎

PR #739, opened 2025-01-14
More SIMD opcodes

andreaTP:
Some things might benefit from refactoring later on, or to be better implemented using the Vector API.
For now, this looks as a reasonable step forward.

cc. @mfirry


andreaTP:
Adding the remaining op codes, moving to draft for the moment.

evacchi (approved):

Issue #535, opened 2024-09-20
Check Android support in CI


andreaTP:
It would be great to verify that Chicory works in an Android application.

I just casually found a starting point for implementing this:
- found [here](https://github.com/mockk/mockk/blob/edf9f95ecfae314c22452a1776b074803ae4a42d/.github/workflows/gradle.yml#L87)
- and [this](https://github.com/AdamMc331/AndroidStudyGuide/tree/development) looks to be an easy way to get started

yigit:
Hi,
So i took a look at this (totally random, found this project when i was looking into [mcp.run](https://docs.mcp.run/developing-a-servlet/) for some personal project ideas)

Unfortunately, because you are using maven, there is no easy way to set this up, because there is no official tooling for maven.
You are also using junit5, which is not officially supported for android instrumentation tests (but there is a [3rd party plugin](https://github.com/mannodermaus/android-junit5) ).

Here is what I've done:
* locally compile chicory 
   ` mvn deploy -DaltDeploymentRepository=local-repo::default::file:./local-repo`
* create a new android gradle project next to `chichory` checkout
   ```
    ➜  src ls |grep chicory
    chicory
    chicory-android
* included that repo in my project as a maven repository
   ```
   settings.gradle.kts
   maven {
       url = uri(settingsDir.resolve("../chicory/local-repo"))
   }
   ```
* converted `.class` files into a jar from `<chicory-root>/runtime/target/test-classes` and included in my android project
* run tests 

[chicory-android.zip](https://github.com/user-attachments/files/18354488/chicory-android.zip)

yigit:
Hi,
So i took a look at this (totally random, found this project when i was looking into [mcp.run](https://docs.mcp.run/developing-a-servlet/) for some personal project ideas)

Unfortunately, because you are using maven, there is no easy way to set this up, because there is no official tooling for maven.
You are also using junit5, which is not officially supported for android instrumentation tests (but there is a [3rd party plugin](https://github.com/mannodermaus/android-junit5) ).

Here is what I've done:
* locally compile chicory 
   ` mvn deploy -DaltDeploymentRepository=local-repo::default::file:./local-repo`
* create a new android gradle project next to `chichory` checkout
   ```
    ➜  src ls |grep chicory
    chicory
    chicory-android // this is the android project zipped in attachments
* included that repo in my project as a maven repository
   ```
   settings.gradle.kts
   maven {
       url = uri(settingsDir.resolve("../chicory/local-repo"))
[...]

PR #743, opened 2025-01-18
Run Chicory tests on Android

yigit:
This change adds a new Android project to run Chicory tests on Android devices.

Android doesn't have official maven support, hence this is not created as a maven project. Instead, this Android project has the ability to compile the main project and copy its tests to run the as Android tests.

See the android-tests/README.md for details on how it works.

This first change only adds the runtime tests but similar setup can be created for other modules.

Furhermore, currently only API 33+ is supported due to InputStream.readAllBytes API usages, which is only added on API 33 to Android.

The CI is setup to run tests on 33 and 35. It is kept as a separate workflow file to avoid interference with the main build.

Spotless is setup with the KOTLINLANG default style, as there is no matching AOSP style for ktfmt.

Bug: #535
Test: CI


yigit:
i'm guessing we'll need to iterate more here but just wanted to start the PR to get it rolling.

I think this setup is a feasible compromise between being too magical vs too many copy paste to add more modules.

I've only added runtime tests for now, as other tests will require more work and would better be done as followups.

Finally, I've kept the CI separate so it doesn't slow down pre-submit much and won't fight for resources w/ the main project. (not sure if you are using free or paid CI machines).

yigit:
as for how to read this PR:
* workflows/android.yaml -> CI setup
* android-tests/README.md -> lmk if that is good enough for contributors to understand
* android-tests/build.gradle.kts -> main file that takes care of building main project as a repository
* android-tests/device-tests/build.gradle.kts -> where we setup main project's modules as android tests
* pom.xml -> spotless setup

The rest is mostly general android setup goup. :)

android-tests/README.md
```diff
@@ -0,0 +1,40 @@
...
+The dependencies between this project and the maven project are setup properly
+such that, if the code in the main maven project changes, this Android project
+will recompile the repository and run up-to-date tests.
+
+To avoid re-building the main project (e.g. in CI), you can also pass
+`CHICORY_REPO` environment variable, in which case, this Android project will
+re-use its output instead of recompiling the main project.
+(It won't make any attempt to compile their tests either)
+```
+mvn deploy -DaltDeploymentRepository=local-repo::default::file:./local-repo -DskipTests
```
evacchi, line 26: the top-level project already requires to do a local `mvn install` (because of the setup with codegen plugins), so my suggestion would be to keep this super-simple, and instead just add `mavenLocal()` to

```kotlin
dependencyResolutionManagement {
    ...
    repositories {
        google()
        mavenLocal()
        mavenCentral()
    }
} 
 ```
 
 in settings.gradle.kts and then 

1) setup a Maven submodule that generates a test jar and publishes it to the local maven directory (in fact, `runtime-tests` should contain already pretty much all we need for the interpreter mode tests.)

2) then all this gradle project will need is in to depend on that jar (possibly we might need to instruct the gradle/junit runner to look into that local jar)

NOTE: I am not saying _you_ should do it @yigit, we can assist with that; this might help guidance if you want to look into it, but TBH I don't remember how that works and I'd have to read through it myself :P  https://maven.apache.org/guides/mini/guide-attached-tests.html

then all of this setup should become quite straightforward.

Users will have to either run this gradle subproject manually after running `mvn install -Dsomeprofile`;
we could even add an `invoke` task to the top-level pom.xml so that `gradle` is invoked from the top-level maven project (probably for a specific profile), but that is not strictly required (the install step would be still needed).
 
 WDYT?

Comment (yigit) [reply android-tests/README.md:26]: Sorry, I've never used maven professionally so my solutions tend to use gradle because that is what I know :).

The reason why I setup that task in gradle is to allow iterating on Android and isolate it from the system.

* iteration: With this setup, it is possible to iterate on the android while changing the main project. The build will be consistent, such that, it will recompile the main project if necessary ([this code](https://github.com/dylibso/chicory/pull/743/files#diff-63503b85050c812becfcf55164233e8a8bdf1b8ae8d9f9b114eced46f1f2af76R26)). The same thing for the `jar` task, if the files in the `target/classes` directory changes, it will know to repackage them, otherwise, re-use previous output. I even wanted to set it up such that it only compiles main modules that we depend on (instead of publish all) but left it for later.
* isolation: Maven local tends to be brittle as it is re-used between projects. So putting the repo into a custom build/folder and letting gradle handle the sync ensures the build is not affected by other things running on the computer. I added the env variable to allow skipping another build for CI.

Now, these may not be valuable for the project, so I just wanted to explain the choices there.

I'm curious about the `mavenLocal` requirement in the main project. In Gradle, when you need a project that needs to run on compilation, this is usually done via included builds (the previous more complicated setup I had). Does maven not have something similar?

As for the suggestions, I think the curent setup is better for the reasons above. But it is **your** project and I'm **happy to follow** whichever solution you think is the best.
Specifically, moving the jars to the maven module sounds like a good idea. I couldn't figure out how maven handles invalidation (or rather, it seems to invalidate a lot?). That is one concern, where I wanted it to be possible to re-run the android project without recompiling the maven project, when not needed. So if it is possible to delegate test-jars to the maven project and make sure it is incremental, that would be great (even if it re-runs, as long as it runs fast and produces the exact same jar (e.g. no timestamps etc), that would be good enough for gradle since they do content hashing for invalidation). Invalidation has downstream affects (desugaring etc) so that is what I'm trying to avoid mainly, not necessarily the jar task itself.

So tl;dr; I'm happy to change it the way you want so it fits better to the existing project. I'll give the jar stuff a shot and lmk if you want me to move to mavenLocal as well.

Comment (andreaTP) [reply android-tests/README.md:26]: I don't have a super strong opinion on this point, either solution have slight pros and cons.

File: android-tests/README.md [hunk 2]
```diff
@@ -0,0 +1,40 @@
...
+cd android-tests && CHICORY_REPO=../local-repo ./gradlew device-tests:connectedCheck
```
andreaTP, line 27: I attempted to go through the instructions but I got stuck here:

```
Define a valid SDK location with an ANDROID_HOME
```

I don't have an Android SDK installed globally, I attempted to download a few packages, but it seems not enough, can I ask for a link to a guide on how to have a working environment? Preferably on the command line ...

andreaTP (commented):
Thanks a lot for this @yigit !
It looks like a great start, left a couple of comments but we can merge when you prefer considering it an iteration.

Do we need to have a separate issue for `readAllBytes` or are you going to tackle it in the context of this effort?

.github/workflows/android.yaml
```diff
@@ -0,0 +1,68 @@
+name: Android Device Tests
+
+on:
+  push:
```
evacchi, line 4:
```suggestion
on:
  push:
    branches: [ main ]
  pull_request:
```

let's align to `ci.yaml`

android-tests/README.md
evacchi: TBH I am just nitpicking, here :P we can even start off with your contribution and then iterate :)

android-tests/device-tests/build.gradle.kts
```diff
@@ -0,0 +1,80 @@
...
+
+plugins {
+    alias(libs.plugins.android.library)
+    alias(libs.plugins.kotlin.android)
+    alias(libs.plugins.android.junit5)
+}
+
+android {
+    namespace = "com.dylibso.runtimeTests"
+    compileSdk = 35
```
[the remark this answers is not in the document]
yigit: No, these are not intended to be injected. 
Compile sdk and minSdk are similar to java's taget compatibility and source compatibility options. So these are usually encoded here or in gradle.properties file and they don't change per CI etc. 

For android projects, we don't tend to try to compile with older SDK versions (i actually never seen it). What really matters here is the `min` we support, which will affect the output of desugaring tasks and also limit downstream users to that min API (granted, we don't publish this so there is no downstream users).

android-tests/README.md
yigit: Sorry i missed this. I'll write some instructions for it.
The easiest way is to actually use android studio. We also have command line tools. Unfortunately, it is a [bunch of steps](https://github.com/ReactiveCircus/android-emulator-runner/blob/main/src/sdk-installer.ts#L7) and I don't think we have a clear documentation around it that I can link to.

I'll see how complex that is and worst case we can always have the studio instructions (sorry, Android development and Android studio are very interlocked)

.github/workflows/android.yaml
yigit: sgtm. i'll just leave this comment open until we are happy with the PR so it keeps running on my fork :) and will fix it before merge.

yigit:
> Thanks a lot for this @yigit ! It looks like a great start, left a couple of comments but we can merge when you prefer considering it an iteration.
> 
> Do we need to have a separate issue for `readAllBytes` or are you going to tackle it in the context of this effort?

Since it is not affecting this project, i'll not touch readAllBytes for now. We can do that as we extend to other projects in followups.

yigit:
I've pushed a new version that uses maven to build test jars and also expanded the README to add instructions for setup and also adding new tests.

I think this is a reasonable step that can be merged, the only remaining part is updating the CI to not run on every push (just leaving to the end so i can keep testing the CI setup).

Lmk what you think.

android-tests/README.md
evacchi: I think with homebrew you can install `android-platform-tools`, `android-sdk` and `android-commandlinetools` and get pretty much there. Then you'll have to setup a virtual device (e.g. I am using Medium Phone API 35). Indeed, the easiest way is to setup through Android Studio -- also installing the Android plugin on a plain IntelliJ will get you there.

android-tests/build.gradle.kts
```diff
@@ -0,0 +1,114 @@
...
+            val inputRepo = File(prebuiltRepositoryArg.get())
+            check(inputRepo.exists()) {
+                "Cannot find input repository in ${inputRepo.absolutePath}"
+            }
+            filesystemOps.copy {
+                from(prebuiltRepositoryArg.get())
+                into(repositoryLocation.get().asFile)
+            }
+        } else {
+            execOps.exec {
```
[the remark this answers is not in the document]
yigit: Hmm, not sure if I fully understand. The dependency here is unidirectional where Gradle depends on maven but maven doesn't depend on Gradle.

The optional `prebuiltsRepo` argument is more intended for CI if we built these together, so we can avoid a rebuild (because we are using snapshots, it cannot avoid rebuild on maven on subsequent calls).

Is your suggestion to ask users to always call `mvn` manually if they change the project?

Or let me explain what my work setup was here. I opened chicory in IJ, this project in Studio.
(1) make chicory changes in IJ
(2) run the gradle tests (`./gradlew cC`) (or use studio to run)

If we don't call mvn from gradle automatically, that workflow would be:
(1) make chicory changes in IJ
(2) compile repository  (`mvn deploy ...`)
(3) run the gradle tests (`CHICHORY_REPO=<localtion> ./gradlew cC`) (or use studio to run)

This is not necessarily too bad but the problem is that, if someone forgets step 2, step (3) will still run, but with the old code. And that might be red herring.

Hence, i think it is better for gradle to take care of its dependencies automatically (e.g. there is no explicit maven in the CI workflow).

But as mentioned above, I'm happy to follow whatever you prefer, just wanted to explain why it is setup like this.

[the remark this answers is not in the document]
andreaTP: > The dependency here is unidirectional where Gradle depends on maven but maven doesn't depend on Gradle.

Correct, Gradle doesn't depend on Maven, still, I'd prefer to keep the two build system separated, ppl tend to have interesting setups.

> This is not necessarily too bad but the problem is that, if someone forgets step 2, step (3) will still run, but with the old code. 

I think that this degree of flexibility is fine, the CI will prevent red herring from being merged and we have a streamlined setup.

yigit:
I merged your PR then made a cleanup pass to remove all tasks that we don't need anymore. Updated readme etc accordingly.
Made CHICORY_REPO default to `<checkout-root>/local-repo` with support for override.

.github/workflows/android.yaml
```diff
@@ -62,7 +62,7 @@ jobs:
         force-avd-creation: false
         emulator-options: -no-snapshot-save -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
         working-directory: android-tests
-        script: ls -al ../ && CHICORY_REPO=../local-repo ./gradlew device-tests:connectedCheck
```
andreaTP: thanks @yigit , leftover :sweat:

andreaTP (approved):
Thanks for the further cleanup @yigit , this PR looks great!

LGTM, good to merge :+1:

android-tests/gradle/libs.versions.toml
```diff
@@ -0,0 +1,18 @@
+[versions]
+agp = "8.8.0"
+junitJupiterApi = "5.11.2"
+kotlin = "1.9.24"
```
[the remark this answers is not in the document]
yigit: yea 2.0 is used. It doesn't matter much here since we don't have code (we could even remove it probably, i didn't bother and kotlin-stdlib is already included due to android testing dependencies).

It could become relevant if there was some future demand to create a package like chicory-ktx (extensions to ease usage from kotlin).

yigit:
thanks for merging this. When i have next time, I'll try to get runtime-tests in. It had some more complications around guava usage (guava has a different android artifact). If someone else tries it, they'll need these qualifiers:

```
configurations.configureEach {
    resolutionStrategy.capabilitiesResolution.withCapability("com.google.guava:listenablefuture") {
        select("com.google.guava:guava:0")
    }
}
```

I'm actually not sure if it is the right thing to move to guava-android for android artifacts or do this solution. I think it is only used in tests so it wouldn't matter much as long as it doesn't go into the main artifact (then, it might be worth considering the [guava way](https://github.com/google/guava?tab=readme-ov-file#adding-guava-to-your-build) to prevent downstream users from suffering w/ guava).

android-tests/gradle/libs.versions.toml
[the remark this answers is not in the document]
andreaTP: thanks!

andreaTP:
> thanks for merging this. 

Thank you for doing the amazing job @yigit , this contribution is highly appreciated!

> I'll try to get runtime-tests in

Check what I'm doing in #746 , I decided to skip the tests that are using `Wat2Wasm` as it requires a "more complicated" classpath and those tests are not really stressing Chicory.

The next issue looks to be on `skip-stack-guard-page.wast` that is always triggering this error:
https://github.com/andreaTP/chicory/actions/runs/12913361020/job/36010420899#step:9:1129

The test is stressing this codepath:
https://github.com/dylibso/chicory/blob/9926038965c131e88ff8846f40b9e866df2da0f4/runtime/src/main/java/com/dylibso/chicory/runtime/InterpreterMachine.java#L90-L92

And I'm not sure how to debug/narrow it down ...

PR #744, opened 2025-01-20
Load wasm files from classpath in generated tests

andreaTP:
@yigit follow up from your comment regarding loading files, if I understand the problem correctly, this should generate a test artifact that is self-contained and easier to run using gradle.

Can you confirm that this is going in the right direction?

### DIFF
--- a/aot-tests/src/test/java/com/dylibso/chicory/testing/TestModule.java
+++ b/aot-tests/src/test/java/com/dylibso/chicory/testing/TestModule.java
@@ -8,7 +8,7 @@ import com.dylibso.chicory.wabt.Wat2Wasm;
 import com.dylibso.chicory.wasm.MalformedException;
 import com.dylibso.chicory.wasm.Parser;
 import com.dylibso.chicory.wasm.WasmModule;
-import java.io.File;
+import java.io.IOException;
 
 public class TestModule {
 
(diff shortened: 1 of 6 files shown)


evacchi:
I haven't tested this specifically, but I think the approach should work

yigit:
yes, this direction should work, let me give it a try in a branch to see if this PR works for runtime-tests project.

yigit:
yep, this seems to be working 🥳 
https://github.com/dylibso/chicory/commit/c7e6bfd7be332323985278090495a3680a3d36c3
https://github.com/yigit/chicory/actions/runs/12874204085 (it broke artifact upload, need to check)

I've run that locally, 120 tests, 15 failures. Caused by failures similar to this:

```
Caused by: java.lang.NoSuchMethodError: No static method getLogger(Ljava/lang/String;)Ljava/lang/System$Logger; in class Ljava/lang/System; or its super classes (declaration of 'java.lang.System' appears in /apex/com.android.art/javalib/core-oj.jar)
at com.dylibso.chicory.log.SystemLogger.<clinit>(SystemLogger.java:4)
```

In that change, i'm adding resources in Gradle. I think this is more reason to delegate to maven to generate the test.jar as discussed in #743, so it is can be source of truth for test jars.

andreaTP:
Uhm, `SystemLogger` should not be used unless using wasi, I'd love to be able to reproduce locally to better bisect the issue.

That said, if building a `test.jar` in Maven works, it is definitely a good approach to the problem (to be good scouts we should leave a couple of lines detailing how it works but can be done later too).

We can merge this and, please, let me know how I can assist on the Maven side!

evacchi (approved):

PR #746, opened 2025-01-22
WIP - run 'runtime-tests' on Android

andreaTP:
Opening for visibility @yigit and @evacchi .

I'm not done with this:
- I have had headaches trying to get a correct classpath for `wasi` (`Wat2Wasm`), decided to give up and disabled the tests, @yigit help here will be appreciated :pray: 
- I keep having an OOM that causes a failure on the `skip-stack-guard-page` test (it's checking the call stack exhausted)
- The tests are taking considerable time to run now, I'd like to consider running only a linter on PR and the full tests as a nightly

Thoughts?

### DIFF
--- a/.github/workflows/android.yaml
+++ b/.github/workflows/android.yaml
@@ -2,8 +2,8 @@ name: Android Device Tests
 
 on:
   push:
-    branches: [ main ]
-  pull_request:
+  #   branches: [ main ]
+  # pull_request:
 
 jobs:
   build:
@@ -50,6 +50,8 @@ jobs:
         api-level: ${{ matrix.api-level }}
         target: google_apis
         arch: x86_64
+        heap-size: 512M
+        ram-size: 4096M
         force-avd-creation: false
         emulator-options: -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
         script: echo "Generated AVD snapshot for caching."
--- a/android-tests/device-tests/build.gradle.kts
+++ b/android-tests/device-tests/build.gradle.kts
@@ -17,6 +17,7 @@ android {
     flavorDimensions += chicoryDimension
     productFlavors {
         create("runtime") { dimension = chicoryDimension }
+        create("runtimeTests") { dimension = chicoryDimension }
         // add future modules similar to the runtime configuration above.
     }
 
@@ -31,11 +32,16 @@ dependencies {
     // "androidTestRuntimeImplementation" name here comes from Android's product
     // flavor convention. androidTest<productFlavorName>Implementation
     addLibraryTests(configurationName = "androidTestRuntimeImplementation", libraryPath = "runtime")
+    addLibraryTests(
+        configurationName = "androidTestRuntimeTestsImplementation",
+        libraryPath = "runtime-tests",
+    )
     // common dependencies can be added here
     // if you need to add a dependency on a specific module, you can use
     // "androidTest<productFlavorName>Implementation"(<your dependency>)
     // e.g.
     // "androidTestRuntimeImplementation"(libs.chicory.runtime)
+    androidTestImplementation(libs.chicory.runtimeTests)
     androidTestImplementation(libs.chicory.runtime)
     androidTestImplementation(libs.chicory.wasm)
     androidTestImplementation(libs.chicory.wasmCorpus)
--- a/android-tests/device-tests/src/main/AndroidManifest.xml
+++ b/android-tests/device-tests/src/main/AndroidManifest.xml
@@ -1,2 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
-<manifest xmlns:android="http://schemas.android.com/apk/res/android"></manifest>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+    <application    android:hardwareAccelerated="false"
+                    android:largeHeap="true" >
+    </application>
+</manifest>
--- a/android-tests/gradle/libs.versions.toml
+++ b/android-tests/gradle/libs.versions.toml
@@ -8,7 +8,6 @@ chicory = "999-SNAPSHOT"
 junit-jupiter-api = { module = "org.junit.jupiter:junit-jupiter-api", version.ref = "junitJupiterApi" }
 chicory-runtimeTests = { module = "com.dylibso.chicory:runtime-tests", version.ref = "chicory"}
 chicory-runtime = { module = "com.dylibso.chicory:runtime", version.ref = "chicory"}
-chicory-wabt = { module = "com.dylibso.chicory:wabt", version.ref = "chicory"}
 chicory-wasm = { module = "com.dylibso.chicory:wasm", version.ref = "chicory"}
 chicory-wasmCorpus = { module = "com.dylibso.chicory:wasm-corpus", version.ref = "chicory"}
 android-gradle-plugin-api = { module = "com.android.tools.build:gradle-api", version.ref = "agp"}
--- a/android-tests/gradle/libs.versions.toml
+++ b/android-tests/gradle/libs.versions.toml
@@ -6,7 +6,9 @@ chicory = "999-SNAPSHOT"
 
 [libraries]
 junit-jupiter-api = { module = "org.junit.jupiter:junit-jupiter-api", version.ref = "junitJupiterApi" }
+chicory-runtimeTests = { module = "com.dylibso.chicory:runtime-tests", version.ref = "chicory"}
 chicory-runtime = { module = "com.dylibso.chicory:runtime", version.ref = "chicory"}
+chicory-wabt = { module = "com.dylibso.chicory:wabt", version.ref = "chicory"}
 chicory-wasm = { module = "com.dylibso.chicory:wasm", version.ref = "chicory"}
 chicory-wasmCorpus = { module = "com.dylibso.chicory:wasm-corpus", version.ref = "chicory"}
 android-gradle-plugin-api = { module = "com.android.tools.build:gradle-api", version.ref = "agp"}
--- a/pom.xml
+++ b/pom.xml
@@ -919,6 +919,7 @@
       </activation>
       <properties>
         <skipTests>true</skipTests>
+        <wasm-test-gen.disable-wat>true</wasm-test-gen.disable-wat>
         <altDeploymentRepository>local-repo::default::file:${session.executionRootDirectory}/local-repo</altDeploymentRepository>
       </properties>
       <build>
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/InterpreterMachine.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/InterpreterMachine.java
@@ -88,7 +88,9 @@ public class InterpreterMachine implements Machine {
             try {
                 eval(stack, instance, callStack);
             } catch (StackOverflowError e) {
-                throw new ChicoryException("call stack exhausted", e);
+                // We avoid including the StackTrace as it's usually not useful and consumes a lot
+                // of resources
+                throw new ChicoryException("call stack exhausted: " + e.getMessage());
             }
         } else {
             var stackFrame = new StackFrame(instance, funcId, args);
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/StackFrame.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/StackFrame.java
@@ -27,7 +27,7 @@ public class StackFrame {
     private final int funcId;
     private int pc;
     private final long[] locals;
-    private final List<ValueType> localTypes;
+    private final ValueType[] localTypes;
     private final int[] localIdx;
     private final Instance instance;
 
@@ -54,10 +54,15 @@ public class StackFrame {
         this.instance = instance;
         this.funcId = funcId;
         this.locals = Arrays.copyOf(args, sizeOf(argsTypes) + sizeOf(localTypes));
-        this.localTypes = new ArrayList<>();
-        this.localTypes.addAll(argsTypes);
-        this.localTypes.addAll(localTypes);
-        this.localIdx = new int[this.localTypes.size()];
+        int localsSize = argsTypes.size() + localTypes.size();
+        this.localTypes = new ValueType[localsSize];
+        for (int i = 0; i < argsTypes.size(); i++) {
+            this.localTypes[i] = argsTypes.get(i);
+        }
+        for (int i = 0; i < localTypes.size(); i++) {
+            this.localTypes[argsTypes.size() + i] = localTypes.get(i);
+        }
+        this.localIdx = new int[localsSize];
 
         // initialize codesegment locals.
         int j = 0;
@@ -76,7 +81,7 @@ public class StackFrame {
 
         // initialize local indexes
         j = 0;
-        for (int i = 0; i < this.localTypes.size(); i++) {
+        for (int i = 0; i < this.localTypes.length; i++) {
             this.localIdx[i] = j;
             if (localType(i) != ValueType.V128) {
                 j += 1;
@@ -98,7 +103,7 @@ public class StackFrame {
     }
 
     ValueType localType(int i) {
-        return this.localTypes.get(i);
+        return this.localTypes[i];
     }
 
     public int localIndexOf(int idx) {
--- a/test-gen-plugin/src/main/java/com/dylibso/chicory/maven/JavaTestGen.java
+++ b/test-gen-plugin/src/main/java/com/dylibso/chicory/maven/JavaTestGen.java
@@ -46,17 +46,21 @@ public class JavaTestGen {
 
     private final List<String> excludedUnlinkableWasts;
 
+    private final boolean disableWat;
+
     public JavaTestGen(
             List<String> excludedTests,
             List<String> excludedMalformedWasts,
             List<String> excludedInvalidWasts,
             List<String> excludedUninstantiableWasts,
-            List<String> excludedUnlinkableWasts) {
+            List<String> excludedUnlinkableWasts,
+            boolean disableWat) {
         this.excludedTests = excludedTests;
         this.excludedMalformedWasts = excludedMalformedWasts;
         this.excludedInvalidWasts = excludedInvalidWasts;
         this.excludedUninstantiableWasts = excludedUninstantiableWasts;
         this.excludedUnlinkableWasts = excludedUnlinkableWasts;
+        this.disableWat = disableWat;
     }
 
     public CompilationUnit generate(String name, Wast wast, String wasmClasspath) {
@@ -324,7 +328,8 @@ public class JavaTestGen {
             List<String> excludedTests) {
         var methodName = "test" + testNumber;
         var method = testClass.addMethod(methodName, Modifier.Keyword.PUBLIC);
-        if (excludedTests.contains(methodName)) {
+        if (excludedTests.contains(methodName)
+                || (disableWat && cmd.filename() != null && cmd.filename().endsWith(".wat"))) {
             method.addAnnotation(
                     new SingleMemberAnnotationExpr(
                             new Name("Disabled"), new StringLiteralExpr("Test excluded")));
--- a/test-gen-plugin/src/main/java/com/dylibso/chicory/maven/JavaTestGen.java
+++ b/test-gen-plugin/src/main/java/com/dylibso/chicory/maven/JavaTestGen.java
@@ -196,6 +196,12 @@ public class JavaTestGen {
                 case ASSERT_TRAP:
                 case ASSERT_EXHAUSTION:
                     {
+                        if (disableWat
+                                && cmd.filename() != null
+                                && cmd.filename().endsWith(".wat")) {
+                            continue;
+                        }
+
                         method =
                                 createTestMethod(
                                         wast.sourceFilename().getName(),
@@ -328,8 +334,7 @@ public class JavaTestGen {
             List<String> excludedTests) {
         var methodName = "test" + testNumber;
         var method = testClass.addMethod(methodName, Modifier.Keyword.PUBLIC);
-        if (excludedTests.contains(methodName)
-                || (disableWat && cmd.filename() != null && cmd.filename().endsWith(".wat"))) {
+        if (excludedTests.contains(methodName)) {
             method.addAnnotation(
                     new SingleMemberAnnotationExpr(
                             new Name("Disabled"), new StringLiteralExpr("Test excluded")));
--- a/test-gen-plugin/src/main/java/com/dylibso/chicory/maven/JavaTestGen.java
+++ b/test-gen-plugin/src/main/java/com/dylibso/chicory/maven/JavaTestGen.java
@@ -196,12 +196,6 @@ public class JavaTestGen {
                 case ASSERT_TRAP:
                 case ASSERT_EXHAUSTION:
                     {
-                        if (disableWat
-                                && cmd.filename() != null
-                                && cmd.filename().endsWith(".wat")) {
-                            continue;
-                        }
-
                         method =
                                 createTestMethod(
                                         wast.sourceFilename().getName(),
@@ -258,6 +252,12 @@ public class JavaTestGen {
                 case ASSERT_UNINSTANTIABLE:
                 case ASSERT_UNLINKABLE:
                     {
+                        if (disableWat
+                                && cmd.filename() != null
+                                && cmd.filename().endsWith(".wat")) {
+                            break;
+                        }
+
                         method =
                                 createTestMethod(
                                         wast.sourceFilename().getName(),
--- a/test-gen-plugin/src/main/java/com/dylibso/chicory/maven/TestGenMojo.java
+++ b/test-gen-plugin/src/main/java/com/dylibso/chicory/maven/TestGenMojo.java
@@ -102,6 +102,12 @@ public class TestGenMojo extends AbstractMojo {
     private List<String> excludedWasts;
 
     /**
+     * Disable the execution of tests on WAT files.
+     */
+    @Parameter(property = "wasm-test-gen.disable-wat", defaultValue = "false")
+    private boolean disableWat;
+
+    /**
      * Skip execution of this Mojo.
      */
     @Parameter(property = "wasm-test-gen.skip", defaultValue = "false")
@@ -136,7 +142,8 @@ public class TestGenMojo extends AbstractMojo {
                         excludedMalformedWasts,
                         excludedInvalidWasts,
                         excludedUninstantiableWasts,
-                        excludedUnlinkableWasts);
+                        excludedUnlinkableWasts,
+                        disableWat);
 
         JavaParserMavenUtils.makeJavaParserLogToMavenOutput(getLog());
 


yigit:
Have you tried running this locally?

```
./gradlew device-tests:connectedRuntimeTestsDebugAndroidTest -Pandroid.testInstrumentationRunnerArguments.class=com.dylibso.chicory.test.gen.SpecV1SkipStackGuardPageTest#skip-stack-guard-page.wast
```

I'm wondering if we are leaking some memory during tests or it is just this test itself.

evacchi:
just curious, I was expecting the time to increase a lot, how long is it taking now? Anyway, +1 on running nightly and/or on-demand.

andreaTP:
> Have you tried running this locally?

Yes, same result, but thanks for the command line, it's helpful to continue debugging.

> I was expecting the time to increase a lot

It takes approx 10 minutes to get to the end of the tests.

evacchi:
Not too bad after all :) but it still probably make sense to make it optional/nightly for now

yigit:
I can take a look at the test time eventually.
If we can make the maven task produce reproducible outputs (e.g. not snapshots and no timestamps in jars), we should be able to setup gradle cache to avoid unnecessary runs. 

We can also shard and/or run tests in parallel. 
But until then, nightly seems fine. 

There is always the option to use firebase test lab to further speed them up. It will not cost much, but that would be a consideration for later. Just mentioning it for reference that we have options if we need these tests to run fast.

evacchi:
> ```
> ./gradlew device-tests:connectedRuntimeTestsDebugAndroidTest -Pandroid.testInstrumentationRunnerArguments.class=com.dylibso.chicory.test.gen.SpecV1SkipStackGuardPageTest#skip-stack-guard-page.wast
> ```

yes! the test actually passes with:

```
./gradlew device-tests:connectedRuntimeTestsDebugAndroidTest -Pandroid.testInstrumentationRunnerArguments.class=com.dylibso.chicory.test.gen.SpecV1SkipStackGuardPageTest
```

so we pair-debugged this a little and it looks like the main issue is memory pressure because a lot of tests are run in a small amount of time. We tried adding a `Runtime.getRuntime().gc()` in a strategic place (`TestModule`) but probably the issue is scoped to that particular test (so far) because it intentionally grows the virtual stack of the interpreter, causing a lot of garbage to be retained. 

so, either we find a way to setup the VM so that is not an issue (we did try to change the `vm.heap` flag on the VM but that does not seem to really affect the outcome), or we find a way to isolate the units (maybe each in its own VM / or a group of them) so that we don't incur in the issue

finally, one last resort would be to add a stack count in the interpreter, and flag that specific test so that the we put a limit on it.

evacchi:
FYI on my machine (M3 Pro) it takes 6m42s to run the full suite excluding `SpecV1SkipStackGuardPageTest`:

```
        testInstrumentationRunnerArguments["notClass"] = "com.dylibso.chicory.test.gen.SpecV1SkipStackGuardPageTest"
```

for some reason the task still fails but the suite ran successfully:

```
> Task :device-tests:connectedRuntimeTestsDebugAndroidTest FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':device-tests:connectedRuntimeTestsDebugAndroidTest'.
> There were failing tests. See the report at: file:///Users/evacchi/Devel/java/chicory/android-tests/device-tests/build/reports/androidTests/connected/debug/flavors/runtimeTests/index.html

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --debug option to get more log output.
> Run with --scan to get full insights.
> Get more help at https://help.gradle.org.

BUILD FAILED in 6m 42s
45 actionable tasks: 2 executed, 43 up-to-date
Watched directory hierarchies: [/Users/evacchi/Devel/java/chicory/android-tests]
```

Screenshot 2025-01-23 at 14 02 35

andreaTP:
With the updated commits, especially after `android:largeHeap=true`, I have the full `runtime-tests` running to completion successfully.

Now the next issue is that, as @evacchi noticed too, the Android SDK disconnects/crash when trying to send back the tests results ( [its reproducible in CI](https://github.com/andreaTP/chicory/actions/runs/12933691842/job/36072906101#step:9:366) ).

@yigit any clue? Can you try to take a look?

yigit:
First time I've seen that issue but looks like it was already reported:

https://issuetracker.google.com/issues/330756856

For short term, i think sharding is the easiest solution we have (assuming the junit5 runner doesn't have an issue with it).

Hopefully this weekend I'll have some time to play with this.
Given that this test runs by itself but not with others, I'm curious if there might be a leak somewhere. 

> but probably the issue is scoped to that particular test (so far) because it intentionally grows the virtual stack of the interpreter, causing a lot of garbage to be retained.

Is there a limit on the virtual stack or is it basically the limit of the host machine. If it is the limit of the host, yes we'll likely hit that no matter what. If it is bounded by the host process, it might make sense to limit it to (max memory - some buffer like a mb). But I'm guessing the same thing would be a problem on JVM. Maybe the overhead of the test runner is bigger as it has to keep everything in memory until reporting (about which i'm surprised but also not that surprised :) )

yigit:
I've created 2 variants with sharding.
* [One of them](https://github.com/yigit/chicory/actions/runs/12944848802/workflow) just shards,
* The [other one](https://github.com/yigit/chicory/actions/runs/12944814346/workflow) moves the maven build to a separate step so others can re-use its output. This would reduce github resource usage but would likely increase CI times.

We also don't have to use matrix build and can simply invoke gradle multiple times in the same build to run shard by shard. We might need [some massaging](https://medium.com/mesmerhq/shard-your-android-espresso-tests-for-faster-execution-in-parallel-e66f1b5061ae) to prevent results from overriding each-other. We may also not want to shard small flavors (e.g. run `connectedRuntimeDebugAndroidTest` and then `connectedRuntimeTestsDebugAndroidTest` with shards)

The other 2 things in those branches are:
* zip the results: GH-upload does not support `:` in file names, so we need to zip them before calling github upload action.
* gradle scans: I've enabled gradle scans. Gradle doesn't really understand android tests but it will give some overall visibility to what is happening in the gradle build. Is just good to have for reference and diagnosis in the future.

It is failing with 
```
java.lang.NullPointerException: Attempt to invoke virtual method 'com.dylibso.chicory.runtime.Instance$Exports com.dylibso.chicory.runtime.Instance.exports()' on a null object reference
at com.dylibso.chicory.test.gen.SpecV1AddressTest.test6(SpecV1AddressTest.java:105)
at java.lang.reflect.Method.invoke(Native Method)
at java.util.ArrayList.forEach(ArrayList.java:1262)
at java.util.ArrayList.forEach(ArrayList.java:1262)
```
like errors but I didn't get a chance to diagnose.

andreaTP:
@yigit thanks a lot for getting back and all the information!

> Is there a limit on the virtual stack or is it basically the limit of the host machine.

I did the research(specs/proposals etc.) and haven't found any good reference, looks like everyone is relying on the underlying runtime limit.
The interesting story is that the emulator was throwing this error:

```
java.lang.OutOfMemoryError: OutOfMemoryError thrown while trying to throw an exception; no stack trace available
```
already [reported here](https://github.com/android/android-test/issues/285).
Basically everything is working as expected, but, given the `StackOverflowError` contains a huge StackTrace, the emulator was crashing with OOM trying to fill it in :sweat_smile: 


> i think sharding is the easiest solution

I'm fine with it if it's reproducible locally (and we can bundle the command to execute everything locally in a bash script).

Trying out the commands in your branch, sharding is trying to execute single tests methods on different shards and this is not going to work (that's the root cause of the NPE you noticed).
We discovered early that WebAssembly `wast` tests([for example](https://github.com/WebAssembly/testsuite/blob/main/linking.wast)) are not self-contained, they are dependent on the ordered executions of all the methods in each class (that's why we heavily depend on the `Ordered` annotation).
Basically, we should ensure that sharding is keeping classes and not methods as the minimal split unit.

Regarding "zipping the results" and "Gradle scans," I truly appreciate your enthusiasm and your drive to make things better—thank you!

That said, it’s important to keep the scope in mind. The Gradle build we use for running tests on Android serves as a basic sanity check. The simpler it is, the better.

More specifically:
- Test results: These tests are focused solely on ensuring compatibility with Android. I don’t anticipate gaining deep insights from the results. In the case of a failure, the primary need is simply the ability to reproduce it locally.
- Gradle scans: My goal is for us to have a setup that is stable and rarely needs updating or fixing. Once we achieve that, debugging the setup should ideally not be necessary.

Thank you again for your efforts—I really value your commitment and thoughtfulness!

andreaTP:
Forgot to follow up regarding the build matrix:
I think that we can keep a single execution testing only the stricter requirement, @yigit do you see any issue with this approach?

yigit:
I'm afraid android test runner does not support sharding by class 😞 .
We could consider doing it in Gradle. Since we already have a jar, it wouldn't be hard to partition that by class into multiple jars (and multiple flavors). Obviously, that would mean more complex build logic in gradle but I'm not sure how easy that is in maven. Maybe you can create a maven task to do that ?

> I think that we can keep a single execution testing only the stricter requirement, @yigit do you see any issue with this approach?

Not sure if I understand. You mean invoking Gradle multiple times per shard? We can definitely extract this to a shell script or a single Gradle task (might be nicer for local analysis as Studio would discover that as well, so one can debug a specific failure). But I'm guessing shell script will be easier for contributors.

Also, about the test results/scans:
I think tests results are important as it immediately tells what failed in a nice html page. I can remove the xml reports though (i think those are the ones creating the problem).

Scans are good if we want to ever optimize this etc but doesn't really matter (it also doesn't cost to publish them for open source projects). 

But we can remove either of these (though I would recommend keeping test results).

yigit:
I played a bit more with this, sharding by class etc.
But just the `SpecV1MemoryCopyTest` is big enough that it hits the grpc limit by itself. The limit might be per class, i've not checked what is really going on there.

Maybe that can be avoided by changing codegen to divide tests but based on the order problem you've mentioned, I'm not even sure if that is feasible.

I think the only viable option might be to generate fat tests for android. e.g. JavaTestGen could generate tests that look like this, when invoked for android.

```
class GeneratedTest {
    // @Test() // remote test annoatations from test methods
    // @Order(1)
     // @DisplayName("memory_copy.wast:15")
    public void test1() {
        var varTest = testModule0Instance.exports().function("test");
        var exception = assertDoesNotThrow(() -> varTest.apply());
    }
    ...
    @Test // have 1 test to call all others in the right order.
    @DisplayName("run all tests in wasmFileName")
    public void runAllTests() {
        // maybe call instantiates here or keep relying on order for them
        test1();
        test2();
        test3();
    }
}
```

Not sure if that is something you would be interested in doing but I'm a bit out of options here :'(. Sorry, i didn't expect android support to get so complicated :/.

evacchi:
no worries, all your suggestions are very welcome :D 

> keep a single execution testing

I think what @andreaTP meant is "a single command to run the entire suite", but let's hear from him

> I think the only viable option might be to generate fat tests for android. e.g. JavaTestGen could generate tests that look like this, when invoked for android.

I don't know what you guys think of this approach, but if we know what tests cause the problem (and I believe they will be a handful) maybe we could just exclude them from the main run and then create a task that run those separately?

e.g. 

```
testAll() {
   runFirstBatch()
   runSecondBatch()
   // if necessary, more batches
}
```

we could also "tag" the tests with an annotation (e.g. [`@LargeTest`](https://developer.android.com/training/testing/instrumented-tests/androidx-test-libraries/runner#filter-tests)), if the number of methods is above some threshold

andreaTP:
Quickly answering from the phone.

> You mean invoking Gradle multiple times per shard?

No, I mean removing the matrix for Android SDK and test only on 33.

Re: test results, I'm happy to keep whatever comes for free, e.g. a step in CI is fine.
If it requires work or a complicated setup we should try to avoid it.

Re: scans, no strong opinions, the Gradle build is a side build for this project and I'm not sure how I would action a feedback from the scans.

>  is big enough that it hits the grpc limit by itself.

This is really worrying!
Some Simd tests are generating more than 3 Mb of source code... we are going to hit the limit anyhow.
Do you know where to open a relevant issue about this bug?
The grpc limit should be tweakable with properties somewhere...

> Maybe you can create a maven task to do that ?

The current test gen is already able to split the generation based on configuration, but I don't think we should invest in a complex setup there.

All in all, I think that the priority here is to open a bug report and try to overcome the grpc limit/client disconnection (this is the biggest blocker).

Second we should start considering running a few smoke tests instead of the full TCK.
I'm thinking about something like `machine-test` running a real world wasm workload, using the interpreter and the AOT at compile time, including Wasi etc.

andreaTP:
> Sorry, i didn't expect android support to get so complicated :/.

@yigit I was honestly expecting some hiccups :-) and I'm grateful to have someone knowledgeable like you assisting in the process. Thanks a ton and no worries!

yigit:
I noticed that bug somehow didn't get triaged and was reported in the wrong component. I moved it to the Android Gradle Plugin and will ping someone during the week.

I'm afraid [this is the code](https://cs.android.com/android-studio/platform/tools/base/+/mirror-goog-studio-main:transport/test-framework/src/main/java/com/android/tools/transport/grpc/Grpc.kt;l=22?q=grpc&ss=android-studio%2Fplatform%2Ftools%2Fbase) that creates that grpc client:



And it doesn't receive any configuration parameters. 
I'll check w/ the gradle team later in the week, meanwhile, I like the annotation solution from @evacchi . FYI we don't need to use android annotations (e.g. LargeTest), we can make up an annotation and filter them out while running tests (or "shard" runs using annotations).

Are the `machine-tests` more end to end tests? If so, definitely worth prioritizing them.

andreaTP:
Separating the tests in groups with annotations is definitely possible and easy to implement :+1: as soon as we verify that it can work I volunteer to make the changes (e.g. `@Group1` `@Group2` etc.).

`machine-tests` are running QuickJs(compiling and running a simple Js example) with Wasi and mixing "Machines".
For Android we should be able to check compile time AOT and the interpreter.
The idea is: if a JS compiler and runtime works we have some confidence that things will work on Android, and we can treat the rest as bugs increasing coverage when needed.

andreaTP:
> And it doesn't receive any configuration parameters.

Usually it is possible to use env vars and global properties anyhow, I'll check on Mon.

andreaTP:
@yigit no pressure and no timelines, just curious if you managed to keep the ball rolling on this subject and you have any kind of news :slightly_smiling_face: .

I'd be happy to help, but this is all very new for me, and I'm unsure where to start, for example, to try to tweak the grpc limit. Any link/information will be much appreciated!

yigit:
Hey sorry i forgot to update here but i did re-route it internally. The team is looking into it but it being fixed then shipping in a version that we can use is going to be a long time :'(

And there doesn't seem to be a workaround.

I think our 2 options here are:
* don't run big tests
* create fat tests that run all off these 1 by 1 without them being individual tests.

If (2) is not too complex of a code-generator change, I think that is the better option so we get coverage without waiting for another AGP release. I can give it a try if you are open to that solution.

andreaTP:
Thanks a lot for getting back @yigit !

> I can give it a try if you are open to that solution.

Given the past conversation where you mentioned that even some single test classes are exceeding the limits I'm not sure how it will work.
That said, if you have something in mind, please go ahead with a poc and we iterate on it, I have no specific concerns with the approach.

In my mind, after #748 , you should be able to avoid generating the test jars, and directly invoke the codegen from Gradle to materialize test classes on the fly.
In this way you have more granular control over what will execute.
Again, this is merely "one way" of doing things, and I'm happy to support any viable alternative.

If everything fails for running the testsuite we can always fallback to smoke testing, in this case the next step would be to include the manually written wasi tests as they are running some popular languages "hello world" samples.

yigit:
> Given the past conversation where you mentioned that even some single test classes are exceeding the limits I'm not sure how it will work. That said, if you have something in mind, please go ahead with a poc and we iterate on it, I have no specific concerns with the approach.
Oh i think that was a mis-understanding. Just running 1 test where each test case is a separate `@Test`  doesn't work (hits the limit). If we merge them into 1 test, it should work (i assume, haven't tried). 

> 
> In my mind, after #748 , you should be able to avoid generating the test jars, and directly invoke the codegen from Gradle to materialize test classes on the fly. In this way you have more granular control over what will execute. Again, this is merely "one way" of doing things, and I'm happy to support any viable alternative.
I'll take a look. So this is basically import that test-gen library from gradle (like a regular dependency for the build) and then invoke it directly with proper source sets?

andreaTP:
> If we merge them into 1 test, it should work (i assume, haven't tried).

This should be doable but a bit of work ... I'd try to avoid it if possible.

Re Gradle, I was thinking about doing something like this:

https://github.com/fabric8io/kubernetes-client/issues/4829#issue-1569942558

yigit:
So i prototyped this a bit but I'm not sure if this is what you want, so I wanted to double check.

We can definitely create a Gradle task to execute TestGen and make it a generated source task for gradle.
https://github.com/dylibso/chicory/compare/main...yigit:chicory:throwaway/generate-tests-via-gradle?expand=1

But that means we have significant configuration in gradle duplicated from maven (included excluded wast files etc).

And I know your preference is to keep more stuff in mvn than gradle so I wanted to double check if this is the direction you want to take. Note that the prototype is not complete but it should give an idea on what it would look like.

We can also move such configuration outside the build.gradle.kts into some properties file as well but it will still be duplicate of the pom.xml files.

andreaTP:
Thanks for taking a look @yigit , a question to help me understand: does this already work in CI?

Moving the configuration to a properties file seems feasible, currently we have 3 different configurations:

- aot
- runtime for Java < 21
- runtime with Simd

But I'd like to keep the current flexibility of inclusion/exclusion as it helps a lot during development (e.g. enables incremental iterations for big features/proposals).

For Android I think it's fine to start with an hardcoded small subset and add things incrementally as we see need. WDYT?

A side note that in the Gradle build we can just assume the `testsuite` folder is already there in the expected place.

andreaTP:
Bonus: one thing that I'd value is IDE integration when things go wrong.
E.g. I would try to break something and check that the dev flow is not too painful to get it fixed.

yigit:
> Thanks for taking a look @yigit , a question to help me understand: does this already work in CI?

No. Or rather, I've not tied all of it since I thought you wouldn't want to have so much Gradle configuration. Should be close.  I can carry it to finish since we resolved that part.

> But I'd like to keep the current flexibility of inclusion/exclusion as it helps a lot during development (e.g. enables incremental iterations for big features/proposals).
For Android I think it's fine to start with an hardcoded small subset and add things incrementally as we see need. WDYT?

sgtm

> A side note that in the Gradle build we can just assume the testsuite folder is already there in the expected place.

Yea, i had to make some changes in the test generator (downloader) because it was assuming running on the project root. You still probably want that change because it doesn't use the `testSuiteFolder` parameter when downloading. 
Can you help me update `android-prepare` to do that, without building tests?
Btw, if I were to set this up properly in gradle, it would be:
one DownloadTestSuite task that would only take care of downloading
and multiple generate test tasks, that would use the output of the download task (per flavor, assuming these tests are different).
Given that we already require running mvn first, it would be OK to move DownloadTestSuite to maven.

>Bonus: one thing that I'd value is IDE integration when things go wrong.
E.g. I would try to break something and check that the dev flow is not too painful to get it fixed.

This part is not great unfortunately. e.g. when i was trying to fix the test downloader, every time I change its code, I have to re-run `mvn -Dandroid-prepare` then re-run gradle. There is probably more specific tasks that I could've run in maven but I have no idea about maven :D. But it would be great if we can remove things like spotless from `android-prepare` and make it only publish artifacts + download the wasm test suite.

yigit:
oh actually we are missing an important point. 
If we were to generate tests from gradle, that means we are no longer getting the hand written test sources  either :/
So that will be another problem.

e.g. it fails with:

```
chicory/android-tests/device-tests/build/generated/java/generateTestSuite/com/dylibso/chicory/test/gen/SpecV1AddressTest.java:16: error: package com.dylibso.chicory.testing does not exist
import com.dylibso.chicory.testing.TestModule;
```
We can definitely figure out a way to include them, or include the jar and filter out generated test classes. 
But i don't think it helps at this point, vs depending on the generated test jar. If all we want is to exclude certain tests, we can annotate test classes and pass an exclusion annotation while running these tests on Android.

yigit:
here is a quick implementation of fat test:

https://github.com/andreaTP/chicory/compare/add-runtime-tests-android...yigit:chicory:fat-tests-2?expand=1

This can definitely be done better but just to test it out, I added a post processing method to remove all annotations and instead call each test 1 by 1. and it did pass on CI:

https://github.com/yigit/chicory/actions/runs/13095218702

I think it would make sense to group them module to at least get some isolation.

andreaTP:
@yigit thanks for all your effort! It is highly appreciated!

I'll take a closer look tomorrow, but I'm mostly aligned with your takes (e.g. if you wanna separate the testsuite downloader just go for it :+1: ).

Want to double check if we explored using Chicory from sources, in this way after one cycle of mvn + Gradle build you should be able to directly make changes and the IDE will pick them up on the fly.
There will be cases (e.g. during this dev phase) where things won't work, that's inevitable, still, sounds like a possible improvement, wdyt?

yigit:
The sources is tricky. We would need to maintain 2 builds (maven & gradle) and need to synchronize build changes.
And that introduces some risks. e.g. if you change a dependency that would break Android, but forget to change the Gradle file, that could give false positives.

It is definitely possible but I'm not sure if that is a good direction wrt maintenance.

Another option is to also consider moving to Gradle completely. Idk if you would be interested in that unless your team is already familiar with Gradle. There is also the risk of finding the same plugins in Gradle but I would assume by now they all probably have Gradle support or replacements. Given that Android is not a big focus for you, it seems excessive to move because of it, unless you were already thinking about moving to Gradle.

If you are open to having 2 builds, for synchronizing dependency versions, probably the best option is to write a script that will convert your dependency declarations (in the main pom.xml) and generate the libs.versions.toml file (the equivalent in Gradle).  And have a CI step to validate that it is updated. 

Making Gradle load maven projects is an option as well but you've seen that variant before, requires a fairly complex Gradle setup.

So tl;dr; I think the best return for the effort is the fat test direction. It keeps most build logic in maven, allows us to validate Android. Albeit, local iteration experience is not great. But I don't know how to get to there without making Gradle understand the whole project  😞

andreaTP:
Sorry for the delay @yigit , I'm reviewing your exploration and your comments.

> We would need to maintain 2 builds

We don't want to duplicate effort here, is already challenging enough, for example, when it comes to Java versions handling etc.

> Another option is to also consider moving to Gradle completely

This has been proposed in the past a few times, I'm not opposed to it in any way, I have some Gradle experience and no specific issues with it.
That said, I have been the one person taking care of the build so far, it's a fact that I'm more proficient with Maven than Gradle.
If anyone want to step up as a maintainer and take care of the project build in the longer term, I'm happy to give responsibilities (and time-consuming tasks :sweat_smile: ) away, as long as I'm the one (mostly) in charge of it, Maven is going to stay.

> If you are open to having 2 builds, for synchronizing dependency versions

Here I'm confused, using `999-SNAPSHOT` as a stable identifier in the Gradle build the transitive dependencies should be resolved automatically, or am I missing something?

> fat test direction

I took the time to evaluate more in detail this option, I understand the reasons for your proposed draft implementation, and thanks a lot for spending the time to put it together.
Although I'm not 100% comfortable with it, as it makes the Test extremely involved to use for debugging.
I attempted to implement a more "idiomatic" encoding, splatting everything in the method body ( in [this diff](https://github.com/dylibso/chicory/compare/main...andreaTP:chicory:fatTest) ), and I'm not happy with the result, furthermore not all `.wast` files can be used as we hit the `code too large` (restrictions over the size of single method) exception by `javac`.


All of that said, I'm really happy to have the `runtime/WasmModuleTest` already running, but, given all the complications we are finding, I propose to pause the effort on having the full WebAssembly testsuite running on Android.
The next steps I see are:
- get `wasi-tests/WasiPreview1Test` to run
- enable running the [AOT Maven plugin](https://chicory.dev/docs/experimental/aot#pre-compiled-aot) from a Gradle build
- run only the interpreter + pre-compiled AOT combination of `machine-tests/MachinesTest` on Android as a "smoke E2E test"
What do you think about this proposal @yigit ?

I'm sorry that you invested so much time with not big progress @yigit , but this exploration has been highly instructive and pave the way for trying again in the future, maybe when the issues in the tooling have been solved :sweat_smile:

yigit:
re: Gradle. Yea, totally make sense. I firmly believe the best build tool for the job is the tool you know the best.

> If you are open to having 2 builds, for synchronizing dependency versions
>> Here I'm confused, using 999-SNAPSHOT as a stable identifier in the Gradle build the transitive dependencies should be resolved automatically, or am I missing something?

This was the case where I thought we would compile tests from sources. Then we need to duplicate your test dependencies. But maybe it is not a big of a deal because if you add some dependency that would break android testing and we would miss it because we didn't bump the android tests, i don't think that is actually a real problem (It doesn't mean the library is broken and the course of action would be to use the old version while testing anyways).
tl;dr; as long as we use the main build for main artifacts, this should be fine.

> I attempted to implement a more "idiomatic" encoding, splatting everything in the method body ( in [this diff](https://github.com/dylibso/chicory/compare/main...andreaTP:chicory:fatTest) ), and I'm not happy with the result, furthermore not all .wast files can be used as we hit the code too large (restrictions over the size of single method) exception by javac.

Yea, this is a problem we hit before with Data Binding or Room :(. One thing though, since your tests are dependent on the order, how does local testing work? You already cannot run a single test while iterating, can you?
What I was thinking as next iteration was to generate:

```
@Test
fun testModule() { // the equivelant of initiate module
    // each step in this module
    testStep1()
    testStep2()
}

@Test
fun testModule2() { // the equivelant of initiate next module
    // each step in this module
    testStep3()
    testStep4()
}
```

With the "removed @Test annotations" from testXXX methods, this should work fine. And wrt local iteration, since your devs cannot already run a single test, this would be the best they get anyways, for maven too?
Now, this will not print the best errors but the assertion generation could be improved to have the line numbers so the failures might be more clear.

> I'm sorry that you invested so much time with not big progress @yigit , but this exploration has been highly instructive and pave the way for trying again in the future, maybe when the issues in the tooling have been solved 😅

Please do not worry about that. My full time job is doing open source for the last 11 years so I'm very much used to being on your side of the aisle. I'm doing this because it is interesting. Of course I would be happy if things merge but I won't get offended or discouraged when things don't align.

> The next steps I see are:

I have to check these in a bit more detail to fully understand but whatever you think is best. I might just also update my CL to have that per-module variant but you don't need to merge (just to close that loop). Then next time I have time, I can try to look into those steps (or the "compile tests sources in gradle" case.

andreaTP:
> how does local testing work? You already cannot run a single test while iterating, can you?

you are right this doesn't work out of the box, a common trick that can be applied often is to copy paste the module initialization code inside the single method, things like `linking.wast` won't be easy to replicate though.

re: splitting on module initialization:
this might work, but it has two problems:
- there are tests that depends on the initialization of 2+ different modules (look at `linking.wast`)
- we'd need to make a lot of changes to the codegen

I'd like to avoid changing or maintaining multiple versions of the codegen, it's an instrumental piece of this project and the simpler and most straightforward is to make changes to it, the faster the whole project can keep moving forward, implementing new proposals etc.

I'd rather wait for the upstream bug to be fixed (do you have a way to link at the grpc issue in a way where we get notifications?)

> My full time job is doing open source for the last 11 years

Very good to hear and it shows with your attitude :pray: 
Thanks, again, a lot for all your effort @yigit I think there are two possible ways to get things merged quickly:
- get `wasi-tests/WasiPreview1Test` to run
- fix the issue with `readAllBytes`

I'm travelling this week, but I'll eventually come back to those tasks if no one volunteer for them :slightly_smiling_face:

PR #748, opened 2025-01-27
Move test-gen-lib to it's own submodule

andreaTP:
To make it easier to run the testsuite in Gradle/other build systems.

cc. @yigit


evacchi (approved):
thanks!

PR #749, opened 2025-01-30
minor optimization of StackFrame instantiation

andreaTP:
Investigating #746 I noticed that the resizes of the `ArrayList` were consuming a lot of memory.


evacchi (approved):
excellent! 👏 👏 👏

PR #751, opened 2025-01-31
Memory improvements

andreaTP:
**Problem:**

Working on using Chicory with real-world Wasm payloads, I noticed a huge pressure on the `memory_grow` op code.
Basically, some C module, will keep growing the memory in very very small chunks, and, currently, this operation is really expensive as it needs a full copy to a new `ByteBuffer`.

**Proposal:**

While it's definitely possible to tweak the WASM module memory allocator and swap in something more optimized, it's much better for us to decouple the guest module consumption.
In this way, we can decide to plug in more advanced and sophisticated heuristics based on the usage.

**Implementation:**

The implementation puts in places the necessary pieces, the "Default" allocator is not greatly smarter than the previous, but it already improves performance in the use case I was looking at.
It should be a bit better overall, but it definitely has margin for improvement.

**Notes:**

I provided an implementation to preserve the current behavior when necessary, but this is a behavioral change and we should roll it (if approved) with a minor version bump (i.e. `1.1.X`).

@electrum : I'd be extremely interested if you take a little time to try out those changes with Python :pray:


andreaTP:
ping @electrum :pray:

andreaTP:
In absence of @electrum I do feel I still want to move this forward.

@bhelx / @evacchi  do you know if this is something other runtimes are doing?

evacchi:
this is somewhat related https://github.com/tetratelabs/wazero/issues/1331 
I don't see why memory.grow should be actually coupled to real memory allocation, as long as the observable behavior is correct :)

I think these are related too, the actual allocated memory is 8GB and Grow only changes the ProtectedLoad 
https://chromium.googlesource.com/v8/v8.git/+/eaed31c5f59abfb146c3ca82059ca1c41b1729b9
> GrowMemory is now implemented by changing the protection on the guard regions to
make the new portions of the heap accessible.
https://chromium.googlesource.com/v8/v8.git/+/0027218e1a0021af8729ed0d6c169d45b548f853

andreaTP:
Thanks for the references @evacchi , I think that's a confirm this is a good path to go down.
Now the question becomes the public API, it's not completely cleaned up, I see that in `wazero` they used a different encoding (including `cap` for example), happy to receive review and suggestions about it :pray:

runtime/src/main/java/com/dylibso/chicory/runtime/ByteBufferMemory.java
```diff
@@ -130,7 +146,7 @@ public void initPassiveSegment(int segmentId, int dest, int offset, int size) {
 
     @Override
     public void write(int addr, byte[] data, int offset, int size) {
-        checkBounds(addr, size, buffer.limit());
+        checkBounds(addr, size, (PAGE_SIZE * nPages));
```
electrum, line 149: Add a helper method to return the size?

runtime/src/main/java/com/dylibso/chicory/runtime/alloc/OneToOneMemAllocStrategy.java
```diff
@@ -0,0 +1,16 @@
+package com.dylibso.chicory.runtime.alloc;
+
+public final class OneToOneMemAllocStrategy implements MemAllocStrategy {
```
electrum, line 3: Maybe name this `ExactMemAllocStrategy`

runtime/src/main/java/com/dylibso/chicory/runtime/Memory.java
```diff
@@ -18,6 +18,14 @@ public interface Memory {
      */
     public static final int RUNTIME_MAX_PAGES = 32767;
 
+    static int bytes(int pages) {
+        if (pages >= RUNTIME_MAX_PAGES) {
+            return PAGE_SIZE * RUNTIME_MAX_PAGES;
+        } else {
+            return PAGE_SIZE * pages;
+        }
```
electrum, line 26:
```suggestion   
        return PAGE_SIZE * Math.min(pages, RUNTIME_MAX_PAGES);
```

runtime/src/main/java/com/dylibso/chicory/runtime/alloc/DefaultMemAllocStrategy.java
```diff
@@ -0,0 +1,23 @@
...
+    @Override
+    public int initial(int min) {
+        return min;
+    }
+
+    @Override
+    public int next(int current, int target) {
+        int next = (current <= 0) ? target : current;
+        while (next < target && next < max) {
+            next = next << 1;
```
electrum, line 19: Can you add a unit test for this class?

I think this will overflow when `current` is greater than `Integer.MAX_VALUE / 2` and return a negative value

electrum (approved):
I checked and this will help Python as well. It often grows memory one or two pages at a time.

runtime/src/main/java/com/dylibso/chicory/runtime/ByteBufferMemory.java
andreaTP: done

runtime/src/main/java/com/dylibso/chicory/runtime/alloc/OneToOneMemAllocStrategy.java
andreaTP: :+1: thanks for the good suggestion!

runtime/src/main/java/com/dylibso/chicory/runtime/Memory.java
andreaTP: :+1:

runtime/src/main/java/com/dylibso/chicory/runtime/alloc/DefaultMemAllocStrategy.java
andreaTP: Added the unit test, it doesn't break, please take a look and let me know if I missed any case.

andreaTP:
This PR stayed open long enough to let me bomb the original and more limited scope :sweat_smile: 

I spoke with @franz1981 about performance and he pointed me out to a few improvements that directly apply.
I included those improvements in this PR, there are 2(possibly) open questions I'd like the reviewers to provide feedback on:

- change the default Memory implementation to be `ByteArrayMemory`
- public API for the mem alloc strategy

runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
```diff
@@ -0,0 +1,350 @@
...
+import java.util.function.Function;
+
+/**
+ * Represents the linear memory in the Wasm program. Can be shared
+ * reference b/w the host and the guest.
+ */
+public final class ByteArrayMemory implements Memory {
+    // trick from:
+    // https://stackoverflow.com/a/65276765/7898052
+    private static final VarHandle SHORT_ARR_HANDLE =
```
andreaTP, line 29: We have been reluctant to use `VarHandle` and `MethodHandle` in the past, in this case I believe that the usage is justified:

- there is an alternative `Memory` implementation not leveraging those
- is the fastest way to access the underlying `byte[]` (confirmed benchmarking on Photon)
- the usage is extremely limited and scoped

runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
```diff
@@ -0,0 +1,350 @@
...
+
+    private int sizeInBytes() {
+        return PAGE_SIZE * nPages;
+    }
+
+    @Override
+    public void write(int addr, byte[] data, int offset, int size) {
+        try {
+            System.arraycopy(data, offset, buffer, addr, size);
+        } catch (RuntimeException e) {
```
andreaTP, line 179: Apparently, the JVM is able to optimize bound checks, thus is convenient to use `try` - `catch` than explicit checks.

We are catching `RuntimeException` to keep the method body short and easily inlinable.

evacchi: can you add this note as a comment (even in the heading of the class is fine)

aot/src/main/java/com/dylibso/chicory/experimental/aot/AotMethods.java
```diff
@@ -91,69 +91,55 @@ public static int memoryPages(Memory memory) {
...
-        memory.writeF64(base + offset, value);
+        memory.writeF64(getAddr(base, offset), value);
     }
 
-    public static void validateBase(int base) {
-        if (base < 0) {
-            throwOutOfBoundsMemoryAccess();
-        }
+    public static int getAddr(int base, int offset) {
+        return (base < 0) ? base : base + offset;
```
evacchi, line 142: maybe add a note here that we let the memory access throw if the base is actually negative.

runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
```diff
@@ -0,0 +1,350 @@
...
+import java.util.Arrays;
+import java.util.function.Function;
+
+/**
+ * Represents the linear memory in the Wasm program. Can be shared
+ * reference b/w the host and the guest.
+ */
+public final class ByteArrayMemory implements Memory {
+    // trick from:
+    // https://stackoverflow.com/a/65276765/7898052
```
evacchi, line 28:
```suggestion
    // get access to the byte array elements viewed as if it were 
    // a different primitive array type, such as int[], long[], etc.
    // This is actually the fastest way to access and reinterpret the underlying bytes.
    // see https://stackoverflow.com/a/65276765/7898052
```

runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
electrum: We use this in Trino and it works well.

andreaTP: Thanks for confirming, slightly surprised it works OOB with Android, cc. @yigit

yigit: nice. I've never dabbed into these areas so i have no idea.
One thing though, we are running android tests only with very recent versions so behavior might be different but since you have this abstracted, it would be possible to have a different default Memory implementation for older API levels if necessary.

evacchi: @yigit `ByteBufferMemory` should work on Android too :)

aot/src/main/java/com/dylibso/chicory/experimental/aot/AotMethods.java
andreaTP: done

runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
andreaTP: changed :+1:

PR #767, opened 2025-02-21
Add the Maven Wrapper and use Maven 4

andreaTP:
First off, this doesn't break the build using Maven 3, we keep using `mvn` and now also `mvnw`.

- Added a Maven Wrapper to make the environment better reproducible.
- Now the testing in CI happens with both versions

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -682,7 +682,12 @@
           <failFast>false</failFast>
           <rules>
             <banDuplicatePomDependencyVersions/>
-            <requireUpperBoundDeps/>
+            <requireUpperBoundDeps>
+              <excludes>
+                <!-- Need an update of Jimfs + Guava: https://github.com/google/jimfs/issues/394 -->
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -682,7 +682,11 @@
           <failFast>false</failFast>
           <rules>
             <banDuplicatePomDependencyVersions/>
-            <requireUpperBoundDeps/>
+            <requireUpperBoundDeps>
+              <excludes>
+                <exclude>org.checkerframework:checker-qual</exclude>
+              </excludes>
+            </requireUpperBoundDeps>
```
evacchi, line 689: why is this needed? can you add a note?

andreaTP: We need a newer release of Jimfs, added a note

jerrinot:
It looks like after the change, Intellij IDEA fails to import the project.

andreaTP:
Thanks for the feedback @jerrinot , rolling back to Maven 3 in #777

PR #781, opened 2025-03-03
Switch to wasm-tools instead of wabt

andreaTP:
In my quest to implement Exception Handling I found that the current version of `wabt` doesn't support the latest(and hopefully definitive) version of EH.

I got everything to work with `wasm-tools`, that seems to be more updated and closer to the WebAssembly specification.

there are a couple of downsides though:
- This is running in interpreted mode, which is slow, I have not been able to compile `wasm-tools` with the AOT, hitting all kind of limits
- The `wasm-tools.wasm` file is about 14.4Mb, hence I'm downloading it from the internet everytime. We are going to hit GH rate-limiting with this approach. I can try to play smarter with profiles, maybe ...


wasm/src/main/java/com/dylibso/chicory/wasm/Validator.java
```diff
@@ -1183,6 +1181,9 @@ void validateFunction(int funcIdx, FunctionBody body, FunctionType functionType)
                 case SELECT_T:
                     {
                         popVal(ValueType.I32);
+                        if (op.operands().length <= 0 || op.operands().length > 1) {
```
andreaTP, line 1184: this should be an unrelated but slight improvement of validation.

andreaTP:
Let's go with #782 for now.

PR #784, opened 2025-03-04
Split the precompiled AOT logic to a separate jar

andreaTP:
This is the agreed split of the precompiled AOT logic that have been discussed.

cc. @illarionov

### DIFF
--- a/aot-maven-plugin/src/main/java/com/dylibso/chicory/experimental/maven/aot/AotGenMojo.java
+++ b/aot-maven-plugin/src/main/java/com/dylibso/chicory/experimental/maven/aot/AotGenMojo.java
@@ -1,44 +1,9 @@
 package com.dylibso.chicory.experimental.maven.aot;
 
-import static com.dylibso.chicory.wasm.WasmWriter.writeVarUInt32;
-import static com.github.javaparser.StaticJavaParser.parseClassOrInterfaceType;
-import static com.github.javaparser.StaticJavaParser.parseType;
-
-import com.dylibso.chicory.experimental.aot.AotCompiler;
-import com.dylibso.chicory.runtime.Instance;
-import com.dylibso.chicory.runtime.Machine;
-import com.dylibso.chicory.wasm.Parser;
-import com.dylibso.chicory.wasm.WasmModule;
-import com.dylibso.chicory.wasm.WasmWriter;
-import com.dylibso.chicory.wasm.types.OpCode;
-import com.dylibso.chicory.wasm.types.RawSection;
-import com.dylibso.chicory.wasm.types.SectionId;
-import com.github.javaparser.ast.CompilationUnit;
-import com.github.javaparser.ast.Modifier.Keyword;
-import com.github.javaparser.ast.NodeList;
-import com.github.javaparser.ast.body.ClassOrInterfaceDeclaration;
-import com.github.javaparser.ast.body.VariableDeclarator;
-import com.github.javaparser.ast.expr.ClassExpr;
-import com.github.javaparser.ast.expr.MethodCallExpr;
-import com.github.javaparser.ast.expr.NameExpr;
-import com.github.javaparser.ast.expr.ObjectCreationExpr;
-import com.github.javaparser.ast.expr.StringLiteralExpr;
-import com.github.javaparser.ast.expr.VariableDeclarationExpr;
-import com.github.javaparser.ast.stmt.BlockStmt;
-import com.github.javaparser.ast.stmt.CatchClause;
-import com.github.javaparser.ast.stmt.ReturnStmt;
-import com.github.javaparser.ast.stmt.ThrowStmt;
-import com.github.javaparser.ast.stmt.TryStmt;
-import com.github.javaparser.utils.SourceRoot;
-import java.io.ByteArrayOutputStream;
+import com.dylibso.chicory.experimental.build.time.aot.Config;
+import com.dylibso.chicory.experimental.build.time.aot.Generator;
 import java.io.File;
 import java.io.IOException;
-import java.io.InputStream;
-import java.io.UncheckedIOException;
-import java.nio.file.Files;
-import java.nio.file.Path;
-import java.util.Map;
-import java.util.StringJoiner;
 import org.apache.maven.model.Resource;
 import org.apache.maven.plugin.AbstractMojo;
 import org.apache.maven.plugin.MojoExecutionException;
@@ -82,6 +47,14 @@ public class AotGenMojo extends AbstractMojo {
     private File targetSourceFolder;
 
     /**
+     * the target wasm folder to generate the stripped meta wasm module
+     */
+    @Parameter(
+            required = true,
+            defaultValue = "${project.build.directory}/generated-resources/chicory-aot")
+    private File targetWasmFolder;
+
+    /**
      * The current Maven project.
      */
     @Parameter(property = "project", required = true, readonly = true)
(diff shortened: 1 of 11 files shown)


evacchi (approved):

pom.xml
```diff
@@ -848,6 +853,7 @@
       <modules>
         <module>aot</module>
         <module>aot-maven-plugin</module>
+        <module>aot-precompiled</module>
```
evacchi, line 856: I am not 100% sure about the naming (it is _always_ AoT after all); personally I've been using "build-time" or "off-line". But I am approving anyway.

aot-precompiled/src/main/java/com/dylibso/chicory/experimental/precompiled/aot/Generator.java
```diff
@@ -0,0 +1,214 @@
...
+        this.config = config;
+    }
+
+    public byte[] wasmMetaModule() throws IOException {
+        byte[] wasmBytes = Files.readAllBytes(config.wasmFile());
+        var module = Parser.parse(config.wasmFile());
+        return rewriteWasm(wasmBytes, module);
+    }
+
+    public void resources() throws IOException {
```
electrum, line 56: A verb such as "generate" would sound better.

electrum (approved):

aot-precompiled/src/main/java/com/dylibso/chicory/experimental/precompiled/aot/Generator.java
```diff
@@ -0,0 +1,214 @@
...
+
+public class Generator {
+
+    private final Config config;
+
+    public Generator(Config config) {
+        this.config = config;
+    }
+
+    public byte[] wasmMetaModule() throws IOException {
```
illarionov, line 50: This method returns a byte array, but the user of the generator has no way to know where the resource file should be written so that the generated `<basename>Module.load()` method can load it correctly. 
Currently, the filename is hardcoded to `<basename>.meta` and the file is expected to be placed in the package directory (getPackageName()). 
Possible improvements:
* Either have `wasmMetaModule()` write the file (like `resources()` does).
* Or add a public method to the Generator that returns the expected path.

aot-precompiled/src/main/java/com/dylibso/chicory/experimental/precompiled/aot/Generator.java
```diff
@@ -0,0 +1,214 @@
...
+        byte[] wasmBytes = Files.readAllBytes(config.wasmFile());
+        var module = Parser.parse(config.wasmFile());
```
illarionov, line 52: (Minor nitpick) `Parser.parse` is called both in `wasmMetaModule()` and in `resources()`.  It could be cached in a lazy field within the Generator class to avoid redundant parsing. Not strictly necessary, I assume you split it intentionally since it likely doesn't impact performance.

aot-precompiled/src/main/java/com/dylibso/chicory/experimental/precompiled/aot/Generator.java
illarionov, line 56: The `resources()` method currently generates two  things:
* The `<baseName>Module.java` file.
* The `<basename>Machine.class` files.

These are effectively two independent generators that could even run in parallel.
 It feels a bit strange to have them in a single method, especially when there's a separate method for generating resources `resources()`.
Also, they have different dependencies — javaparser is only needed for generating `<baseName>Module.java`.
Maybe it would make sense to split this into two methods, or possibly merge `wasmMetaModule()` and `resources()` under a single public entry point.

aot-precompiled/src/main/java/com/dylibso/chicory/experimental/precompiled/aot/Config.java
```diff
@@ -0,0 +1,85 @@
...
+        private Path targetSourceFolder;
+
+        private Builder() {}
+
+        public Builder withWasmFile(Path wasmFile) {
+            this.wasmFile = wasmFile;
+            return this;
+        }
+
+        public Builder withName(String name) {
```
illarionov, line 66: Currently, the config takes a single `name` field, which is expected to be the full name including the target package, e.g. "com.example.HelloWorld".
Then, the Generator exposes two methods — `getPackageName()` and `getBaseName()` — that split this string into parts.
This format seems unintuitive to me. A possible alternative is to replace `name` in Config with two fields: `withPackageName()` and `withBaseName()`. That would allow us to remove those two methods. WDYT?

illarionov (commented):
With this API, I would need to remove the ability to customize the names of the generated Machine, Module, and meta files — but I think that’s the right decision. It’s unlikely anyone would need that flexibility anyway.
Overall, this looks good to me.

pom.xml
andreaTP: renamed to build time everywhere

aot-precompiled/src/main/java/com/dylibso/chicory/experimental/precompiled/aot/Config.java
andreaTP: Good idea, moved to `Config`

aot-precompiled/src/main/java/com/dylibso/chicory/experimental/precompiled/aot/Generator.java
andreaTP: Yeah, build time is not on a critical path, unless we have specific complaints I'd rather keep things more isolated.
I slightly improved one of the `parse` methods though, thanks for noticing.

andreaTP:
Thanks @illarionov for calling out my laziness in doing this :sweat_smile: really appreciated!
I think I have incorporated all of your comments, let me know how it looks now, it doesn't need to be "perfect" as it's an "experimental" module, but good enough and easy to evolve.

For the rest of the reviewers, I included a CLI that is handy mostly for us developers, I'm not going to document it.
With it, we can generate all the output of the AOT without the need to setup a full Maven/Gradle build.

illarionov:
Now it looks great, thank you @andreaTP !

aot-maven-plugin/src/main/java/com/dylibso/chicory/experimental/maven/aot/AotGenMojo.java
```diff
@@ -91,178 +64,28 @@ public class AotGenMojo extends AbstractMojo {
...
+        var generator = new Generator(config);
 
-        var rewrittenWasm = rewriteWasm(wasmBytes, module);
-        var newWasmFile =
-                targetClassFolder.toPath().resolve(packageName.replace('.', '/')).resolve(wasmName);
         try {
-            Files.write(newWasmFile, rewrittenWasm);
+            generator.generateMetaWasm();
+            generator.generateResources();
+            generator.generateSources();
```
electrum, line 81: Why not have a single `generate()` method? Having separate methods seems to leak implementation details to callers, and allows a mistake where you only generate partial resources which creates a broken artifact.

andreaTP: Thanks for following up @electrum , since we are in an experimental module we are still free to change the public API when necessary.

I did the split encouraged by [this comment](https://github.com/dylibso/chicory/pull/784#discussion_r1981195177) and, analizing more in detail:
- `generateResources`: this generates the self contained `.class` files
- `generateMetaWasm`: this generates the `.meta` wasm module(with the stripped function bodies), it's self contained and I do see ourselves making this more configurable/optional in the future (e.g. I need to opt-out [for observability](https://github.com/roastedroot/sqlite4j/compare/main...andreaTP:sqlite4j:observability#diff-1a493a32f09ee1536de69f5d02ec596c502848972fce51a73d0ecccecbc97861R47) )
- `generateSources`: this is the `.java` glue code that, indeed, allows for mistakes and can generate broken artifacts

I'm not 100% sure what's the best shape for this code, but I see value in being able to granularity produce the separate things and making them more configurable in the future.
An alternative approach would be to have a single `generate` with options but I'm open to alternatives.

cc. @illarionov

PR #792, opened 2025-03-07
Easy HostModule rename mechanism

andreaTP:
This adds a new generated overload where the module name can be supplied explicitly

```java
    public static HostFunction[] toHostFunctions(Simple functions, String moduleName) {
```

Looks like there are Wasm modules where the wasi function module is called `wasi_unstable`.

This change is backward-compatible and provides an easy mechanism for changing the module name to all the functions defined in a "HostModule".

cc. @chirino


evacchi:
so if I understand correctly this adds a new generated overload where the module name can be supplied explicitly

```
    public static HostFunction[] toHostFunctions(Simple functions, String moduleName) {
```

Can you add this note to the top post?

Other notes:

1. if the limitation is known ahead of time, then we could allow the user to supply the name at codegen time instead. What do you think?
2. this seems to address the fact that WasiP1 is known with an older name in some old Wasm modules. We could also specialize the case on `Wasi` instead (or in addition).

andreaTP:
> Can you add this note to the top post?

done

> 1.

the WASI module is built in this codebase, and we don't know what's needed when it's going to be used

> 2. We could also specialize

`WasiPreview1` is a final class at the moment and cannot be specialized.

andreaTP:
@evacchi have I missed any further question on this one?

evacchi (approved):

PR #801, opened 2025-03-17
feat(aot): tail call proposal

evacchi:
Because the JVM does not support tail calls natively, this only **desugars** RETURN_CALL into CALL+RETURN and RETURN_CALL_INDIRECT into CALL_INDIRECT+RETURN. 

However, there are indeed tests that exercise the stack; some of these _might_ be optimized (e.g. factorial of 1_000_000), but the others are mutually recursive functions that cannot be optimized in a general sense.  Notice that because of the massive size, even a custom stack size (in the `-Xss` sense) does not make much difference. 

Except for these cases (indeed those where proper tail calls would make sense 😅) the implementation passes all the other tests.

### DIFF
--- a/aot-tests/pom.xml
+++ b/aot-tests/pom.xml
@@ -152,7 +152,8 @@
             <wast>utf8-import-module.wast</wast>
             <wast>utf8-invalid-encoding.wast</wast>
           </includedWasts>
-          <excludedTests/>
+          <!-- Skip stack-overflows in tail calls -->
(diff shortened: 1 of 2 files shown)


andreaTP (approved):
Thanks @evacchi for looking into this!
I'm ok with this initial implementation that should help when porting EH.

To merge the only blocker is to open an Issue to track future development on tail call optimization in the AOT.

For example, seems that tail recursion on the same function can be implemented:
https://github.com/Sipkab/jvm-tail-recursion

For mutually recursive functions I'd like to see some explorations, e.g.: inlie all the mutually recursive functions in a single one etc.

andreaTP (approved):
:+1: good to go, thanks!

electrum (approved):

PR #802, opened 2025-03-17
Clear the call stack when performing a return

andreaTP:
I found this when working on #794 , and it's extremely confusing, so separating the fix to get it merged early.


evacchi (approved):
makes sense 🤔

PR #805, opened 2025-03-18
Allow import overloaded aliases

### LINKED ISSUE #795: Multiple imports of same name will fail to import
Having multiple host function under e.g. namespace "logger" name "log", but with different types, will fail to import, despite being legal wasm.

https://webassembly.github.io/spec/core/syntax/modules.html#imports

> Unlike export names, import names are not necessarily unique. It is possible to import the same module/name pair multiple times; such imports may even have different type descriptions, including different kinds of entities. A module with such imports can still be instantiated depending on the specifics of how an [embedder](https://webassembly.github.io/spec/core/intro/overview.html#embedder) allows resolving and supplying imports. However, embedders are not required to support such overloading, and a WebAssembly module itself cannot implement an overloaded name.

andreaTP:
Fix #795 

Allow multiple aliases of the same import with different types, and map those correctly.


runtime/src/main/java/com/dylibso/chicory/runtime/TagInstance.java
```diff
@@ -1,22 +1,31 @@
...
+import com.dylibso.chicory.wasm.types.FunctionType;
 import com.dylibso.chicory.wasm.types.TagType;
 
 public class TagInstance {
 
     private final TagType tag;
-    private final Instance instance;
+    private FunctionType type;
 
-    public TagInstance(TagType tag, Instance instance) {
```
andreaTP, line 10: This is an API breaking change, but I believe that the risk is really low.
`Tag`s are used by the EH proposal and ppl could not have been using this API for anything useful.
The EH PR might be impacted and iterate on this detail though.

andreaTP:
Ready for review :pray:

evacchi (approved):

andreaTP:
Thanks for the review @evacchi :pray:

PR #809, opened 2025-03-20
Switch to wasm-tools

andreaTP:
Sadly Wabt [doesn't support WasmGC](https://github.com/WebAssembly/wabt/issues/2348), so, even to start working toward it we need to switch to `wasm-tools` to use `wast2json`.

cc. @hypercubestart found this issue when starting to look at reference types/wasm GC


andreaTP:
Note, this change is impacting the build time as I was only able to run `wasm-tools` with the interpreter.
If anyone want to give it another try, this is where I left: https://github.com/WebAssembly/binaryen/issues/7340

evacchi (approved):
as long as the execution time is not massive (and it is not) I'd say it's fine!

Issue #813, opened 2025-03-27
typed function references (proposal)


hypercubestart:
Issue for tracking typed function references proposal.

Relevant links:
https://github.com/WebAssembly/gc/blob/main/proposals/function-references/Overview.md
https://webassembly.github.io/function-references/core/appendix/algorithm.html

This a blocker for WASMGC

andreaTP:
Thanks a lot for this effort @hypercubestart !
I'm looking forward for the next steps for WasmGC!

I believe we can close this issue.
The changes are not implemented in the compiler but we can open a new issue for it when time comes.

PR #814, opened 2025-03-27
(proposal) Typed Function References Part 1: ValueType Changes

hypercubestart:
Introduces necessary changes to ValueType class to support function index type.

This change is purely syntactic and shouldn't change any existing behavior

#813


wasm/src/main/java/com/dylibso/chicory/wasm/types/ValueType.java
```diff
@@ -2,31 +2,75 @@
...
+    public static ValueType I32 = new ValueType(ValueTypeOpCode.I32);
+
+    public static ValueType V128 = new ValueType(ValueTypeOpCode.V128);
+    public static ValueType FuncRef = new ValueType(ValueTypeOpCode.FuncRef);
+    public static ValueType ExternRef = new ValueType(ValueTypeOpCode.ExternRef);
+
+    private final Variant opcode;
+
+    // some value types have an argument, conveniently this fits in an int
+    private final int operand;
```
andreaTP, line 30: This is named `typeidx` or `argument` in the spec, correct?
Naming is hard but keeping it closer to the spec makes everything more easily readable.

wasm/src/main/java/com/dylibso/chicory/wasm/types/ValueType.java
```diff
@@ -2,31 +2,75 @@
...
+        this.opcode = var;
+        this.operand = operand;
     }
 
     /**
-     * @return the numerical identifier for this type
+     * @return the numerical identifier for this type. Conveniently, all value types we want to represent
+     *     for now (constructor + arguments) can fit inside a Java long.
+     *
+     *     We store as operand in the MSB and the opcode in the LSB.
```
andreaTP, line 66: This is fine for me, but I'm a little confused, either we keep the "operand" in the `id` or as a separate field, but not both.
Any specific reason for this choice?

wasm/src/main/java/com/dylibso/chicory/wasm/types/ValueType.java
```diff
@@ -175,20 +222,101 @@ public static int sizeOf(List<ValueType> args) {
...
+        return Objects.hash(opcode, operand);
+    }
+
+    @Override
+    public boolean equals(Object other) {
+        if (!(other instanceof ValueType)) {
+            return false;
+        }
+        ValueType that = (ValueType) other;
+        return this.opcode == that.opcode && this.operand == that.operand;
```
andreaTP, line 236: If I'm follow correctly:

```java
return this.id() == that.id();
```

wasm/src/main/java/com/dylibso/chicory/wasm/types/ValueType.java
```diff
@@ -175,20 +222,101 @@ public static int sizeOf(List<ValueType> args) {
...
+        return opcode.name();
+    }
+
     /**
-     * A separate holder class for ID constants.
-     * This is necessary because enum constants are initialized before normal fields, so any reference to an ID constant
-     * in the same class would be considered an invalid forward reference.
+     * basically ValueTypeOpcode, but removes FuncRef and ExternRef, which alias Ref/RefNull
      */
-    static final class ID {
```
andreaTP, line 183: I think is possible to keep this class here, it will be better for backward compat.

wasm/src/main/java/com/dylibso/chicory/wasm/types/ValueType.java
```diff
@@ -175,20 +222,101 @@ public static int sizeOf(List<ValueType> args) {
...
-        private ID() {}
-
-        static final int ExternRef = 0x6f;
-        static final int FuncRef = 0x70;
-        static final int V128 = 0x7b;
-        static final int F64 = 0x7c;
-        static final int F32 = 0x7d;
-        static final int I64 = 0x7e;
-        static final int I32 = 0x7f;
+    public enum Variant {
```
andreaTP, line 259: I'm a little confused by this abstraction:

```java
ValueType(ValueTypeOpCode, int operand);
Variant(id) == Variant(ValueTypeOpCode);
```

do we need `Variant` at all? Seems to encode a concept already included in `ValueType` or am I missing something?

wasm/src/main/java/com/dylibso/chicory/wasm/types/ValueTypeOpCode.java
```diff
@@ -0,0 +1,82 @@
...
+
+    private final int opcode;
+
+    // trick: the enum constructor cannot access its own static fields
+    // but can access another class
+    private static final class ValueTypeOpCodes {
+        private ValueTypeOpCodes() {}
+
+        private static final Map<Integer, ValueTypeOpCode> byOpCode = new HashMap<>();
+        private static final Map<Integer, List<WasmEncoding>> signatures = new HashMap<>();
```
andreaTP, line 27: I understand why you did it, but, in this case we have exactly 2 special cases:
- `Ref`
- `RefNull`

Instead of keeping around an additional Map I'd prefer to directly handle the different encoding explicitly in the `Parser`.

wasm/src/main/java/com/dylibso/chicory/wasm/types/ValueTypeOpCode.java
```diff
@@ -0,0 +1,82 @@
...
+    Ref(ID.Ref, List.of(WasmEncoding.VARSINT32));
+
+    private final int opcode;
+
+    // trick: the enum constructor cannot access its own static fields
+    // but can access another class
+    private static final class ValueTypeOpCodes {
+        private ValueTypeOpCodes() {}
+
+        private static final Map<Integer, ValueTypeOpCode> byOpCode = new HashMap<>();
```
andreaTP, line 26: Instead of having this map we can use a `switch` case in the `byOpCode` method, I think ...

aot/src/main/java/com/dylibso/chicory/experimental/aot/TypeStack.java
```diff
@@ -34,14 +34,14 @@ public void push(ValueType type) {
 
     public void pop(ValueType expected) {
         var actual = types().pop();
-        if (expected != actual) {
+        if (!expected.equals(actual)) {
```
andreaTP, line 37: Do you have a "rule of thumb" to follow on when comparing `opcodes` only and when using `equals`?

andreaTP (commented):
Thanks a lot for the effort @hypercubestart and sorry for the delay in this review.
It's a great start! I made a few comments(mostly stylistic nitpicks) and questions.

Let me know if you need any help in rebasing etc.

aot/src/main/java/com/dylibso/chicory/experimental/aot/TypeStack.java
hypercubestart: I think as a rule of thumb, use `equals` now that ValueType is an object (this is equivalent to comparing ValueType.id()). I'm not sure of a case where we only want to compare opcodes, possibly when the code looks like:

```
private Class javaClassFromValueType(ValueType type) {
        switch (type.opcode()) {
            case I32:
                return int.class;
            case I64:
                ....
```

so enumerating based on the type

andreaTP: looks good :+1:

wasm/src/main/java/com/dylibso/chicory/wasm/types/Table.java
```diff
@@ -1,12 +1,20 @@
...
     private final TableLimits limits;
 
     public Table(ValueType elementType, TableLimits limits) {
+        this(
+                elementType,
+                limits,
+                List.of(new Instruction(-1, OpCode.REF_NULL, new long[] {elementType.typeIdx()})));
+    }
+
+    public Table(ValueType elementType, TableLimits limits, List<Instruction> init) {
```
andreaTP, line 17: `init` is not used in this PR, is it preparation for the follow up?

andreaTP (changes requested):
I have not looked deeply into the "backward compatibility" aspect, but this is really going in a good direction! Thanks a lot!

I left a couple of comments and sent a demo PR for changing an aspect(since we are at it ...).
When ready, I encourage you to run in the root of the project:

```
./mvnw spotless:apply clean install
```

wasm/src/main/java/com/dylibso/chicory/wasm/types/Table.java
hypercubestart: yep! in part 2 this might be the result of a constant expression

wasm/src/main/java/com/dylibso/chicory/wasm/types/FunctionType.java
```diff
@@ -1,6 +1,6 @@
 package com.dylibso.chicory.wasm.types;
 
-import java.util.EnumMap;
```
andreaTP, line 3: Do you mind to incorporate: https://github.com/hypercubestart/chicory/pull/1 ?

wasm/src/main/java/com/dylibso/chicory/wasm/types/ValueType.java
```diff
@@ -1,51 +1,79 @@
...
 
-import static com.dylibso.chicory.wasm.types.ValueType.ID.ExnRef;
-
 import com.dylibso.chicory.wasm.MalformedException;
 import java.util.List;
 
 /**
  * The possible WASM value types.
  */
-public enum ValueType {
```
andreaTP, line 11: This is the biggest breaking change in this PR, from the binary point of view it's going to break, still, the source compatibility is preserved.

I'm looking for advise from @evacchi and @chirino :
after we merge this one, should we bump the major or is it ok to bump only the minor version?

evacchi: even though it's a breaking ABI change I wouldn't call for a major version. Let's highlight it in the release notes.

andreaTP: The more I think about this the more I think it's a bit risky.
`ValueType.XXX` is exposed and used by almost anyone using Chicory, and I'm afraid that this change would break code :cold_sweat:

chirino: yeah.. expressions like `switch(valueType)` would break.

hypercubestart: this is a good point. Even refactoring within chicory seemed to break random places that did `valueType == ValueType.V128`. I think you had a question before about using .equals or comparing opcode @andreaTP , I think I'm now leaning towards always using opcodes to compare and removing the `ValueType.XXX` static constants

evacchi: mm yeah I didn't think about that :D

andreaTP: > removing the ValueType.XXX static constants

I'm not sure this is a good idea, I really like being able to use static constants when possible.

That said, this means that we need to plan a ahead on how to make this work, the problem is that this is extensive and maintaining the diff in a branch is going to be a lot of work.
On the other side merging means that we need to release a 2.X as next.

I'll think a bit about this, happy to hear feedback!

andreaTP: @hypercubestart sorry for the delay we brainstormed a bit on this during yesterday's meeting, I think I have a plan to try to make this change not break the API.

wasm/src/main/java/com/dylibso/chicory/wasm/types/Table.java
andreaTP: :+1:

wasm/src/main/java/com/dylibso/chicory/wasm/types/ValueType.java
andreaTP: Finalized the idea we had yesterday:
https://github.com/hypercubestart/chicory/pull/2

cc. @chirino @evacchi can you double-check that I have not missed something?

andreaTP: resolved

andreaTP:
As soon as we release 1.3.0 this is ready to be merged on `main`

andreaTP (approved):
LGTM!

Thanks a lot for the amazing work @hypercubestart !
Ready to merge, but, if anyone of the maintainers spot a compatibility issue we should fix it on top asap(should be easy enough though).

PR #815, opened 2025-03-27
(proposal) Typed Function References Part 2: Passing testsuite

hypercubestart:
Makes necessary changes to validator, parser, etc. to pass the typed function testsuite

#813


andreaTP:
Sorry for the delay @hypercubestart , this is ready to be rebased on latest `main`, feel free to ping if you need any help!

wasm/src/main/java/com/dylibso/chicory/wasm/types/ValType.java
```diff
@@ -229,8 +315,12 @@ public String name() {
 
     public enum TypeIdxCode {
         // heap type
-        EXTERN(0x6F),
-        FUNC(0x70);
+        EXTERN(-17), // 0x6F
+        FUNC(-16), // 0x70
+        EXN(-23), // 0x69
+        BOT(-1),
```
andreaTP, line 321: Back to `UNKNOWN` ?

wasm/src/main/java/com/dylibso/chicory/wasm/types/ValType.java
```diff
@@ -229,8 +315,12 @@ public String name() {
...
+        // definedType
+        DEFINED(-2);
```
andreaTP, line 323: I don't find this `DEFINED` type in the spec, seems like an implementation detail that spills to the user API.
Do you have any reference? Is it necessary or do we have alternative ways to achieve the same?

wasm/src/main/java/com/dylibso/chicory/wasm/types/ValType.java
```diff
@@ -281,7 +376,7 @@ public static String toName(int opcode) {
                     return "I32";
             }
 
-            throw new IllegalArgumentException("got invalid opcode in ValType.toName: " + opcode);
+            throw new IllegalArgumentException("got invalid opcode in ValueType.toName: " + opcode);
```
andreaTP, line 379: nit: revert

wasm/src/main/java/com/dylibso/chicory/wasm/types/ValType.java
```diff
@@ -191,6 +207,69 @@ public static int sizeOf(List<ValType> args) {
         return total;
     }
 
+    public ValType substitute(WasmModule context) {
```
andreaTP, line 210: Is it useful to have the entire `WasmModule` here?
What about using only the `TypeSection`?

andreaTP (changes requested):
@hypercubestart amazing job, getting closer and closer to be merged.

I asked a few questions but only 2 are blocking imho:

- rename from `UNKNOWN` to `BOT`, either there is a good reason or I'd ask to revert
- the usage of the `DEFINED(-2)` type id, do you have a reference in the spec? Alternatives?

wasm/src/main/java/com/dylibso/chicory/wasm/types/ValType.java
```diff
@@ -11,7 +13,7 @@ public final class ValType {
     private static final long OPCODE_MASK = 0xFFFFFFFFL;
     private static final long TYPEIDX_SHIFT = 32;
 
-    public static ValType UNKNOWN = new ValType(ID.UNKNOWN);
+    public static ValType BOT = new ValType(ID.BOT);
```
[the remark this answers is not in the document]
andreaTP: Thanks for the link!
It irks me, but your reasoning is valid, let's go with `Bot` :cry:

wasm/src/main/java/com/dylibso/chicory/wasm/types/ValType.java
andreaTP: `resolvedFunctionType = null` works for me as a backup option.

The builders plus an explicit resolution phase sounds really appealing though.
I'm not really concerned to deviate from the reference implementation if it's good design and passes the testsuite :-)

andreaTP (approved):
Awesome job @hypercubestart !

This is around since enough time and is a great step forward, there are details that might be improved but I'll follow up with PRs on top.

LGTM, thanks a lot again for the great contribution!

andreaTP:
Sent this [PR](https://github.com/hypercubestart/chicory/pull/4) to rework a bit the Builder logic, but it's not mandatory to merging this PR (I can follow up in a separate PR).

andreaTP:
Merged! Thanks a lot for the effort @hypercubestart ! What's next? :slightly_smiling_face:

Issue #830, opened 2025-04-10
AOT java.lang.IllegalArgumentException: bad parameter count


andreaTP:
If a Wasm Module contains a function with more than 255 parameters the AOT compiler fail since it's hitting the maximum number of parameters of a Java function.

In this case we should be able to fallback to `long[]` and forth and back conversions, is suboptimal but it's the same compromise Java does when using varargs.

PR #831, opened 2025-04-10
fix: #830 AOT java.lang.IllegalArgumentException: bad parameter count.

chirino:
If a wasm function has too many parameters, falls back to passing the parameters via long[].


aot/src/main/java/com/dylibso/chicory/experimental/aot/AotUtil.java
```diff
@@ -140,8 +140,14 @@ public static MethodType methodTypeFor(FunctionType type) {
         return rawMethodTypeFor(type).appendParameterTypes(Memory.class, Instance.class);
     }
 
+    public static boolean hasTooManyParameters(FunctionType type) {
+        return type.params().stream().mapToInt(AotUtil::slotCount).sum() > 253;
```
andreaTP, line 144: nitpick: please define a static value for `253` with the rationale in a comment.

aot/src/main/java/com/dylibso/chicory/experimental/aot/AotCompiler.java
```diff
@@ -688,6 +710,18 @@ private void compileFunction(
...
+                asm.load(0, OBJECT_TYPE);
+                asm.iconst(i);
+                asm.aload(LONG_TYPE);
+                emitLongToJvm(asm, param);
+                asm.store(ctx.localSlotIndex(i), asmType(param));
+            }
+        }
+
         // initialize local variables to their default values
         int localsCount = type.params().size() + body.localTypes().size();
```
andreaTP, line 726: Should `type.params().size()` be 1 when we use `hasTooManyParameters` ?

andreaTP (commented):
For visibility, I added a couple of comments but the diff looks good.
Though we are missing something since now we hit:

```
Caused by: java.lang.ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0
        at org.objectweb.asm.Frame.merge(Frame.java:1280)
        at org.objectweb.asm.Frame.merge(Frame.java:1255)
        at org.objectweb.asm.MethodWriter.computeAllFrames(MethodWriter.java:1612)
        at org.objectweb.asm.MethodWriter.visitMaxs(MethodWriter.java:1548)
        at org.objectweb.asm.MethodVisitor.visitMaxs(MethodVisitor.java:787)
        at com.dylibso.chicory.experimental.aot.AotCompiler.emitFunction(AotCompiler.java:314)
```

there is, probably some missing detail here.

aot/src/main/java/com/dylibso/chicory/experimental/aot/AotUtil.java
chirino: done.

aot/src/main/java/com/dylibso/chicory/experimental/aot/AotCompiler.java
chirino: I wasn't too sure. but I think your right I updated it and the test still passes.

andreaTP (approved):
LGTM, thanks a lot for fixing this!

aot/src/main/java/com/dylibso/chicory/experimental/aot/AotUtil.java
andreaTP: thanks!

PR #839, opened 2025-04-15
tests: add tests to verify that we can generate a large wasm from a wat file.

chirino:


chirino:
This test case shows that the aot failing for wat2wasm on SOME jvms.  very weird.

chirino:
The same problem is showing up in https://github.com/dylibso/chicory/pull/833

PR #840, opened 2025-04-16
wasm-tools wat2wasm utility

andreaTP:
There is a mysterious bug happening only in compiled mode when specific versions of Java are used, this PR adds another way to use `wat2wasm` with the interpreter mode.

To unblock #833 @chirino you can now use this one (please keep the numbers low as it's taking forever to finish ~2 minutes).


andreaTP:
FYI this is not failing on Java 17 Temurin as the compiled code:
https://github.com/andreaTP/chicory/actions/runs/14496475456

chirino (approved):
lgtm

PR #841, opened 2025-04-16
feat: support passing execution options in wat2wasm.

chirino:


chirino:
Is there a downside to the PR.  We may need those options in the future.

andreaTP:
I don't see a current use and don't foresee any, I'd rather rebase this PR when is needed.
Thanks a lot for your effort! Appreciated :pray:

PR #842, opened 2025-04-16
Add test that illustrates that Aot doing memory access slightly differently than the interpreter.

chirino:
I assume we want the memory access to be consistent between the two.


chirino:
Seems like the main reason the memory access is different is due to the Interpreter storing all stack values as long values instead of the actual value type.   I'm going to see if I can have the LockStepMemory event track remap events to a logical equivalent.

andreaTP:
Do you plan to work more on this now that we have pin pointed the root cause of the original issue?
I believe that this is a good exploration(don't cancel the branch!) but I would not merge to `main` unless there is a compelling reason.

chirino:
Yeah not strictly needed.

Issue #843, opened 2025-04-17
JVM issue with `memoryCopy`


andreaTP:
This is a tracking Issue for a JVM bug.

When the method:

```
com.dylibso.chicory.experimental.aot.AotMethods.memoryCopy
```

is inlined on some JVMs(like Temurin 17) a test is crashing.

The issue is demonstrated in: #839

changing the configuration of `surefire` to disable the in-lining of the single method "fix" the problem:

```xml
        <plugin>
          <groupId>org.apache.maven.plugins</groupId>
          <artifactId>maven-surefire-plugin</artifactId>
          <version>${maven-surefire-plugin.version}</version>
          <configuration>
            <argLine>-XX:CompileCommand=dontinline,com/dylibso/chicory/wabt/Wat2WasmMachine$AotMethods.memoryCopy</argLine>
          </configuration>
        </plugin>
```

andreaTP:
Verified again #755 after those findings, and it works correctly after excluding `I32_GE_U`:

```xml
<configuration>
  <argLine>-XX:CompileCommand=dontinline,com/dylibso/chicory/runtime/OpcodeImpl.I32_GE_U</argLine>
</configuration>
```

chirino:
for anyone interested in reproducing the issue on the main branch:

Apply this patch to disable our current workaround:

```
diff --git a/runtime/src/main/java/com/dylibso/chicory/runtime/MemCopyWorkaround.java b/runtime/src/main/java/com/dylibso/chicory/runtime/MemCopyWorkaround.java
index a50318db..3feff819 100644
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/MemCopyWorkaround.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/MemCopyWorkaround.java
@@ -13,7 +13,8 @@ public final class MemCopyWorkaround {
     }
 
     public static boolean shouldUseMemWorkaround() {
-        return shouldUseMemWorkaround(System.getProperty("java.version"));
+        return false;
+        // return shouldUseMemWorkaround(System.getProperty("java.version"));
     }
 
     public static boolean shouldUseMemWorkaround(String version) {
```

Then build chicory using  one of the affected JDKs like Temurin-17

This should result in a build failure like:

```
[ERROR] Tests run: 5, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 3.415 s <<< FAILURE! -- in com.dylibso.chicory.wabt.Wat2WasmTest
[...]

PR #844, opened 2025-04-17
Workaround failing `memoryCopy` when inlined on some JVMs

andreaTP:
This is a workaround for #843 
Tradeoffs:

- it always kicks in for Java < 21
- it can be turned globally on or off using `System.properties`


andreaTP:
the failing test on Win 11 is proving me wrong :sweat: reverting to draft

andreaTP:
this should be reliable enough and self contained, cc. @chirino

aot/src/main/java/com/dylibso/chicory/experimental/aot/AotMethods.java
```diff
@@ -64,8 +66,45 @@ public static void tableInit(
...
+        // up to Java 17 the bug happens on various platforms we need to be conservative
+        if (enableMemCopyWorkaround) {
+            notInlinableMemoryCopy(destination, offset, size, memory);
+        } else {
+            memory.copy(destination, offset, size);
+        }
+    }
+
+    // HACK: this is a whole trick to prevent incorrect inlining of the JVM
+    private static ReadWriteLock lock = new ReentrantReadWriteLock();
```
evacchi, line 92: with the println, is the lock still needed?

aot/src/main/java/com/dylibso/chicory/experimental/aot/AotMethods.java
```diff
@@ -64,8 +66,45 @@ public static void tableInit(
...
+    // This is an ugly hack to workaround a bug on some JVMs (Temurin 17-)
+    private static final boolean enableMemCopyWorkaround;
+
+    static {
+        String workaround = System.getProperty("chicory.enableMemCopyWorkaround");
+
+        if (workaround != null) {
+            enableMemCopyWorkaround = Boolean.parseBoolean(workaround);
+        } else {
+            enableMemCopyWorkaround = Runtime.version().feature() < 21;
```
evacchi, line 78: why not <=17 ?

evacchi (approved):
just nits, looks good

aot/src/main/java/com/dylibso/chicory/experimental/aot/AotMethods.java
chirino: The lock is not needed if we implement the workaround  using megamorphic callsites.  
see: https://github.com/andreaTP/chicory/pull/14

WIth this lock based implemenation, all JDKs < 21 will have a big performance hit since all memcopy operations across ALL instances will contend on this one lock.

The downside of that alternate workaround is the runtime will need to hold a temporary class that implements the workaround until the JVM implement a fix.  I tend to think that's better than reducing the performance experience.

andreaTP: I'm open to go with your proposal @chirino , probably the deal breaker would be running a few benchmarks and comparing results

chirino: I'll try to whip something up.

andreaTP: Quick mention, do we need to copy over the compiled additional class? Probably yes

chirino: > do we need to copy over the compiled additional class

The workaround class was placed in Runtime so it's available without emitting it as part of the code generation.

andreaTP: Cool, thank you for taking the time to do the checks and clean things up!

Issue #853, opened 2025-04-23
Quarkus Native Build fails with: `Error: Detected an instance of Random/SplittableRandom class in the image heap.`


chirino:
```
Error: Detected an instance of Random/SplittableRandom class in the image heap. Instances created during image generation have cached seed values and don't behave as expected. If these objects should not be stored in the image heap, you can use 

    '--trace-object-instantiation=java.security.SecureRandom'

to find classes that instantiate these objects. Once you found such a class, you can mark it explicitly for run time initialization with 

    '--initialize-at-run-time=<culprit>'

to prevent the instantiation of the object.
The object was probably created by a class initializer and is reachable from a static field. You can request class initialization at image runtime by using the option --initialize-at-run-time=<class-name>. Or you can write your own initialization methods and call them explicitly from your main entry point.
Trace: Object was reached by
  reading field com.dylibso.chicory.wasi.WasiPreview1.random of constant 
    com.dylibso.chicory.wasi.WasiPreview1@733e5e76: com.dylibso.chicory.wasi.WasiPreview1@733e5e76
  reading field io.roastedroot.proxywasm.internal.ProxyWasm.wasi of constant 
    io.roastedroot.proxywasm.internal.ProxyWasm@736d957e: io.roastedroot.proxywasm.internal.ProxyWasm@736d957e
  reading field io.roastedroot.proxywasm.internal.Plugin.wasm of constant 
    io.roastedroot.proxywasm.internal.Plugin@49e696fc: io.roastedroot.proxywasm.internal.Plugin@49e696fc
  reading field io.roastedroot.proxywasm.internal.Pool$SharedPlugin.plugin of constant 
    io.roastedroot.proxywasm.internal.Pool$SharedPlugin@7d9102b8: io.roastedroot.proxywasm.internal.Pool$SharedPlugin@7d9102b8
  reading field java.util.HashMap$Node.value of constant 
 ```

PR #855, opened 2025-04-24
Switch default to ThreadLocalRandom in WasiOpts

andreaTP:
Fix for: #853

It took forever to figure out the best solution.
Handling of `ThreadLocalRandom` is already baked in GraalVM so we can safely leverage it.

note to self:
I have no clue why the relevant substitutions for all the standard library `Random` implementation are not provided by default. No one wants a Random generator that is seeded at compile time ...


andreaTP:
cc. @chirino

chirino (approved):
Works great!

evacchi (approved):
🚀

PR #857, opened 2025-04-24
feat: Extract source line numbers from wasm module to fill SMAP information in AoT class files.

chirino:
Pushing draft PR to share the progress.

I now have class files getting emitted with SMAP debug info.   This is similar to what JSPs do.

todo: test debugging /w different IDEs to see if these SMAP help the experience.


andreaTP:
FYI: there is a pure Java Dwarf "library" in GraalVM:

https://github.com/oracle/graal/tree/master/substratevm/src/com.oracle.objectfile/src/com/oracle/objectfile/elf/dwarf

I haven't attempted to extract it yet though ...

PR #859, opened 2025-04-25
chore: use cargo to build rust corpus examples

chirino:


wasm-corpus/src/main/resources/rust/greet-wasi/Cargo.toml
```diff
@@ -0,0 +1,3 @@
+[package]
+name = "greet-wasi"
+version = "0.0.1"
```
andreaTP, line 3: is:
```toml
[lib]
crate-type = ["cdylib"]
```
optional?

andreaTP (approved):
One question but the change looks good to me, thanks for this improvement!

PR #864, opened 2025-05-01
Make the generated FuncGroup and Dispatch classes nested.

chirino:


andreaTP:
Is this PR fixing a bug?

In general, we have had issues with nested classes in the past.
What do you think about removing the nesting at all and linearizing at the top level?

chirino:
I think it was a bug since we were doing:

```
        for (String name : extraClasses.keySet()) {
            classWriter.visitNestMember(internalClassName(name));
        }
```

This means we missed some of the corresponding `visitNestHost` calls.

andreaTP:
@electrum any insight here?

@chirino what about using only top level classes?

chirino:
We could do it either way.  Seems we were going to nested before.  Is there a reason to not continue doing it that way?

andreaTP:
> Is there a reason to not continue doing it that way?

We have found a few issues and idiosyncrasy in generating nested classes before ( cc. @electrum ).
The safest path would be to use top level classes only, and I honestly cannot think about a scenario where an inner class has strong advantages, especially when generated.

chirino:
This commit https://github.com/dylibso/chicory/pull/860/commits/2adcb1e4b9cc503ecf04f9b3536651ba795815e6 in #860 implements the reverse of what we are doing in this PR.  Making all extra classes top-level.  Either one works for me.   But I don't think we should use both styles.

electrum:
There’s no technical difference with nested classses, but it makes the namespacing cleaner and it is easier to have the correct visibility for the public API.

andreaTP:
If there is no technical limitation let's go with the most straightforward approach to keep everything in top classes.

electrum:
Having a huge number of top level classes can be annoying from a code completion perspective in the IDE. It's cleaner to have nested classes.

andreaTP:
The IDE is already broken on the generated classes.

electrum:
I'm referring to autocomplete when referencing the generated classes. For example, in Trino, we generate using
```xml
<name>io.trino.wasm.python.Python</name>
```
If you autocomplete `Python`, it suggests only `PythonMachine` and `PythonModule`. If all of the additional classes were top-level classes, then it's likely they would be shown by the IDE. 

It's ugly to have a bunch of class names like `CompiledMachineFuncGroup_0` as top-level classes. You probably wouldn't do this if you were writing it by hand, so we shouldn't do it in the generated code.

andreaTP:
I updated IntelliJ and checked things again.
I'm going to merge #874 as it move forward #860 , if I missed any invariant, please open an issue and we can re-iterate when necessary.

PR #868, opened 2025-05-02
Report Wat2Wasm errors in the thrown exception

### LINKED ISSUE #867: [wabt] If you parse an invalid wat file with `Wat2Wasm.parse()` the cause is not included in the thrown com.dylibso.chicory.wasi.WasiExitException

andreaTP:
Fix #867


andreaTP:
> I think I would have created a new type of RuntimeException for this

Makes sense, done, thanks for the suggestion :+1:

PR #869, opened 2025-05-02
feat[AoT]: add .builder() method creates an Instance.Builder for you 

### LINKED ISSUE #866: Aot: add `.builder()` method creates an Instance.Builder for you
This should allow code that looks like:

```java
                    Instance.builder(MODULE)
                            .withMachineFactory(Wat2WasmModule::create)
                            .withImportValues(imports)
                            .build();
```

to get simplified to:

```java
                    Wat2WasmModule.builder()
                            .withImportValues(imports)
                            .build();
```

This will require the `load()` to lazy load the module since we don't want to keep parsing the module every time a builder() is created.

chirino:
We also simplify the java source generation by parsing it from an initial java template And just replace some bits in it.

Fixes #866

### DIFF
--- a/aot-build-time/pom.xml
+++ b/aot-build-time/pom.xml
@@ -21,10 +21,6 @@
     </dependency>
     <dependency>
       <groupId>com.dylibso.chicory</groupId>
-      <artifactId>runtime</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>com.dylibso.chicory</groupId>
       <artifactId>wasm</artifactId>
     </dependency>
     <dependency>
--- a/aot-build-time/src/main/java/com/dylibso/chicory/experimental/build/time/aot/Generator.java
+++ b/aot-build-time/src/main/java/com/dylibso/chicory/experimental/build/time/aot/Generator.java
@@ -5,35 +5,21 @@ import static com.github.javaparser.StaticJavaParser.parseClassOrInterfaceType;
 import static com.github.javaparser.StaticJavaParser.parseType;
 
 import com.dylibso.chicory.experimental.aot.AotCompiler;
-import com.dylibso.chicory.runtime.Instance;
-import com.dylibso.chicory.runtime.Machine;
 import com.dylibso.chicory.wasm.Parser;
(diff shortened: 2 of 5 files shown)


andreaTP:
- CI: :sweat: seems like the workaround for `memoryCopy` is not correctly working on MacOs 11 (at least is flaky):
https://github.com/dylibso/chicory/actions/runs/14805229639/job/41572255174?pr=869
- Goals of the PR: wold you mind sharing your thought about: https://github.com/dylibso/chicory/issues/866#issuecomment-2847555711
- `Template`: I'm undecided here, I think the little change doesn't deserve the refactoring done, we are using JavaParser in various places in this repo and it will be great to keep a consistent style(e.g. full generation in plain code). Templating works good in this simple use case, but it doesn't always scale well.

chirino:
> * CI: 😓 seems like the workaround for `memoryCopy` is not correctly working on MacOs 11 (at least is flaky):
>   https://github.com/dylibso/chicory/actions/runs/14805229639/job/41572255174?pr=869

Maybe we need to warm up the metamorphic call site a little more.  I had used the smallest number of iterations that seemed to make JDK 17 work.

> * Goals of the PR: wold you mind sharing your thought about: [Aot: add `.builder()` method creates an Instance.Builder for you #866 (comment)](https://github.com/dylibso/chicory/issues/866#issuecomment-2847555711)

Ther advantages:
1. the `load()` is now smart: it caches the WasmModule in a static the first time `load()` is called.  Users DON'T have to do it themselves anymore.
2. using the `builder()` is single simpler.  About 3 times fewer method calls.
3. using the `builder()` is safer.  You don't  run the danger of using the wrong module with the wrong AotMachine.

> * `Template`: I'm undecided here, I think the little change doesn't deserve the refactoring done, we are using JavaParser in various places in this repo and it will be great to keep a consistent style(e.g. full generation in plain code). Templating works good in this simple use case, but it doesn't always scale well.

Well at least in this specific case, the code became much simpler to read and modify so I think it was a big win.  We should always be striving to keep things simple.

chirino:
Closing as not other support for the idea was expressed.

PR #873, opened 2025-05-05
Remove usages of Runtime.version

### LINKED ISSUE #870: Runtime.version() is not available on Android
In #844, the use of the `Runtime.version()` method was introduced. However, this method is not available on Android.  As a result, AOT engine does not work with Chicory 1.3.0 on Android.

When trying to run the AOT-compiled code, it crashes with the following error:
```
AndroidRuntime: FATAL EXCEPTION: Thread-3
AndroidRuntime: Process: com.example.wasm2class.android.java.app, PID: 26084
AndroidRuntime: java.lang.NoSuchMethodError: No static method version()Ljava/lang/Runtime$Version; in class Ljava/lang/Runtime; or its super classes (declaration of 'java.lang.Runtime' appears in /apex/com.android.art/javalib/core-oj.jar)
AndroidRuntime: 	at com.example.wasm2class.android.java.app.HelloworldMachine$AotMethods.<clinit>(AotMethods.java:76)
AndroidRuntime: 	at com.example.wasm2class.android.java.app.HelloworldMachineFuncGroup_0.func_3(Unknown Source:3)
AndroidRuntime: 	at com.example.wasm2class.android.java.app.HelloworldMachineFuncGroup_0.call_3(Unknown Source:0)
AndroidRuntime: 	at com.example.wasm2class.android.java.app.HelloworldMachine$MachineCall.call(Unknown Source:48)
AndroidRuntime: 	at com.example.wasm2class.android.java.app.HelloworldMachine.call(Unknown Source:6)
AndroidRuntime: 	at com.dylibso.chicory.runtime.Instance$Exports.lambda$function$0$com-dylibso-chicory-runtime-Instance$Exports(Instance.java:214)
AndroidRuntime: 	at com.dylibso.chicory.runtime.Instance$Exports$$ExternalSyntheticLambda0.apply(D8$$SyntheticClass:0)
AndroidRuntime: 	at com.example.wasm2class.android.java.app.MainActivity.executeWasiStart(MainActivity.java:66)
AndroidRuntime: 	at com.example.wasm2class.android.java.app.MainActivity.runHelloWorld(MainActivity.java:50)
AndroidRuntime: 	at com.example.wasm2class.android.java.app.MainActivity.testWasm2class(MainActivity.java:37)
AndroidRuntime: 	at com.example.wasm2class.android.java.app.MainActivity.$r8$lambda$SNY29VUloGRe4p-kc1nCrCblEAw(Unknown Source:0)
AndroidRuntime: 	at com.example.wasm2class.android.java.app.MainActivity$$ExternalSyntheticLambda3.run(D8$$SyntheticClass:0)
AndroidRuntime: 	at java.lang.Thread.run(Thread.java:1012)
```

This issue can be resolved by wrapping the `Runtime.version()` call in a try-catch block, like this:

```
boolean enableMemCopyWorkaround;
if (workaround != null) {
    enableMemCopyWorkaround = Boolean.parseBoolean(workaround);
} else {
    try {
        enableMemCopyWorkaround = Runtime.version().feature() < 21;
    } catch (NoSuchMethodError nsme) {
        // Runtime.version() is not available on Android
        enableMemCopyWorkaround = false;
    }
}
```

However, this call appears in two different places, each with its own JVM value and with different system properties for disabling the workaround:
1) https://github.com/andreaTP/chicory/blob/b9d7159fc2abd554d44fffe05fe6210e0226c1a9/aot/src/main/java/com/dylibso/chicory/experimental/aot/AotMethods.java#L76
2) https://github.com/andreaTP/chicory/blob/b9d7159fc2abd554d44fffe05fe6210e0226c1a9/runtime/src/main/java/com/dylibso/chicory/runtime/MemCopyWorkaround.java#L18

I'm not sure if this is intentional or not.

Alternatively, the version check can be rewritten by parsing the `java.version` system property. But then keep in mind that on Android the value of this property is 0: https://developer.android.com/reference/java/lang/System#getProperties().

andreaTP:
Fix #870 

@illarionov the CI didn't reported this issue as we are only running tests on the interpreter, another task to be done before declaring the aot stable is to include a CI job for it on android.

(cc. @chirino )

### DIFF
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/MemCopyWorkaround.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/MemCopyWorkaround.java
@@ -25,18 +25,22 @@ public final class MemCopyWorkaround {
     }
 
     static {
-        MemoryCopyFunc noop1 = (destination, offset, size, memory) -> {};
-        MemoryCopyFunc noop2 = (destination, offset, size, memory) -> {};
-        // Warm up the JIT... to make it see memoryCopyFunc.apply is megamorphic
-        for (int i = 0; i < 1000; i++) {
-            memoryCopyFunc = noop1;
-            MemCopyWorkaround.memoryCopy(0, 0, 0, null);
-            memoryCopyFunc = noop2;
-            MemCopyWorkaround.memoryCopy(0, 0, 0, null);
-        }
+        if (shouldUseMemWorkaround()) {
+            MemoryCopyFunc noop1 = (destination, offset, size, memory) -> {
+            };
+            MemoryCopyFunc noop2 = (destination, offset, size, memory) -> {
+            };
+            // Warm up the JIT... to make it see memoryCopyFunc.apply is megamorphic
+            for (int i = 0; i < 1000; i++) {
+                memoryCopyFunc = noop1;
+                MemCopyWorkaround.memoryCopy(0, 0, 0, null);
+                memoryCopyFunc = noop2;
+                MemCopyWorkaround.memoryCopy(0, 0, 0, null);
+            }
 
-        memoryCopyFunc =
-                (destination, offset, size, memory) -> memory.copy(destination, offset, size);
+            memoryCopyFunc =
+                    (destination, offset, size, memory) -> memory.copy(destination, offset, size);
+        }
     }
 
     static MemoryCopyFunc memoryCopyFunc;
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/MemCopyWorkaround.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/MemCopyWorkaround.java
@@ -8,6 +8,22 @@ public final class MemCopyWorkaround {
         void apply(int destination, int offset, int size, Memory memory);
     }
 
+    public static boolean shouldUseMemWorkaround() {
+        String version = System.getProperty("java.version");
+        if (version.equals("0")) {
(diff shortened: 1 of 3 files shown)


chirino (approved):
lgtm.  We could move all of the memCopyWorkaround init logic to a helper method in the runtime jar so that we don't need to generate the same code for each module.  It would also make the approval smaller.

andreaTP:
Good suggestions, incorporated, thanks :+1:

runtime/src/main/java/com/dylibso/chicory/runtime/MemCopyWorkaround.java
```diff
@@ -8,27 +8,33 @@ private interface MemoryCopyFunc {
...
-                memoryCopyFunc = noop1;
-                MemCopyWorkaround.memoryCopy(0, 0, 0, null);
-                memoryCopyFunc = noop2;
-                MemCopyWorkaround.memoryCopy(0, 0, 0, null);
-            }
+    static {
+        MemoryCopyFunc noop1 = (destination, offset, size, memory) -> {};
+        MemoryCopyFunc noop2 = (destination, offset, size, memory) -> {};
+        // Warm up the JIT... to make it see memoryCopyFunc.apply is megamorphic
+        for (int i = 0; i < 1000; i++) {
```
chirino, line 31: We should skip doing this loop if the workaround is not enabled.  No point in paying this class startup cost if it's not needed.

illarionov:
I tested this on an Android device, and it works, thanks!

PR #878, opened 2025-05-06
fix: only build and attach javadoc for modules which are intended to  only build and attach javadoc for module which are intended to be consumed as libraries

chirino:
Fixes #877

PR #882, opened 2025-05-07
fix: attach an empty javadoc for modules which are not intended to be consumed as libraries

### LINKED ISSUE #877: We should only build and attach javadoc for module which are intended to be consumed as libraries
Right now after a build if you run: `find . | grep javadoc.jar` you get:

```
./host-module/processor/target/host-module-processor-experimental-999-SNAPSHOT-javadoc.jar
./host-module/annotations/target/host-module-annotations-experimental-999-SNAPSHOT-javadoc.jar
./wasm-tools/target/wasm-tools-999-SNAPSHOT-javadoc.jar
./machine-tests/target/machine-tests-999-SNAPSHOT-javadoc.jar
./aot-build-time/target/aot-build-time-experimental-999-SNAPSHOT-javadoc.jar
./aot-maven-plugin/target/aot-maven-plugin-experimental-999-SNAPSHOT-javadoc.jar
./wasm-corpus/target/wasm-corpus-999-SNAPSHOT-javadoc.jar
./runtime-tests/target/runtime-tests-999-SNAPSHOT-javadoc.jar
./wasm/target/wasm-999-SNAPSHOT-javadoc.jar
./runtime/target/runtime-999-SNAPSHOT-javadoc.jar
./simd/target/simd-999-SNAPSHOT-javadoc.jar
./cli/target/cli-experimental-999-SNAPSHOT-javadoc.jar
./aot-build-time-cli/target/aot-build-time-cli-experimental-999-SNAPSHOT-javadoc.jar
./docs-lib/target/docs-lib-999-SNAPSHOT-javadoc.jar
./test-gen-plugin/target/test-gen-plugin-999-SNAPSHOT-javadoc.jar
./test-gen-lib/target/test-gen-lib-999-SNAPSHOT-javadoc.jar
./log/target/log-999-SNAPSHOT-javadoc.jar
./fuzz/target/fuzz-999-SNAPSHOT-javadoc.jar
./wasi/target/wasi-999-SNAPSHOT-javadoc.jar
./jmh/target/jmh-999-SNAPSHOT-javadoc.jar
./wasi-test-gen-plugin/target/wasi-test-gen-plugin-999-SNAPSHOT-javadoc.jar
./wabt/target/wabt-999-SNAPSHOT-javadoc.jar
./aot/target/aot-experimental-999-SNAPSHOT-javadoc.jar
```

Seems like too many,

chirino:
The empty javadoc.jar contains a README.md stating why no javadoc was included.

Fixes #877

### DIFF
--- a/cli/pom.xml
+++ b/cli/pom.xml
@@ -90,6 +90,7 @@
         <version>2.1.1</version>
         <configuration>
           <flags>-Xmx1G</flags>
+          <inputFile>${project.build.directory}/${project.build.finalName}.jar</inputFile>
           <programFile>chicory</programFile>
           <attachProgramFile>true</attachProgramFile>
         </configuration>
--- a/pom.xml
+++ b/pom.xml
@@ -89,6 +89,7 @@
     <maven-install-plugin.version>3.1.4</maven-install-plugin.version>
     <maven-jar-plugin.version>3.4.2</maven-jar-plugin.version>
     <maven-javadoc-plugin.version>3.11.2</maven-javadoc-plugin.version>
+    <maven-build-helper-maven-plugin>3.6.0</maven-build-helper-maven-plugin>
     <maven-plugin-plugin.version>3.15.1</maven-plugin-plugin.version>
     <maven-resources-plugin.version>3.3.1</maven-resources-plugin.version>
     <maven-shade-plugin.version>3.6.0</maven-shade-plugin.version>
@@ -722,28 +723,6 @@
 
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-javadoc-plugin</artifactId>
-        <version>${maven-javadoc-plugin.version}</version>
-        <configuration>
-          <doclint>none</doclint>
-          <source>11</source>
-        </configuration>
-        <executions>
-          <execution>
-            <id>attach-javadocs</id>
-            <goals>
-              <goal>jar</goal>
-            </goals>
-            <configuration>
-              <doclint>all,-missing</doclint>
-              <release>${maven.compiler.release}</release>
-            </configuration>
-          </execution>
-        </executions>
-      </plugin>
-
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-source-plugin</artifactId>
         <version>${maven-source-plugin.version}</version>
         <executions>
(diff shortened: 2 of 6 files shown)


andreaTP:
I'm unsure on the goal of this PR, FWIW the approach of #878 seems preferable when it solves a problem.

andreaTP (changes requested):
I can accept this change, but, please, scope it just to the modules that are [effectively published](https://github.com/dylibso/chicory/blob/4af3abf3abe9b2c28f37e6b7181e48e6012b2026/pom.xml#L860-L887).

pom.xml
```diff
@@ -763,6 +742,89 @@
...
+        </file>
+      </activation>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-javadoc-plugin</artifactId>
+            <version>${maven-javadoc-plugin.version}</version>
+            <configuration>
+              <!--<doclint>none</doclint>-->
```
andreaTP, line 807: leftover?

andreaTP (approved):
Applied the suggested changes, LGTM!

andreaTP (changes requested):
argh, sorry for the forth and back, I tested locally and seems like it's not working as expected.

andreaTP (changes requested):
I'm not sure why we're focusing on generating Javadocs for modules like `fuzz` or `docs-lib`, which are local-only and not published. These aren't intended for end users, so maintaining their Javadoc seems unnecessary.

On top of that, the proposed setup appears fragile and error-prone. To date, we haven't received any user complaints about Javadocs, and realistically, there's limited interest in the public API of something like a Maven plugin.

Given that, I suggest we drop this for now and revisit it only when a concrete user-facing issue comes up.

chirino:
> I'm not sure why we're focusing on generating Javadocs for modules like `fuzz` or `docs-lib`, which are local-only and not published. These aren't intended for end users, so maintaining their Javadoc seems unnecessary.

Ah I just realized we skip all of those in a release build.  I assumed all modules were getting published.  I'll back those parts.

> On top of that, the proposed setup appears fragile and error-prone. 

"appears" is subjective.

> To date, we haven't received any user complaints about Javadocs, and realistically, there's limited interest in the public API of something like a Maven plugin.

That's right.  That's why I'm proposing to not run javadoc against any of the classes we never intend to document for the end user.

> Given that, I suggest we drop this for now and revisit it only when a concrete user-facing issue comes up.

Javadoc is user facing since it gets published.

pom.xml
```diff
@@ -770,6 +749,90 @@
...
+            </configuration>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
+    <profile>
+      <id>package-standard-javadocs</id>
+      <activation>
+        <file>
+          <missing>${project.basedir}/src/main/javadoc/javadoc.jar</missing>
```
andreaTP: Is this referring to the old setup?

build-time-compiler/src/main/javadoc/README.md
```diff
@@ -0,0 +1,4 @@
+# Notice: Private API
```
andreaTP: `build-time-compiler` is one of the gray areas ... we want to expose a stable enough API for integrators (like the Gradle plugin), I don't want to scare ppl away with this message.

cli/pom.xml
```diff
@@ -90,6 +90,7 @@
         <version>2.1.1</version>
         <configuration>
           <flags>-Xmx1G</flags>
+          <inputFile>${project.build.directory}/${project.build.finalName}.jar</inputFile>
```
andreaTP, line 93: why this line is needed in the CLIs and not in the other modules?

andreaTP (changes requested):
> "appears" is subjective.

Right, this setup broke a couple of times in this PR without us noticing, it also needs manual testing and review.

cli/pom.xml
chirino: I think that's because our goal is attaching the -javadoc.jar before this one executes and then it's not sure which of the attached jar to used for execution.

build-time-compiler/src/main/javadoc/README.md
chirino: I'll remove this one then.

pom.xml
chirino: ah yes.. need to fix that.

PR #884, opened 2025-05-08
Nightly Zig testsuite

andreaTP:
Opening for visibility, if anyone is interested in picking up the challenge to fix the Symlink support of our Wasi layer :slightly_smiling_face: 

results:
https://github.com/andreaTP/chicory/actions/runs/14913567382/job/41893594463#step:5:6908

> 2677 passed; 118 skipped; 8 failed.

all of the failing tests are related to links/symlinks on the FileSystem.

The patch is used to fix the printing to standard out and skip a few additional tests.


andreaTP:
This builds on top of #891 but this is the result: https://github.com/andreaTP/chicory/actions/runs/15273720201/job/42955016140#step:5:6887

We need to exclude a few tests that are exercising the wasi link/symlink support, but it should be fine.

evacchi (approved):
not a blocker, but can you describe what the patch is doing in the top post (or somewhere, a comment in build-zig-testsuite.sh is fine too)

chirino (approved):
lgtm

PR #885, opened 2025-05-09
[doc]: Update wasi.md, use the created options variable

sgandon:
the java example is creating a `options` variable but it is not used in the next line.


andreaTP:
The CI is fixed in #887

PR #886, opened 2025-05-09
Update select to handle V128

andreaTP:
Opening in draft waiting for https://github.com/WebAssembly/spec/pull/1901 to be resolved.


andreaTP:
This is one way of implementing it ...
I'm looking for feedback, but seems unavoidable that we make either `Instruction` or `FunctionBody` mutable to collect information from the validation phase.

An alternative is to add a separate phase like we already do in the `AotAnalyzer`.

andreaTP:
Moving back to Draft as I want to re-iterate on this.

andreaTP:
Ready for review!

PR #889, opened 2025-05-13
Remove forRemoval from fresh deprecations

andreaTP:
Both the IDE and the command line are extremely aggressive in notifying `Deprecation`s marked for removal, since we have not released yet those changes I take the chance to step back on this detail.


electrum (approved):

evacchi (approved):
approved. IntelliJ puts squiggly red lines everywhere, making it look like a compilation error (and it's not). the `@deprecated` JavaDoc should be good enough for extra context.

PR #891, opened 2025-05-15
[WASI] Support for path_link and path_symlink

andreaTP:
This should reduce the gap in #884


evacchi (approved):

andreaTP:
@electrum if you have 5 minutes to look at this I'd love another couple of eyes :pray:

PR #893, opened 2025-05-16
Follow up to #876 rename the AotMethods reference, deprecate AotMachine.

chirino:


aot/src/main/java/com/dylibso/chicory/experimental/aot/AotMachine.java
```diff
@@ -8,11 +8,23 @@
 /**
  * Machine implementation that compiles WASM function bodies to JVM byte code.
  * All compilation is done in a single compile phase during instantiation.
+ * <p>
+ * This class is deprecated and will be removed in a future version. Please use
+ * the {@link com.dylibso.chicory.compiler.MachineFactoryCompiler} instead.
  */
+@Deprecated(since = "1.4.0", forRemoval = true)
```
andreaTP, line 15: I did the same, but, turns out that `forRemoval = true` is getting marked as an error by the IDEs, causing major confusion.

andreaTP (changes requested):
I like the change, but there is still a problem with approvals :sweat_smile: :
Screenshot from 2025-05-19 09-48-12

andreaTP:
@chirino you have [this PR](https://github.com/chirino/chicory/pull/8) to fix the approvals :+1:

andreaTP (approved):
LGTM, thanks for the follow up!

PR #900, opened 2025-05-20
Avoid MH in generated code

andreaTP:
I'm not sure it will cause and compatibility issue, since it is directly generated by the compiler,.

But with this change we remove the MH Lookup.

Noticed in #893 cc. @chirino

### DIFF
--- a/aot/src/main/java/com/dylibso/chicory/compiler/internal/Shaded.java
+++ b/aot/src/main/java/com/dylibso/chicory/compiler/internal/Shaded.java
@@ -188,7 +188,7 @@ public final class Shaded {
     }
 
     public static RuntimeException throwUnknownFunction(int index) {
-        throw new InvalidException("unknown function " + index);
+        throw new InvalidException(String.format("unknown function %d", index));
     }
 
     public static void checkInterruption() {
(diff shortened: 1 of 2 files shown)


chirino (approved):
lgtm

electrum:
What’s the purpose of this change? Many Java features are compiled to method handles. Do you want to remove all lambdas and string concatenation from Chicory?

aot/src/main/java/com/dylibso/chicory/compiler/internal/Shaded.java
```diff
@@ -188,7 +188,7 @@ public static RuntimeException throwTrapException() {
     }
 
     public static RuntimeException throwUnknownFunction(int index) {
-        throw new InvalidException("unknown function " + index);
+        throw new InvalidException(String.format("unknown function %d", index));
```
electrum, line 191: This can use `String.concat()`

andreaTP:
> Do you want to remove all lambdas and string concatenation from Chicory?

No no :slightly_smiling_face: 

> Many Java features are compiled to method handles.

This is correct, it just got my eye in the approvals ...

PR #904, opened 2025-05-22
Switch to ZeroFs and remove Guava

andreaTP:
Just finished releasing the very very first version of [ZeroFs](https://github.com/roastedroot/zerofs) :slightly_smiling_face: .

I'd love to have feedback and real world testing of ZeroFs :pray: let me know how it goes!

Hopefully we can say :wave: bye bye to Guava :slightly_smiling_face: 

We also need to update the docs, but I'd leave a little time to do so.

### DIFF
--- a/machine-tests/pom.xml
+++ b/machine-tests/pom.xml
@@ -37,8 +37,8 @@
       <scope>test</scope>
     </dependency>
     <dependency>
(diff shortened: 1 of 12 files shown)


evacchi (approved):
🔥

PR #907, opened 2025-05-27
fix: handle imported exceptions when tag section is missing

### LINKED ISSUE #906: Can't parse wasm module that catches imported exception but which does not have a tag section.
I get a:
```
java.util.NoSuchElementException: No value present

	at java.base/java.util.Optional.get(Optional.java:143)
	at com.dylibso.chicory.wasm.Validator.getTagType(Validator.java:331)
	at com.dylibso.chicory.wasm.Validator.validateFunction(Validator.java:587)
	at com.dylibso.chicory.wasm.Validator.validateFunctions(Validator.java:543)
	at com.dylibso.chicory.wasm.WasmModule$Builder.build(WasmModule.java:274)
	at com.dylibso.chicory.wasm.Parser.parse(Parser.java:240)
	at com.dylibso.chicory.wasm.Parser.parse(Parser.java:204)
...
```

When parsing the 2nd module defined in the following wast:

```wast
;; exports a tag and a function that throws it.
(module
  (tag $e0 (export "e0"))
  (func (export "throw") (throw $e0))
)

(register "test")

;; catches an imported tag.
(module

  (tag $imported-e0 (import "test" "e0"))
  (func $imported-throw (import "test" "throw"))

  (func (export "catch-imported") (result i32)
    (block $h
      (try_table (result i32) (catch $imported-e0 $h)
        (call $imported-throw (i32.const 1))
      )
      (return)
    )
    (i32.const 2)
  )
)
```

andreaTP:
Fix #906 

2 commits as we need to move `wasm-corpus` to `wasm-tools` to deal with EH


chirino (approved):
lgtm.

PR #908, opened 2025-05-27
[compiler] respect original build time configured class name

### LINKED ISSUE #865: AoT: build time configure class name the class that contains the `load` and `create` methods.
Seems weird you configure 

`org.example.Foo`

But have to use 

`org.example.FooModule.load()`

andreaTP:
Fix #865 

Since we plan to release the compiler as stable we can do a breaking change here, it should be reflected in the release notes, something along those lines:

```
The build time compiler 'name' configuration doesn't append anymore the "Module" suffix to the generated class name.
```

cc. @chirino


chirino (approved):
lgtm.

PR #918, opened 2025-05-29
compiler: implement WebAssembly exception handling with try_table support

chirino:
This gets most of the exception spec tests passing under the wasm to bytecode compiler.


wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java
```diff
@@ -1024,6 +1024,7 @@ private static CodeSection parseCodeSection(ByteBuffer buffer, TypeSection typeS
                             // block start
                             currentControlFlow =
                                     currentControlFlow.spawn(instructions.size(), instruction);
+                            currentControlFlow.addCallback(instruction::withLabelTrue);
```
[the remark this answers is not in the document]
chirino: We need to track where the try block ends.  I'm using the true table to hold that info.  I could have added another field to the instruction, but seems better to reuse that field to avoid increasing memory usage by adding a field.

wasm-tools/pom.xml
```diff
@@ -91,6 +91,21 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-dependency-plugin</artifactId>
```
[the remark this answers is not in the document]
chirino: So I started seem errors from the dependency analyzer.  Everyone in a while I get that.  Let try to back that part out and see if I still see the same proble.

wasm-corpus/src/main/resources/wast/exceptions.wast
```diff
@@ -0,0 +1,14 @@
+(module
```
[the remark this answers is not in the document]
chirino: At the time, it was the only way I figured out to compile stuff that coudl use try_table.

aot/src/main/java/com/dylibso/chicory/compiler/internal/Compiler.java
```diff
@@ -1257,6 +1293,133 @@ private void compileFunction(
...
+
+                case TRY_TABLE:
+                    {
+                        TryTableInfo info = tryTableInfos.get(ins);
+                        assert info != null;
+                        tryTableStack.push(info);
+                        asm.mark(info.tryStart);
+                        break;
+                    }
+                case TRY_END:
```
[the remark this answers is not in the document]
chirino: I don't see the scope guy in the Compiler.  just the WasmAanalyzer.

aot/src/main/java/com/dylibso/chicory/compiler/internal/Compiler.java
```diff
@@ -1212,17 +1226,39 @@ private void compileFunction(
             asm.store(ctx.localSlotIndex(i), asmType(localType));
         }
 
+        // Pre-process to find try_table blocks and their labels
+        java.util.Deque<TryTableInfo> tryTableStack = new java.util.ArrayDeque<>();
```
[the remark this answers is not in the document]
chirino: Yes.  are you saying encode that into the operands field too?

wasm-corpus/src/main/resources/wast/exceptions.wast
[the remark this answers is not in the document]
andreaTP: rebase on top of #907 (already in `main`), should work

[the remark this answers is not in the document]
chirino: Might be good to keep 1 example of how to compile a wast tho.  Testing against a subset of spec wasts was really helpful when implementing this PR.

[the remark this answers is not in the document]
andreaTP: Not really fussy about this, but no one other than us is going to use `wast` files...

Typically, when starting a new development we tweak the [Maven plugin configuration](https://github.com/dylibso/chicory/blob/42e83ee53c166a121cc8749b9d46302d013560a2/aot-tests/pom.xml#L81) to include ONLY the new wast files and disabling the rest (to have a faster feedback cycle).

chirino:
I guess I need to think about how to move some of the work I'm doing in the compiler to the WasmAanalyzer.  Maybe add the TRY_TABLE CompilerInstruction when the try block ends instead of trying to do it when it starts.

wasm-corpus/src/main/resources/wast/exceptions.wast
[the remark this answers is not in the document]
chirino: Updated PR to just use a wat.  I had to add the wasm-tools to the docker image.

aot/src/main/java/com/dylibso/chicory/compiler/internal/Compiler.java
```diff
@@ -1257,6 +1293,133 @@ private void compileFunction(
                         asm.ifne(labels.get(ins.operand(0)));
                     }
                     break;
+
+                case TRY_TABLE:
+                    {
+                        TryTableInfo info = tryTableInfos.get(ins);
+                        assert info != null;
+                        tryTableStack.push(info);
+                        asm.mark(info.tryStart);
```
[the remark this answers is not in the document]
chirino: All the labels are now added by the analyzer.

aot/src/main/java/com/dylibso/chicory/compiler/internal/Compiler.java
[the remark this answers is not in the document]
chirino: yep..that's what I ended up doing.

compiler/src/main/java/com/dylibso/chicory/compiler/internal/WasmAnalyzer.java
```diff
@@ -110,6 +124,43 @@ public List<CompilerInstruction> analyze(int funcId) {
                     stack.enterScope(ins.scope(), blockType(ins));
                     break;
                 case END:
+                    // Check if this is the end of a TRY_TABLE block
+                    if (ins.scope().opcode() == OpCode.TRY_TABLE) {
+                        Instruction tryTableIns = ins.scope();
+                        var operands = tryTableOperands.remove(tryTableIns.address());
+                        // Werid: sometimes we see END occur multiple times for the same try table.
```
chirino, line 131: @andreaTP do you know why this happens?

compiler-tests/pom.xml
```diff
@@ -174,7 +182,23 @@
...
+            SpecV1TcReturnCallTest.test31,
+            SpecV1TcReturnCallIndirectTest.test40,
+            SpecV1TcReturnCallIndirectTest.test41,
+            SpecV1TcReturnCallIndirectTest.test46,
+            SpecV1TcReturnCallIndirectTest.test47,
+            <!-- Skip exception tests with tail calls -->
+            SpecV1EhTryTableTest.test39,
+            SpecV1EhTryTableTest.test40,
+            <!-- Investigate why this one fails -->
+            SpecV1EhBinaryTest.test40
```
chirino, line 200: @andreaTP I need an expert at reading binary WASM to explain to me what this test is doing..
So I assume this should pass the Parser, and it's up to the compiler to error it out.  Just not sure what I need to look out for.

```
;; Init expression with missing end marker
(assert_malformed
  (module binary
    "\00asm" "\01\00\00\00"
    "\01\04\01\60\00\00"       ;; Type section: 1 type
    "\03\02\01\00"             ;; Function section: 1 function
    "\06\05\01\7f\00\41\00"    ;; Global section: 1 entry with missing end marker
    ;; Missing end marker here
    "\0a\04\01\02\00\0b"       ;; Code section: 1 function
  )
  "unexpected end of section or function"
)
```

compiler/src/main/java/com/dylibso/chicory/compiler/internal/WasmAnalyzer.java
```diff
@@ -110,6 +124,43 @@ public List<CompilerInstruction> analyze(int funcId) {
...
+                        if (operands != null) {
+
+                            // add a NOOP just in case the it's an empty try block.
+                            result.add(new CompilerInstruction(CompilerOpCode.NOP));
```
chirino, line 124: is there a good way to know if we are dealing with an empty try block?

compiler/src/test/java/com/dylibso/chicory/approvals/ApprovalTest.java
```diff
@@ -69,7 +70,7 @@ public void verifyI32() {
     public void verifyI32Renamed() {
         var module = parse(getSystemClassLoader().getResourceAsStream("compiled/i32.wat.wasm"));
         var result = Compiler.builder(module).withClassName("FOO").build().compile();
-        verifyClass(result.classBytes(), false);
+        verifyClass(result.classBytes(), (name) -> false);
     }
```
chirino, line 74: We have finer grained control of which classes are included in the approval now.  what is it that `verifyI32Renamed` is focusing on?  Is it just the Shaded class?

compiler-tests/pom.xml
```diff
@@ -174,7 +182,23 @@
           </includedWasts>
           <!-- Skip Binary as there is a THROW_REF in a Global initializer -->
           <!-- Skip stack-overflows in tail calls -->
```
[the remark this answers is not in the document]
andreaTP: sadly, spotless is reordering the comments in XML :cry:

compiler-tests/pom.xml
andreaTP: IIRC the Global initializer contains > 1 instructions.

I left a comment here:
https://github.com/dylibso/chicory/blob/df5be87550b2d2347f0fc2e4ca5cb4c06a9751bc/runtime-tests/pom.xml#L182-L183

This test is supposed to fail differently depending on the enabled proposals, I think we can hold off a bit on being this strict, i.e. as soon as the module fails it's good enough.

I agree on skipping this one.

compiler/src/main/java/com/dylibso/chicory/compiler/internal/Compiler.java
```diff
@@ -1116,7 +1120,7 @@ private static void compileHostFunction(int funcId, FunctionType type, Instructi
         emitUnboxResult(type, asm);
     }
 
-    private static void emitBoxArguments(InstructionAdapter asm, List<ValType> types) {
+    static void emitBoxArguments(InstructionAdapter asm, List<ValType> types) {
```
andreaTP: leftover? doesn't seem to be useful to this PR.

compiler/src/main/java/com/dylibso/chicory/compiler/internal/Shaded.java
```diff
@@ -204,4 +205,28 @@ public static long readGlobal(int index, Instance instance) {
...
+
+    public static boolean exceptionMatches(WasmException exception, int tag, Instance instance) {
+        if (exception.instance() == instance && exception.tagIdx() == tag) {
+            return true;
+        }
+
+        var currentCatchTag = instance.tag(tag);
+        var exceptionTag = exception.instance().tag(exception.tagIdx());
+        return tag < instance.imports().tagCount()
+                && currentCatchTag.type().paramsMatch(exceptionTag.type())
```
andreaTP: nit: `currentCatchTag.type().typesMatch(exceptionTag.type())`

compiler/src/main/java/com/dylibso/chicory/compiler/internal/WasmAnalyzer.java
andreaTP: Shouldn't, but I'd need to check the flow, is it happening on the "happy path" e.g. no exceptions thrown?

compiler/src/main/java/com/dylibso/chicory/compiler/internal/WasmAnalyzer.java
andreaTP: If a `TRY` block is empty can we skip emitting it at all?
given the structured control flow we are sure that no exception can be thrown inside the body.

compiler/src/test/java/com/dylibso/chicory/approvals/ApprovalTest.java
andreaTP: it's focusing on emitting on the functionality of `withClassName`, any better/alternative proposal for testing this bit is welcome, there is space for improvement here.

wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java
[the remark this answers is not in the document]
chirino: didn't end up needing it.  backed that out.

wasm-corpus/Dockerfile
```diff
@@ -59,6 +59,9 @@ RUN curl -L -O https://github.com/tinygo-org/tinygo/releases/download/v0.30.0/ti
 
 ENV PATH="/opt/wabt/bin:${PATH}"
 
+# Install wasm-tools
+RUN cargo install wasm-tools
```
andreaTP, line 63: Is this a bug fix?

compiler/src/main/java/com/dylibso/chicory/compiler/internal/WasmAnalyzer.java
```diff
@@ -225,6 +265,58 @@ public List<CompilerInstruction> analyze(int funcId) {
                     result.add(new CompilerInstruction(CompilerOpCode.SWITCH, operands));
                     result.addAll(unwinds);
                     break;
+                case TRY_TABLE:
+                    {
+                        stack.enterScope(ins.scope(), blockType(ins));
```
andreaTP, line 296: Those blocks are doing ways too much, and it's not your fault @chirino you have just been working around the current design.

Let me follow up with something that might enable cleaning this mess around.

compiler/src/main/java/com/dylibso/chicory/compiler/internal/WasmAnalyzer.java
chirino: right.  I need need a good way to tell if we are in the empty try block case.

compiler/src/test/java/com/dylibso/chicory/approvals/ApprovalTest.java
chirino: Ah.. i got a an idea.

compiler/src/main/java/com/dylibso/chicory/compiler/internal/Compiler.java
chirino: yep.

wasm-corpus/Dockerfile
chirino: yeah.

andreaTP:
Forgot to answer:

> does the interpreter throw/catch WasmExceptions across method boundaries? 

Yes, ideally we should have tests covering the interactions interpreter -> compiled and vice versa.

compiler/src/main/java/com/dylibso/chicory/compiler/internal/WasmAnalyzer.java
```diff
@@ -110,6 +113,43 @@ public List<CompilerInstruction> analyze(int funcId) {
...
+                            // add a NOOP just in case the it's an empty try block.
+                            result.add(new CompilerInstruction(CompilerOpCode.NOP));
+
+                            var tryEndLabel = nextLabel++;
+                            result.add(new CompilerInstruction(CompilerOpCode.LABEL, tryEndLabel));
+                            operands[1] = tryEndLabel;
+
+                            // Jump over the exception handler during normal execution
+                            var afterHandlerLabel = nextLabel++;
+                            result.add(
```
andreaTP: are we missing the unwinds?
Likely they are not exhaustively tested by the testsuite, but this should behave like a `BR` like instruction.

compiler/src/main/java/com/dylibso/chicory/compiler/internal/WasmAnalyzer.java
```diff
@@ -225,6 +265,58 @@ public List<CompilerInstruction> analyze(int funcId) {
...
+                case TRY_TABLE:
+                    {
+                        stack.enterScope(ins.scope(), blockType(ins));
+
+                        // create the operands of the TRY_TABLE....
+                        ArrayList<Long> ops = new ArrayList<>();
+
+                        // Emit a LABEL instruction to mark the try start
+                        int tryStartLabel = nextLabel++;
+                        result.add(new CompilerInstruction(CompilerOpCode.LABEL, tryStartLabel));
```
andreaTP: Instead of emitting just a `LABEL` here we should already unroll the catches and generate emit the relevant `CompilerInstruction`

compiler/src/main/java/com/dylibso/chicory/compiler/internal/Compiler.java
```diff
@@ -1257,6 +1271,116 @@ private void compileFunction(
...
+                                                .skip(handlerOffset + handlerCount)
+                                                .toArray());
+
+                        // This instruction only executes when an exception is caught
+                        // (not during normal execution flow)
+                        var exceptionSlot = ctx.tempSlot();
+                        asm.store(exceptionSlot, OBJECT_TYPE);
+
+                        if (!catches.isEmpty()) {
+                            for (int catchIdx = 0; catchIdx < catches.size(); catchIdx++) {
```
andreaTP: We should be able to move most of this logic in the analyzer.

The CompilerOpCodes doesn't need to be 1:1 with the Wasm OpCodes, we can add more when necessary.

compiler/src/main/java/com/dylibso/chicory/compiler/internal/Context.java
```diff
@@ -134,4 +138,28 @@ public int tempSlot() {
     public String classNameForFuncGroup(int funcId) {
         return "FuncGroup_" + (funcId / maxFunctionsPerClass);
     }
+
+    public FunctionType getTagFunctionType(int tagId) {
+        var tagSection = module.tagSection();
+        var tagImports =
+                module.importSection().stream()
```
andreaTP: `getTagFunctionType` is an expensive function, currently it's invoked during compilation, but we can isolate the execution to the analyze phase (and possibly memoize `tagImports` compute it it only once at the start of the method?

chirino: Yep thought the same, but I figured we can do a optimization pass once we get done with the correctness.

compiler/src/main/java/com/dylibso/chicory/compiler/internal/WasmAnalyzer.java
chirino: Are you saying emit most of what's being done in the END: case at this point?

chirino:
> * unroll the catches in the analyzer, one option is to emit one dedicated CompilerOpCode for CATCH/CATCH_REF/etc.
> * the `CompilerOpCode.TRY_TABLE` operands are going to be ONLY the 3 labels

Ah I see.  make sense.. should make the CompilerInstruction operand encodings a little simpler.

chirino:
> Next: need to think about how this interacts with interpreted methods. @andreaTP does the interpreter throw/catch WasmExceptions across method boundaries? Do you think I need to generate a big method for a test case that that throws and or catches exceptions?

Added test for this.  Work one way but fails in the other direction: `ExceptionTest.throwFromCompiledCatchInterpreted`

compiler/src/main/java/com/dylibso/chicory/compiler/internal/CompilerInstruction.java
```diff
@@ -1,31 +1,50 @@
...
 import java.util.Arrays;
+import java.util.Map;
+import java.util.Objects;
 import java.util.stream.LongStream;
+import org.objectweb.asm.Label;
+import org.objectweb.asm.commons.InstructionAdapter;
 
 final class CompilerInstruction {
+
+    interface Emitter {
```
andreaTP: This is an escape hatch to emit ASM directly from the analyzer, I understand the motivation of convenience but the resulting code is far less readable to me:

- the boundary of analyze wasm/emit bytecode is broken: we declare asm everywhere and it will eventually mix and match
- the code for emitting wasm is in capturing lambdas instead of plain methods depending only on the input opcode + operands

compiler/src/main/java/com/dylibso/chicory/compiler/internal/WasmAnalyzer.java
```diff
@@ -40,6 +54,18 @@ final class WasmAnalyzer {
     private final List<ValType> tableTypes;
     private final int functionImports;
 
+    private static class TryCatchBlock {
```
andreaTP: I find the previous encoding as a CompilerOpCode with 3 operands for the labels better suited

compiler/src/main/java/com/dylibso/chicory/compiler/internal/WasmAnalyzer.java
```diff
@@ -67,7 +93,13 @@ public List<CompilerInstruction> analyze(int funcId) {
         // find label targets
         Set<Integer> labels = new HashSet<>();
 
-        for (AnnotatedInstruction ins : body.instructions()) {
+        HashMap<Integer, TryCatchBlock> tryCatchBlocks = new HashMap<>();
+        ArrayDeque<TryCatchBlock> tryCatchBlocksStack = new ArrayDeque<TryCatchBlock>();
+        HashMap<Integer, AnnotatedInstruction> tryTables = new HashMap<>();
```
andreaTP: This is a little scary, do we really need 3 stateful data structure?
From the names is hard to guess what is used for what.

compiler/src/main/java/com/dylibso/chicory/compiler/internal/WasmAnalyzer.java
```diff
@@ -281,32 +301,209 @@ public List<CompilerInstruction> analyze(int funcId) {
...
+                            // Emit the exception handler logic directly
+                            long[] labelTargets =
+                                    tryTableIns.catches().stream()
+                                            .mapToLong(CatchOpCode.Catch::resolvedLabel)
+                                            .toArray();
+
+                            result.add(
+                                    new CompilerInstruction(
+                                            labelTargets,
+                                            (ctx, asm, asmLabels) -> {
```
andreaTP: this is where things goes funky, dropping all the logic in a lambda is making it very hard to understand, I feel like emitting separate `CompilerOpCodes` for the different `CATCH` clause is going to make this much easier to follow.

andreaTP (changes requested):
Thanks a lot for the iteration @chirino , moving stuff to the Analyzer is, surely, the path ahead.
That said there are still a few things we can improve on:

- don't break the analyzer/compiler boundary abstraction
- verify the stateful datastructures (Maps, Dequeues, etc.) are they all needed?
- (possibly) split the emission of CATCH blocks by type

chirino:
> Thanks a lot for the iteration @chirino , moving stuff to the Analyzer is, surely, the path ahead. That said there are still a few things we can improve on:
> 
> * don't break the analyzer/compiler boundary abstraction

This new version more asm emitter logic into the Emitter class.  

> * verify the stateful datastructures (Maps, Dequeues, etc.) are they all needed?

Got it down to one.

> * (possibly) split the emission of CATCH blocks by type

I've moved the emission of each CATCH block into the Emitter.  We can probably further decompose if we think it's going to be more maintainable.

compiler-tests/pom.xml
```diff
@@ -174,7 +182,23 @@
...
+            SpecV1TcReturnCallTest.test24,
+            SpecV1TcReturnCallTest.test25,
+            SpecV1TcReturnCallTest.test30,
+            SpecV1TcReturnCallTest.test31,
+            SpecV1TcReturnCallIndirectTest.test40,
+            SpecV1TcReturnCallIndirectTest.test41,
+            SpecV1TcReturnCallIndirectTest.test46,
+            SpecV1TcReturnCallIndirectTest.test47,
+            <!-- Skip exception tests with tail calls -->
+            SpecV1EhTryTableTest.test39,
```
andreaTP, line 197: What's the error?
IIRC those are not really testing tail call so they should pass too.

andreaTP (changes requested):
Thanks a lot for the great work @chirino !

I think this PR is getting closer to be ready, missing steps:
- split the handling of the CATCH blocks - non blocking
- optimize `getTagFunctionType` - non blocking
- remove the usage of the `Emitter` and use separate `CompilerInstructions` instead of capturing lambdas - blocking

There are good reasons to avoid introducing a new abstraction sneaking it into an unrelated feature.
We should discuss the change separately, and keep the codebase consistent meanwhile.

evacchi (commented):
hey this is all great work!

I have mostly one concern, I am not completely sold on this "emitter" change, and I think it's better to discuss this change separately.

As for the current setup, even though it might seem more work, the forced separation between phases ("analysis" and "emit") will pay off in the future.

- First of all, the separation **simplifies the logic** in the emission phase; 
- it **scales better** in case we wanted to **further process the IR**, e.g. insert other phases in between; for instance, the mapping between Wasm opcodes and Compiler opcodes is already _not_ 1:1. This often means emitting N Compiler opcodes for 1 Wasm opcode, but the reverse might be true; it might also mean post-processing the Compiler opcodes and lowering them further (e.g. if we wanted to introduce some sort of optimization step)
- it forces to **pass simple values between phases** (this can be a downside sometimes, but we can always rethink the `CompilerInstruction` class to add more fields), and that has also a benefit to track the memory consumption of the compiler (we haven't done this so far, but it's something to keep in mind)
- it allows to **add more backends** instead of tying to a specific one: this is partially how the Android backend works (*)

While I can see why we might want to "slice" the components differently (e.g. by opcode instead of by phase), this feels like a pretty big change, and I think it begs for more thought

PS: I wonder what @electrum's opinion would be :)

(*) it's unfortunately still a mess for a lot of reasons, including the fact that the Android VM uses registers instead of a stack 🙈

chirino:
> hey this is all great work!
> 

Thanks!

> I have mostly one concern, I am not completely sold on this "emitter" change, and I think it's better to discuss this change separately.
> 
> As for the current setup, even though it might seem more work, the forced separation between phases ("analysis" and "emit") will pay off in the future.
> 
> * First of all, the separation **simplifies the logic** in the emission phase;
> * it **scales better** in case we wanted to **further process the IR**, e.g. insert other phases in between; for instance, the mapping between Wasm opcodes and Compiler opcodes is already _not_ 1:1. This often means emitting N Compiler opcodes for 1 Wasm opcode, but the reverse might be true; it might also mean post-processing the Compiler opcodes and lowering them further (e.g. if we wanted to introduce some sort of optimization step)
> * it forces to **pass simple values between phases** (this can be a downside sometimes, but we can always rethink the `CompilerInstruction` class to add more fields), and that has also a benefit to track the memory consumption of the compiler (we haven't done this so far, but it's something to keep in mind)
> * it allows to **add more backends** instead of tying to a specific one: this is partially how the Android backend works (*)
> 

The only downside is that the IR output from the Analyzer is not documented anywhere. This is in contrast to the WASM and ASM instructions which are well-documented.  Considering the code base doesn’t currently seem to require the features you’ve outlined, expanding it seemed like the wrong direction.  But maybe that's because I've not seen this Android backend.  Where can I find it?

andreaTP:
> The only downside is that the IR output from the Analyzer is not documented anywhere.

This is expected, in my experience, real-world compilers rarely provide exhaustive documentation for their intermediate representations.

> This is in contrast to the WASM and ASM instructions which are well-documented. 

That's accurate, the input and the output are formally defined.

> Considering the code base doesn’t currently seem to require the features you’ve outlined, expanding it seemed like the wrong direction.

Simpler logic, separation of concerns, and the other principles Edoardo outlined are all intentional goals of this codebase. They may not be perfectly realized yet, but they are **deliberately and intentionally pursued**, as stated in the [CONTRIBUTING guidelines](https://github.com/dylibso/chicory/blob/main/CONTRIBUTING.md#coding-philosophy).

All in all, I'm grateful for the current implementation, as it gives a solid foundation at finalizing the feature, but I see two possible outcomes:
- @chirino you bare with us a little longer, we iterate and decouple structural changes from the addition of a feature
- I'll take over and finalize this PR (somewhere next week possibly)

andreaTP:
Sorry for the delay, I took the time to do the refactoring and last review myself, you can find the result here:
https://github.com/chirino/chicory/pull/10

and this is the final diff:
https://github.com/dylibso/chicory/compare/main...andreaTP:compiler-exceptions-review

All in all, it was great work and I mostly moved around things so that we respect the original structure of the compiler.

@chirino if you accept the PR it's ready for merge to me, thanks again for spearheading the effort!

Please @evacchi if you can have a second look it would be appreciated :pray:

compiler/src/main/java/com/dylibso/chicory/compiler/internal/Emitters.java
```diff
@@ -658,4 +668,78 @@ public static BytecodeEmitter intrinsify(CompilerOpCode opcode, Class<?> staticH
...
+            asm.invokevirtual(
+                    getInternalName(WasmException.class),
+                    "args",
+                    getMethodDescriptor(getType(long[].class)),
+                    false);
+
+            // Store the array in a local variable
+            // Unbox each argument from the
+            // long[] array and push onto stack
+            emitUnboxResult(asm, tagFuncType.params(), ctx.tempSlot() + 1);
```
evacchi, line 721: not sure 100% why/how ctx.tempSlot() **+1** works? why can't we reuse tempSlot here? 🤔

compiler/src/main/java/com/dylibso/chicory/compiler/internal/EmitterMap.java
```diff
@@ -19,6 +19,15 @@ final class EmitterMap {
                     .intrinsic(CompilerOpCode.CALL, Emitters::CALL)
                     .intrinsic(CompilerOpCode.CALL_INDIRECT, Emitters::CALL_INDIRECT)
 
+                    // ====== Exception Handling ======
+                    .intrinsic(CompilerOpCode.THROW, Emitters::THROW)
+                    .intrinsic(CompilerOpCode.THROW_REF, Emitters::THROW_REF)
+                    .intrinsic(CompilerOpCode.CATCH_COMPARE_TAG, Emitters::CATCH)
```
evacchi: better to align these names for consistency

compiler/src/main/java/com/dylibso/chicory/compiler/internal/WasmAnalyzer.java
```diff
@@ -264,6 +375,66 @@ public List<CompilerInstruction> analyze(int funcId) {
...
+
+            switch (catchCondition.opcode()) {
+                case CATCH:
+                    result.add(
+                            new CompilerInstruction(
+                                    CompilerOpCode.CATCH_COMPARE_TAG, catchCondition.tag()));
+                    result.add(new CompilerInstruction(CompilerOpCode.IFEQ, afterCatchLabel));
+                    result.add(
+                            new CompilerInstruction(
+                                    CompilerOpCode.CATCH_UNBOX_PARAMS, catchCondition.tag()));
```
evacchi, line 405: a slight nit, but it looks like the sequence is always:

- CATCH_COMPARE_TAG tag
- IFEQ label
- CATCH_UNBOX_PARAMS tag

so that makes me wonder if it is necessary for them to be separate opcodes -- but it's almost a matter of taste

compiler/src/main/java/com/dylibso/chicory/compiler/internal/EmitterMap.java
andreaTP: do you have a suggestion?

compiler/src/main/java/com/dylibso/chicory/compiler/internal/WasmAnalyzer.java
andreaTP: I did the split to keep the `IFEQ label` separate since it needs to be processed with the labels.
The other two op codes becomes straight forward bytecode emission.

compiler/src/main/java/com/dylibso/chicory/compiler/internal/Emitters.java
andreaTP: because the `tempSlot` is already used to stash the exception, so we use the next (e.g. `+1`) slot to store the array.

compiler/src/main/java/com/dylibso/chicory/compiler/internal/EmitterMap.java
evacchi:
```suggestion
                    .intrinsic(CompilerOpCode.CATCH_COMPARE_TAG, Emitters::CATCH_COMPARE_TAG)
```

compiler/src/main/java/com/dylibso/chicory/compiler/internal/EmitterMap.java
```diff
@@ -19,6 +19,15 @@ final class EmitterMap {
...
+                    .intrinsic(CompilerOpCode.CATCH_UNBOX_PARAMS, Emitters::CATCH_UNBOX_PARAMS)
+                    .intrinsic(CompilerOpCode.CATCH_REGISTER_EXCEPTION, Emitters::CATCH_REF)
```
evacchi:
```suggestion
                    .intrinsic(CompilerOpCode.CATCH_REGISTER_EXCEPTION, Emitters::CATCH_REGISTER_EXCEPTION)
```

compiler/src/main/java/com/dylibso/chicory/compiler/internal/EmitterMap.java
evacchi: yes, just use the same names for Emitters and CompilerOpCode, see the inline suggestions

evacchi: either way is fine (CATCH vs CATCH_COMPARE_TAG as long as it's the same)

compiler/src/main/java/com/dylibso/chicory/compiler/internal/EmitterMap.java
evacchi: again, either name is fine as long as the two match

evacchi (approved):
no objections except a nit on naming of the Emitters vs CompilerOpCode

compiler/src/main/java/com/dylibso/chicory/compiler/internal/EmitterMap.java
andreaTP: got it :+1:

andreaTP:
Thanks!

PR #920, opened 2025-05-29
No experimental compiler

andreaTP:
This is the full rename from `aot` to `compiler` and taking the bits out of experimental for the next release.

I'd appreciate a careful review (test things locally etc. etc.) before merging.


andreaTP:
The failure on the Android CI is reproducible but seems unrelated with this PR.
I'd try to submit another PR to fix it.

docs/docs/usage/annotations.md
```diff
@@ -113,12 +113,12 @@ In order to use host modules, you need to import the relevant annotations, e.g.
 ```xml
 <dependency>
   <groupId>com.dylibso.chicory</groupId>
-  <artifactId>host-module-annotations-experimental</artifactId>
+  <artifactId>host-annotations</artifactId>
```
chirino, line 116: Is the `host` part of the name here because this originally existed to hold host function annotations?  Or is it because your using them on the java host?  Do we ever see a different type of annotation than host annotations?  If not maybe we can simplify  the name.

compiler-maven-plugin/src/main/java/com/dylibso/chicory/build/time/maven/ChicoryCompilerGenMojo.java
```diff
@@ -17,8 +17,11 @@
 /**
  * This plugin generates an invokable library from the compiled Wasm
  */
-@Mojo(name = "wasm-aot-gen", defaultPhase = LifecyclePhase.GENERATE_SOURCES, threadSafe = true)
-public class AotGenMojo extends AbstractMojo {
+@Mojo(
+        name = "chicory-compiler-gen",
```
chirino, line 21: mvn plugins already have the group id as namespace.  So I think this should be something simpler like `name = "compile"`

chirino:
Just spotted a few places where we can maybe simplify naming.  I'll test out the branch later tonight to see if I spot anything else.

docs/docs/usage/annotations.md
andreaTP: Historically it started as `host-module` as `@HostModule` was the only "main" annotation.
I kept `host` as, I think, it scopes a little the domain, i.e.: here there are annotations to be used for the Java Host

Not sure about the future, maybe we are going to need more annotations, for example, to control the execution of a module (think plain vs. virtual Threads etc.), and I don't know if we would use the same module.
Maybe I'm just overthinking and `annotations` is enough...

chirino: Yeah given annotations are slim anyways... I think it would be ok have annotations for different use cases in one module.

andreaTP: done :+1:

annotations/it/pom.xml
```diff
@@ -8,16 +8,20 @@
...
-    <!-- Dummy dependency to force the order -->
+    <!-- Dummy dependencies to force the order -->
     <dependency>
       <groupId>com.dylibso.chicory</groupId>
-      <artifactId>host-module-processor-experimental</artifactId>
+      <artifactId>annotations-processor</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.dylibso.chicory</groupId>
+      <artifactId>wasi</artifactId>
```
andreaTP, line 24: this missing dependency was a bug in the build :sweat_smile: fixed now

andreaTP:
@chirino I'll go ahead and merge this one, we can fix on top if you have additional comments.

compiler-maven-plugin/src/main/java/com/dylibso/chicory/build/time/maven/ChicoryCompilerGenMojo.java
chirino: I disagreee.  it's always define in the scope of the plugin so it's obvious your doing a chicory compile.

```xml
    <plugin>
      <groupId>com.dylibso.chicory</groupId>
      <artifactId>chicory-compiler-maven-plugin</artifactId>
      <executions>
        <execution>
          <id>compile</id>
          <goals>
            <goal>compile</goal>
          </goals>
        </execution>
      </executions>
    </plugin>
```

andreaTP: Ok, integrating the change in an iteration on #922

PR #924, opened 2025-05-29
build: fail build if a wildcard import is used.

chirino:
We used to just remove it which can cause the user more pain.  This should improve the situation identified in #309.


andreaTP:
Are unused imports still removed when necessary?

I'll give it a spin next week

PR #931, opened 2025-06-02
chore - Catches in a dedicated field

andreaTP:
This is a cleanup of the EH implementation, moving the catches to their own field should be easier and cleaner.

### DIFF
--- a/wasm/src/test/java/com/dylibso/chicory/wasm/ParserTest.java
+++ b/wasm/src/test/java/com/dylibso/chicory/wasm/ParserTest.java
@@ -208,7 +208,7 @@ public class ParserTest {
 
     @Test
     public void shouldParseSIMD() throws IOException {
-        try (InputStream is = getClass().getResourceAsStream("/wasm/simd_load.0.wasm")) {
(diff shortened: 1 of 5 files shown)


wasm/src/test/java/com/dylibso/chicory/wasm/ParserTest.java
```diff
@@ -208,7 +208,7 @@ public void shouldParseNamesSection() throws IOException {
 
     @Test
     public void shouldParseSIMD() throws IOException {
-        try (InputStream is = getClass().getResourceAsStream("/wasm/simd_load.0.wasm")) {
+        try (InputStream is = ParserTest.class.getResourceAsStream("/wasm/simd_load.0.wasm")) {
```
chirino, line 211: is this related?

evacchi (approved):

evacchi:
looks good!

wasm/src/test/java/com/dylibso/chicory/wasm/ParserTest.java
andreaTP: No, just noticed it and fixed along, it's a small diff :-)
We use this convention in this repo, and using getClass gives issues on Android FWIW

PR #933, opened 2025-06-03
dev-ux: auto format for developers, only check formatting in CI

chirino:
this should make dev cycles require less keystrokes.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -553,15 +553,6 @@
             </ktfmt>
           </kotlin>
         </configuration>
-        <executions>
-          <execution>
-            <id>format</id>
-            <goals>
-              <goal>check</goal>
-            </goals>
-            <phase>process-sources</phase>
-          </execution>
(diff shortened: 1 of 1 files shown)


andreaTP (approved):
Lets give this a spin! LGTM!

pom.xml
```diff
@@ -828,6 +819,62 @@
       </build>
     </profile>
 
+    <profile>
```
andreaTP, line 822: An alternative to this Maven profile would have been moving `spotless:check` as the first goal [here](https://github.com/dylibso/chicory/blob/4a00969cc4aa51a3532f860df40083319b91b271/.github/workflows/ci.yaml#L50) ?

andreaTP:
Opened #936 as an alternative with a smaller diff.
Not fussy, I'm totally ok merging this one.

chirino:
Let's go with yours

PR #936, opened 2025-06-04
Spotless auto formatting

andreaTP:
Alternative to #933


chirino:
Will this fail the build in CI if the formatting is invalid?

chirino (approved):
Lgtm

andreaTP:
> Will this fail the build in CI if the formatting is invalid?

I double checked locally and it works as expected :+1:

PR #938, opened 2025-06-05
[chore] fix cargo projecs names

andreaTP:
Cargo project names are incorrect, we haven't noticed because we never looked at those :sweat_smile: 

Noticed in #937 , it reduces the diff.


evacchi (approved):
lol 👍

PR #941, opened 2025-06-06
refactor: WasmAnalyzer now produces Emitters instead of CompilerInstructions.,

chirino:
* Even flow control instructions are handled as an emitter
* No need for CompilerInstruction or CompilerOpcode
* Simpler emitter interface
* Moves all label management to the Analyzer.

This change is stacked on #918 
It's better to view a diff between this and #918 than main.


chirino:
Closed in favor of https://github.com/chirino/chicory/pull/9/files so that it's easier to review the diff.

PR #942, opened 2025-06-06
Fix tail call optimization bug in the interpreter

andreaTP:
@evacchi is implementing Tail Call Optimization in wazero, and suggested to test SQLite as they found bugs there.
The compiler works fine, but the interpreter crashes due to a bug, you can reproduce with this diff:

https://github.com/roastedroot/sqlite4j/compare/main...andreaTP:sqlite4j:check-tail-call

I'm not 100% confident about the implementation, as there are many interpolations that are still working, but with those changes the testsuite passes.
No issues found with the compiler :+1:


andreaTP:
Not a self contained one unfortunately.
I'd need to extract it from the SQLite compiled module.

evacchi:
I have a few reproducers now, will share them soon

EDIT they are on this branch of you want to take a peek already https://github.com/evacchi/wazero/tree/tail-call

evacchi:
specifically:

- https://github.com/evacchi/wazero/blob/43e7020053f6f51b8a570062ec15da36a13ae626/internal/integration_test/engine/tailcall_test.go
- https://github.com/evacchi/wazero/blob/43e7020053f6f51b8a570062ec15da36a13ae626/internal/engine/wazevo/e2e_test.go#L843-L898

I'll share wat files soon if you need them :)

andreaTP:
Tests added, hopefully everything is covered :crossed_fingers:

evacchi (approved):

PR #943, opened 2025-06-08
feat: add DWARF debugging support and source mapping infrastructure

chirino:
- Add comprehensive SMAP (Source Map) parsing infrastructure to runtime module
- Add new dwarf-rust module for parsing DWARF debugging information from WebAssembly modules to SMAP stratums
- Enhance InterpreterMachine to use parsed SMAP stratums to enhance stack traces.
- Enhance compiler to add SMAPs to class files and use to enhance stack traces.

This enhancement enables developers to get meaningful stack traces with original source line numbers
when debugging WebAssembly modules compiled from high-level languages like Rust and Go. 

Where you would get exceptions that looked like.
```
com.dylibso.chicory.runtime.TrapException: Trapped on unreachable instruction
	at com.dylibso.chicory.runtime.InterpreterMachine.THROW_UNREACHABLE(InterpreterMachine.java:2212)
	at com.dylibso.chicory.runtime.InterpreterMachine.eval(InterpreterMachine.java:182)
	at com.dylibso.chicory.runtime.InterpreterMachine.call(InterpreterMachine.java:100)
	at com.dylibso.chicory.runtime.InterpreterMachine.CALL(InterpreterMachine.java:1715)
	... more of the same …
	at com.dylibso.chicory.runtime.InterpreterMachine.eval(InterpreterMachine.java:550)
	at com.dylibso.chicory.runtime.InterpreterMachine.call(InterpreterMachine.java:100)
	at com.dylibso.chicory.runtime.InterpreterMachine.call(InterpreterMachine.java:65)
	at com.dylibso.chicory.runtime.Instance$Exports.lambda$function$0(Instance.java:219)
	at com.dylibso.chicory.testing.MachinesTest.lambda$shouldEmitUnderstandableStackTraces$0(MachinesTest.java:300)
	at org.junit.jupiter.api.AssertThrows.assertThrows(AssertThrows.java:53)
	at org.junit.jupiter.api.AssertThrows.assertThrows(AssertThrows.java:35)
	at org.junit.jupiter.api.Assertions.assertThrows(Assertions.java:3128)
	at com.dylibso.chicory.testing.MachinesTest.shouldEmitUnderstandableStackTraces(MachinesTest.java:300)
```

You now get exceptions that look like:
```
com.dylibso.chicory.runtime.TrapException: Trapped on unreachable instruction
	at 0x006721: chicory interpreter.rust_panic_with_hook(library/std/src/sys/pal/wasm/../unsupported/common.rs:28)
	at 0x005cc6: chicory interpreter.{closure#0}(library/std/src/panicking.rs:699)
	at 0x005c00: chicory interpreter.__rust_end_short_backtrace<std::panicking::begin_panic_handler::{closure_env#0}, !>(library/std/src/sys/backtrace.rs:168)
	at 0x00627d: chicory interpreter.begin_panic_handler(library/std/src/panicking.rs:697)
	at 0x007e74: chicory interpreter.panic_nounwind_fmt(library/core/src/panicking.rs:117)
	at 0x007ec8: chicory interpreter.panic_nounwind(library/core/src/panicking.rs:218)
	at 0x001cf6: chicory interpreter.func_30(/rustc/17067e9ac6d7ecb70e50f92c1944e545188d2359/library/core/src/ub_checks.rs:68)
	at 0x003948: chicory interpreter.from_raw_parts<u8>(/rustc/17067e9ac6d7ecb70e50f92c1944e545188d2359/library/core/src/ub_checks.rs:75)
	at 0x000d7f: chicory interpreter.count_vowels(src/lib.rs:23)
	at com.dylibso.chicory.runtime.Instance$Exports.lambda$function$0(Instance.java:219)
	at com.dylibso.chicory.testing.MachinesTest.lambda$shouldEmitUnderstandableStackTraces$4(MachinesTest.java:299)
```

and if your using the compiler: 

```
com.dylibso.chicory.runtime.TrapException: Trapped on unreachable instruction
	at com.dylibso.chicory.$gen.CompiledMachineShaded.throwTrapException(Shaded.java:195)
	at com.dylibso.chicory.$gen.CompiledMachineFuncGroup_0.rust_panic_with_hook(library/std/src/sys/pal/wasm/../unsupported/common.rs:28)
	at com.dylibso.chicory.$gen.CompiledMachineFuncGroup_0.{closure#0}(library/std/src/panicking.rs:699)
	at com.dylibso.chicory.$gen.CompiledMachineFuncGroup_0.__rust_end_short_backtrace<std::panicking::begin_panic_handler::{closure_env#0}, !>(library/std/src/sys/backtrace.rs:168)
	at com.dylibso.chicory.$gen.CompiledMachineFuncGroup_0.begin_panic_handler(library/std/src/panicking.rs:697)
	at com.dylibso.chicory.$gen.CompiledMachineFuncGroup_0.panic_nounwind_fmt(library/core/src/panicking.rs:117)
	at com.dylibso.chicory.$gen.CompiledMachineFuncGroup_0.panic_nounwind(library/core/src/panicking.rs:218)
	at com.dylibso.chicory.$gen.CompiledMachineFuncGroup_0.func_30(/rustc/17067e9ac6d7ecb70e50f92c1944e545188d2359/library/core/src/ub_checks.rs:68)
	at com.dylibso.chicory.$gen.CompiledMachineFuncGroup_0.from_raw_parts<u8>(/rustc/17067e9ac6d7ecb70e50f92c1944e545188d2359/library/core/src/ub_checks.rs:75)
	at com.dylibso.chicory.$gen.CompiledMachineFuncGroup_0.count_vowels(src/lib.rs:22)
	at com.dylibso.chicory.$gen.CompiledMachineFuncGroup_0.call_11(com/dylibso/chicory/$gen/CompiledMachine)
	at com.dylibso.chicory.$gen.CompiledMachineMachineCall.call(Unknown Source)
	at com.dylibso.chicory.$gen.CompiledMachine.call(wasm)
	at com.dylibso.chicory.runtime.Instance$Exports.lambda$function$0(Instance.java:219)
	at com.dylibso.chicory.testing.MachinesTest.lambda$shouldEmitUnderstandableStackTracesCompiled$0(MachinesTest.java:370)
	at org.junit.jupiter.api.AssertThrows.assertThrows(AssertThrows.java:53)
	at org.junit.jupiter.api.AssertThrows.assertThrows(AssertThrows.java:35)
	at org.junit.jupiter.api.Assertions.assertThrows(Assertions.java:3128)
	at com.dylibso.chicory.testing.MachinesTest.shouldEmitUnderstandableStackTracesCompiled(MachinesTest.java:370)
```


chirino:
If we combine with the function demangling support  in #937 we should be in a really nice spot.

dwarf-rust/src/test/resources/com/dylibso/chicory/dwarf/rust/ApprovalTest.verifyCountVowels.approved.txt
```diff
@@ -0,0 +1,2316 @@
+// class version 55.0 (55)
+// access flags 0x30
+final class com/dylibso/chicory/$gen/CompiledMachineFuncGroup_0 {
+
+  // compiled from: com/dylibso/chicory/$gen/CompiledMachine
+  // debug info: SMAP
+com/dylibso/chicory/$gen/CompiledMachine.java
+WASM
+*S WASM
+*F
```
chirino: typically SMAPs are put in the debug info section of the class file.  At least this is what JSPs do.  Sadly I don't know of an easy way to get this value back out at runtime.  So I'm putting it in the class a 2nd time in an SMAP field.  Does anyone one know of an easy way to get it back out?  I assume we don't want to require asm in the runtime path.  And we need to get it back so we can enhance the exception stack traces.

Given I'm not sure IDEs will be able to deal with our SMAPs in the debug info.  Should I omit putting in the debug info?   Or should I leave it in for a bit while we test out what IDEs can do with it?

machine-tests/src/test/java/com/dylibso/chicory/testing/MachinesTest.java
```diff
@@ -279,4 +282,106 @@ public void shouldCallIndirectAotToInterpreter() {
...
+    @Test
+    public void shouldEmitUnderstandableStackTracesCompiled() throws Exception {
+        WasmModule module = loadModule("compiled/count_vowels.rs.wasm");
+        var instance =
+                Instance.builder(module)
+                        .withDebugParser(DebugParser::parse)
+                        .withMachineFactory(
+                                MachineFactoryCompiler.builder(module)
+                                        .withDebugParser(DebugParser::parse)
+                                        .compile())
```
chirino: I do wish there was a way we could avoid having to repeat options like `withDebugParser` at the instance and compiler level.

chirino:
the dwarf-rust module uses the https://github.com/chirino/wasm-source-map rust module.  It's a tight coupling, should we inline the source similar to the wasm-corpus stuff?

dwarf-rust/src/main/wasm/README.md
```diff
@@ -0,0 +1,6 @@
+# Attribution
+
+wasm-source-map.wasm was built from:
+
+https://github.com/chirino/wasm-source-map
```
evacchi: cool! maybe add a commit hash just so we know where this version comes from

machine-tests/src/test/java/com/dylibso/chicory/testing/MachinesTest.java
evacchi: feels like something that might go on the Store? 🤔

evacchi:
so from a cursory look, basically we are now extracting DWARF metadata into SMAP (they live!) and using that? pretty cool

chirino:
> so from a cursory look, basically we are now extracting DWARF metadata into SMAP (they live!) and using that? pretty cool

that's exactly what's happening.

chirino:
And now we extract the function names from the dwarf info.

dwarf-rust/src/main/wasm/README.md
chirino: done.

dwarf-rust/src/test/resources/com/dylibso/chicory/dwarf/rust/ApprovalTest.verifyCountVowels.approved.txt
```diff
@@ -0,0 +1,2403 @@
+// class version 55.0 (55)
```
andreaTP: Approval tests are high maintenance as they often cause conflicts and require bulk updates.
Ideally they are intended to be used to have a reference in the codebase that a human can read and reason about, the smallest the better.

In this case the golden master is:
- big
- non human readable for most part

There are a few things that can be done to improve the situation, but the question is:
what are we really testing with this approval?

dwarf-rust/src/test/java/com/dylibso/chicory/dwarf/rust/InstanceTest.java
```diff
@@ -0,0 +1,50 @@
+package com.dylibso.chicory.dwarf.rust;
+
+import com.dylibso.chicory.runtime.internal.smap.ParserException;
+import org.junit.jupiter.api.Test;
+
+public class InstanceTest {
+
+    //    private InputStream getResource(String wasmResource) {
```
andreaTP: leftovers?

dwarf-rust/src/main/wasm/README.md
```diff
@@ -0,0 +1,3 @@
+# Attribution
+
+wasm-source-map.wasm was built from https://github.com/chirino/wasm-source-map/tree/eb37f99319379950b2e50c21d4a90d657becfd5b
```
andreaTP: I checked the repo and the license is BSD 3, is it compatible to include this compiled artifact in an Apache project?
Additionally, in the external repo there is not a lot of code and is mostly tested here, I propose:
- preferred - is it possible to keep the rust library and the Java bindings in an external repo? (e.g. roastedroot)
- alternatively - have some basic testing and build everything here in an `experimental` module

dwarf-rust/pom.xml
```diff
@@ -0,0 +1,92 @@
...
+      <groupId>com.dylibso.chicory</groupId>
+      <artifactId>wasi</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.dylibso.chicory</groupId>
+      <artifactId>wasm</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.google.code.gson</groupId>
+      <artifactId>gson</artifactId>
```
andreaTP: Any specific reason for using Gson here?
We usually, use Jackson as it plays well with the rest of the ecosystem (Quarkus etc.)

runtime/src/main/java/com/dylibso/chicory/runtime/internal/smap/Smap.java
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    public List<Stratum> getStrata() {
+        return strata;
+    }
+
+    // *********************************************************************
+    // For testing (and as an example of use)...
+
+    public static void main(String[] args) {
```
andreaTP: Personally, I'm not a big fan of leaving `main` methods around ... is it any useful? Maybe move it to a test?

runtime/src/main/java/com/dylibso/chicory/runtime/internal/smap/Stratum.java
```diff
@@ -0,0 +1,588 @@
...
+package com.dylibso.chicory.runtime.internal.smap;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+
+/**
+ * Represents the line and file mappings associated with a JSR-045 "stratum".
+ */
+public class Stratum {
```
andreaTP: What is "Stratum"?

runtime/src/test/java/com/dylibso/chicory/runtime/internal/smap/SmapParserTest.java
```diff
@@ -0,0 +1,65 @@
...
+                        + "*L\n"
+                        + "1,5:10,2\n"
+                        + "1#1,2:20,2\n"
+                        + "7#0,2:24,2\n"
+                        + "*E\n",
+                generator.toString());
+    }
+
+    @Test
+    @Disabled("works on my machine.. not others :( ")
```
andreaTP: what?

runtime/src/main/java/com/dylibso/chicory/runtime/InterpreterMachine.java
```diff
@@ -2188,6 +2200,72 @@ private static int numberOfValuesToReturn(Instance instance, AnnotatedInstructio
...
+            }
+        }
+
+        var codeSectionAddress = instance.module().codeSection().address();
+        // add back the wasm debug info.
+        for (int i = te.getCallStackAddresses().length - 1; i >= 0; i--) {
+            var address = te.getCallStackAddresses()[i];
+            var funcId = te.getCallStackFunctionIds()[i];
+
+            var functionName = stratum.getFunctionMapping(address - codeSectionAddress);
```
andreaTP: having a little Markdown table somewhere(even in a comment) that represents those hidden informations(completely made up):

| info | dwarf address | smap address |
|---|---|---|
| function id | .debug_info + byte offset in section | ... |

wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java
```diff
@@ -283,7 +283,7 @@ private void parse(InputStream in, ParserListener listener, boolean decode) {
             var sectionSize = readVarUInt32(buffer);
 
             validator.validateSectionType(sectionId);
-
+            var sectionAddress = buffer.position();
```
andreaTP, line 286: We are already doing it in a few places, but, I think that `buffer.position` is not a great "source of truth", I think we need a slightly more principled approach.

Additionally the module with the debug information can be separate, and we should account for this scenario.

dwarf-rust/src/test/resources/com/dylibso/chicory/dwarf/rust/ApprovalTest.verifyCountVowels.approved.txt
chirino: It lets developers see the data formats that are being:
* sent from wasm-source-map to the DebugParser class
* stored int the compiled compiled classes

I don't think there is another place to see those formats.

Additionally we want to make sure that those formats don't change with each run.  

If those goals are not important I'll remove the tests.

dwarf-rust/src/main/wasm/README.md
chirino: Yes the ASF allows BSD and MIT code to back packaged with it's projects, so It's safe. 

I like the idea of moving it into experimental.

dwarf-rust/pom.xml
chirino: no reason, just simpler.

runtime/src/main/java/com/dylibso/chicory/runtime/internal/smap/Smap.java
chirino: we can drop it.  it was part of the original source file.

runtime/src/main/java/com/dylibso/chicory/runtime/internal/smap/Stratum.java
chirino: It's a term defined in the JSR45 spec.

runtime/src/test/java/com/dylibso/chicory/runtime/internal/smap/SmapParserTest.java
chirino: Lets try enabling again. :)

wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java
chirino: I'm not sure why you think this is a bad source of truth.
buffer.position() seems to give us the current address relative to the start of the wasm module which is exactly the information that I need.

This should work for modules that the debug information has be separated.  The DebugParser does not need the code address.  it only parses the debug sections and what it produces is a Stratum with addresses that are relative to the code section.  So if that code section gets moved or placed in a different module, the Stratum is still valid since the addresses are relative.

runtime/src/main/java/com/dylibso/chicory/runtime/InterpreterMachine.java
chirino: I'll add some more docs there.. not sure exactly what your looking for tho.

dwarf-rust/src/test/resources/com/dylibso/chicory/dwarf/rust/ApprovalTest.verifyCountVowels.approved.txt
andreaTP: I think that the goals are important, they just don't fit an approval tests in the current implementation :slightly_smiling_face: 

Either:
- use a very very small example
- check the presence of specific elements other than the full content

does it make sense?

dwarf-rust/pom.xml
andreaTP: ok, I'd stay on the safe side with Jackson if possible(it's a 4 locs change), ideally we should avoid any external dependency when possible (and it might not be easily possible in this case).

runtime/src/main/java/com/dylibso/chicory/runtime/internal/smap/Smap.java
andreaTP: it reminds me too much the time spent debugging Hadoop and similar :sweat_smile:

runtime/src/main/java/com/dylibso/chicory/runtime/internal/smap/Stratum.java
andreaTP: Found it! For the future self:

> StratumSection
An SMAP may map more than one [translated-source](file:///home/aperuffo/.cache/.fr-KZE472/dsol_spec.html#TranslatedSource) to the output source (the output source is [final-source](file:///home/aperuffo/.cache/.fr-KZE472/dsol_spec.html#FinalSource) if the SMAP is in a [SourceDebugExtension](file:///home/aperuffo/.cache/.fr-KZE472/dsol_spec.html#SourceDebugExtension)). A view of the source is a stratum (whether viewed as translated-source or final-source). Each translated-source language should have its own stratum section with a unique stratum name. The final-source stratum (named "Java") is created automatically and should not have a stratum section. The stratum section should be followed by a file section and a line section which will be associated with that stratum.
The format of the section is simply the stratum section marker "*S" followed by the name of the stratum. The section ends with a line termination. One [FileSection](file:///home/aperuffo/.cache/.fr-KZE472/dsol_spec.html#FileSectionDesc) and one [LineSection](file:///home/aperuffo/.cache/.fr-KZE472/dsol_spec.html#LineSectionDesc) (in either order) must follow the StratumSection (before the next StratumSection or the [EndSection](file:///home/aperuffo/.cache/.fr-KZE472/dsol_spec.html#EndSectionDesc)). One or more [VendorSection](file:///home/aperuffo/.cache/.fr-KZE472/dsol_spec.html#VendorSectionDesc)s may follow a StratumSection. There must be at least one [StratumSection](file:///home/aperuffo/.cache/.fr-KZE472/dsol_spec.html#StratumSection).

I think that adding the first line in a comment is a good reference:

> An SMAP may map more than one [translated-source](file:///home/aperuffo/.cache/.fr-KZE472/dsol_spec.html#TranslatedSource) to the output source (the output source is [final-source](file:///home/aperuffo/.cache/.fr-KZE472/dsol_spec.html#FinalSource) if the SMAP is in a [SourceDebugExtension](file:///home/aperuffo/.cache/.fr-KZE472/dsol_spec.html#SourceDebugExtension)). A view of the source is a stratum (whether viewed as translated-source or final-source).

wdyt?

runtime/src/main/java/com/dylibso/chicory/runtime/InterpreterMachine.java
andreaTP: I'm trying to grasp what is connected to what, e.g. code section instruction number 9 -> smap position XYZ -> stratum info at etc. etc.

wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java
andreaTP: Thanks for the explanation, so the code address is needed to calculate the offset of the module used during runtime, and that's why we are adding the information to the wasm data structure, correct?

dwarf-rust/pom.xml
chirino: changed.

dwarf-rust/src/main/wasm/README.md
chirino: does the module need to be renamed and the rust source added?

dwarf-rust/src/test/java/com/dylibso/chicory/dwarf/rust/InstanceTest.java
chirino: deleted.

runtime/src/main/java/com/dylibso/chicory/runtime/internal/smap/Stratum.java
chirino: Sounds good

wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java
chirino: Yes.  At runtime for the interpreter when it maps instruction addresses to source lines.  

At compile time for the compiler.

dwarf-rust/src/test/resources/com/dylibso/chicory/dwarf/rust/ApprovalTest.verifyCountVowels.approved.txt
chirino: I'll take them out.  Maybe you can help me figure out how to do it better.

runtime/src/main/java/com/dylibso/chicory/runtime/InterpreterMachine.java
chirino: Did the comment I added help?

runtime/src/test/java/com/dylibso/chicory/runtime/internal/smap/SmapParserTest.java
```diff
@@ -0,0 +1,81 @@
...
+                        + "*E\n",
+                generator.toString());
+    }
+
+    @Test
+    public void testParseExampleSmap() throws ParserException, IOException {
+        // Read the example.smap file from test resources
+        try (var inputStream = getClass().getResourceAsStream("/example.smap")) {
+            assertNotNull(inputStream, "example.smap resource not found");
+            String actual = new String(inputStream.readAllBytes(), StandardCharsets.UTF_8);
```
chirino: so I'm banging my head trying to figure out why this fails on windows and then I remember that git on windows checks out text files with different line ending!  We always generate with "\n".  I think I need to check this in as a binary file.

runtime/src/test/java/com/dylibso/chicory/runtime/internal/smap/SmapParserTest.java
chirino: it was the example was getting checked out with different line endings on windows !

docs/docs/experimental/dwarf-debug-symbols.md
```diff
@@ -0,0 +1,156 @@
...
+import com.dylibso.chicory.dwarf.rust.DebugParser;
+import com.dylibso.chicory.compiler.MachineFactoryCompiler;
+import com.dylibso.chicory.runtime.Instance;
+import com.dylibso.chicory.wasm.WasmModule;
+
+var module = Parser.parse(new File("your.wasm"));
+Instance instance = Instance.builder(module)
+        .withMachineFactory(
+                MachineFactoryCompiler.builder(module)
+                        .withDebugParser(DebugParser::parse)
```
andreaTP: Is it possible to use a uniform API across compiler and interpreter?

dwarf-rust-experimental/rust/target/wasm32-wasip1/release/dwarf-rust.wasm
andreaTP, line 1: We use `target` folders for the ephemeral part of the build, I'd cache this file in the root of `dwarf-rust-experimental`, but I understand this might be more of a personal preference.

dwarf-rust-experimental/rust/src/main.rs
```diff
@@ -0,0 +1,502 @@
...
+use std::convert::TryInto;
+use std::io::Read;
+use wasm::{parse_sections, SectionKind};
+
+pub struct Pos {
+  line: u32,
+  // column: u32,
+}
+
+// enum FuncState {
```
andreaTP: clean commented code?

dwarf-rust-experimental/rust/src/main.rs
```diff
@@ -0,0 +1,502 @@
...
+              _ => None,
+            }
+          } else {
+            None
+          };
+
+          // If we have all the information, add it to our functions map
+          if let (Some(func_name), Some(start), Some(end)) = (func_name, low_pc, high_pc) {
+            // Filter out synthetic/relocated addresses similar to line processing
+            const MAX_REALISTIC_WASM_ADDR: u64 = 0x40000000; // 1GB threshold
```
andreaTP, line 180: Why is this needed? Have you found related issues?

dwarf-rust-experimental/rust/src/main.rs
```diff
@@ -0,0 +1,502 @@
...
+  }
+
+  if row.basic_block() {
+    score = score.saturating_sub(200); // Penalty for basic block boundaries
+  }
+
+  // Strategy 1: Prefer user code over library/dependency code
+  if let Some(dir) = directory {
+    if dir.contains("/rustc/") || dir.contains("/rust/deps/") {
+      score -= 300; // Heavily penalize compiler/dependency paths
```
andreaTP, line 371: Where those numbers are coming from?
How/when they should be changed?

dwarf-rust-experimental/rust/src/main.rs
```diff
@@ -0,0 +1,502 @@
...
+      .unwrap();
+      println!("Regenerated test data in ./src/count_vowels.wasm.json");
+    }
+
+    // read count_vowels.rs.wasm.json
+    let expected: SourceResult = serde_json::from_reader(
+      std::fs::File::open("./src/count_vowels.wasm.json").expect("could not read json file"),
+    )
+    .expect("json failed");
+    assert_eq!(actual, expected);
```
andreaTP: this looks to be the library to use in Rust:
https://github.com/mitsuhiko/insta

That said `count_vowels.wasm.json` is too big, there is no chance we can effectively inspect it... we probably need something different than an approval.

dwarf-rust-experimental/rust/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "dwarf-rust"
+version = "0.1.0"
+description = "WebAssembly DWARF parser and location resolver"
+authors = ["Ingvar Stepanyan <rreverser@google.com>", "Hiram Chirino <hiram@hiramchirino.com>"]
```
andreaTP, line 5: ?

runtime/src/main/java/com/dylibso/chicory/runtime/Instance.java
```diff
@@ -373,6 +379,11 @@ public Builder withStart(boolean s) {
             return this;
         }
 
+        public Builder withDebugParser(Function<WasmModule, Stratum> debugParser) {
```
andreaTP: `Stratum` is in the `internal` package, it should not spill to the public API.

andreaTP (commented):
This is a great improvement!
I think we should agree on a few points before moving on with this:

- `Stratum` is currently in the `internal` package, and should not spill to the public API, would it be possible to extract a "generic data structure" that can be eventually used by `Stratum`?
- in this proposal compiler and interpreter are using different mechanisms to enable debugging, I think we should make an effort to keep the API aligned
- what happens when the dwarf mechanism fails?
- all in all, I'd still prefer to start shipping from an external repository, but this PR should go on with the core changes to enable the feature

docs/docs/experimental/dwarf-debug-symbols.md
chirino: I would love that too but I'm not sure how to accomplish it.

dwarf-rust-experimental/rust/src/main.rs
chirino: will do.

dwarf-rust-experimental/rust/src/main.rs
chirino: yeah.. I've seen some line mappings that are too large, obviously don't map into the WASM module address range, so might a well not include it the line mapping results.  What might be better if if we are told what the last address of the code section is so that we can omit all line numbers that are not within the range.

dwarf-rust-experimental/rust/src/main.rs
chirino: they are just heuristics.   I'm assuming a code address can map to multiple source lines.  When this happens we need to pick which one of the lines wins..   The good news is I think any of those lines would be valid responses, but I think we should try to pick the one that is the user's source file instead of something like a rust stdlib function that gets inlined.  I think we will need more examples and user feedback to get this exactly right.

dwarf-rust-experimental/rust/src/main.rs
chirino: The file is not there to be inspected line by line.  It's there to demonstrate an example of the output that this module produces.  Instead of reading rust code to figure out how to implement the java parser for it, I just look at the example data file it produces to understand it's structure so I can write the jackson bindings.

dwarf-rust-experimental/rust/Cargo.toml
chirino: > ?

fix the name ?

runtime/src/main/java/com/dylibso/chicory/runtime/Instance.java
chirino: I think I can do something about that while keeping the Stratum api internal.

dwarf-rust-experimental/rust/Cargo.toml
andreaTP: yeah, well, not sure where this project setup is coming from :sweat_smile:

dwarf-rust-experimental/rust/src/main.rs
andreaTP: I see, can we have a super minimal one?

dwarf-rust-experimental/rust/src/main.rs
andreaTP: Thanks for explaining!

> I think we will need more examples and user feedback to get this exactly right.

How do we get the user feedback?

dwarf-rust-experimental/rust/src/main.rs
andreaTP: Do you happen to know where are those coming from? Synthetic functions?
Should we just drop them when loading?

docs/docs/experimental/dwarf-debug-symbols.md
andreaTP: I have a couple of ideas, but maybe I'm not aware of all the invariants, let me know if this point gets stuck and I'll have a look.

chirino: Please go for it.

dwarf-rust-experimental/rust/target/wasm32-wasip1/release/dwarf-rust.wasm
chirino: I could make the build script copy it out to the Java side of the source tree.

chirino: I could make the build script copy it out to the Java side of the source tree.

dwarf-rust-experimental/rust/src/main.rs
chirino: Bug reports? Does not seem great.  The best strategy is likely to add more examples and examine each line mapping to see if it is actually the best.  But this seems very labor intensive.

andreaTP: This looks like a good reason to leave it out of the main repo (just for now).

Anyhow, having a little doc to explain how to test and tune those parameters seems like a feasible first step.

docs/docs/experimental/dwarf-debug-symbols.md
andreaTP: I'm looking into, what do you think about moving the `debugParser` to be an optional field of `WasmModule`?
In this way it can be easily applied either in the Instance and in any Machine.

chirino: sounds good.

chirino:
squashed and rebased on main

chirino:
I've removed the dwarf-parser, so that we can just host that externally.

chirino:
Added back the dwarf-parser as it needed to really test things end to end.

compiler/src/main/java/com/dylibso/chicory/compiler/internal/Compiler.java
```diff
@@ -92,7 +97,8 @@
 
 public final class Compiler {
 
-    public static final String DEFAULT_CLASS_NAME = "com.dylibso.chicory.$gen.CompiledMachine";
+    public static final String DEFAULT_CLASS_NAME = "com.dylibso.chicory.$gen%d.CompiledMachine";
```
andreaTP, line 100:
```suggestion
    public static final String DEFAULT_CLASS_NAME = "com.dylibso.chicory.$gen.CompiledMachine%d";
```
nit

compiler-maven-plugin/src/main/java/com/dylibso/chicory/build/time/maven/ChicoryCompilerGenMojo.java
```diff
@@ -76,6 +79,19 @@ public class ChicoryCompilerGenMojo extends AbstractMojo {
...
+            for (var service : ServiceLoader.load(DebugParser.class)) {
+                debugParser = service;
+                getLog().info("Using debug parser: " + debugParser.getClass().getName());
+                break;
+            }
+        } catch (ServiceConfigurationError ignore) {
+            debugParser = null;
+        }
+
         getLog().info("Generating AOT classes for " + name + " from " + wasmFile);
```
andreaTP, line 95: unrelated - note to self(or anyone else) - remove `AOT` in this sentence

compiler/src/main/java/com/dylibso/chicory/compiler/internal/Compiler.java
```diff
@@ -119,19 +125,42 @@ public final class Compiler {
     private final InterpreterFallback interpreterFallback;
     private final List<FunctionType> functionTypes;
     private final Map<String, byte[]> extraClasses = new LinkedHashMap<>();
+    private final Map<String, byte[]> extraResources = new LinkedHashMap<>();
```
andreaTP, line 128: as discussed, `extraResources`, seems like a good addition even "per-se" and future development :+1:

compiler/src/main/java/com/dylibso/chicory/compiler/internal/Compiler.java
```diff
@@ -437,6 +499,18 @@ private Consumer<ClassVisitor> emitFunctionGroup(int start, int end, String inte
...
+            if (!debugContext.outputStratum.isEmpty()) {
+                // add the source map entries to the class
+                var baseName = internalClassName(className + funcGroupClassName);
+                var sourceFile = baseName + ".java";
+                String smap = Stratum.toSMapString(sourceFile, debugContext.outputStratum);
+                extraResources.put(
+                        baseName + ".smap", smap.getBytes(java.nio.charset.StandardCharsets.UTF_8));
+
+                if (smap.length() < 65535) {
+                    classWriter.visitSource(internalClassName, smap);
```
andreaTP, line 511: This is super interesting!

It should enable IDEs integration, have you tested it? How?
Why the limit on length?

compiler/src/main/java/com/dylibso/chicory/compiler/internal/Compiler.java
```diff
@@ -472,6 +546,27 @@ private byte[] compileClass() {
                     null);
         }
 
+        // if we have debug info
+        if (!debugContext.inputStratum.isEmpty()) {
```
andreaTP, line 550: Any performance impact?
Should we make it configurable?

compiler/src/main/java/com/dylibso/chicory/compiler/internal/Compiler.java
```diff
@@ -575,6 +670,23 @@ private static void emitCallSuper(InstructionAdapter asm) {
...
+        asm.visitTypeInsn(Opcodes.ANEWARRAY, getInternalName(Class.class));
+
+        for (int i = 0; i < funcGroupCount; i++) {
+            asm.dup();
+            asm.iconst(i);
+            String className = internalClassName + classNameForFuncGroup(i * maxFunctionsPerClass);
+            asm.aconst(Type.getObjectType(className));
+            asm.astore(OBJECT_TYPE);
+        }
+        var classNamePrefix = internalClassName.replaceAll("/", ".") + "FuncGroup_";
```
andreaTP: I think there are a couple of helper functions to not have `replaceAll` all over the places

compiler/src/main/java/com/dylibso/chicory/compiler/internal/Shaded.java
```diff
@@ -229,4 +242,63 @@ public static boolean exceptionMatches(WasmException exception, int tag, Instanc
...
+    private static Stratum[] STRATA_BY_FUNC_GROUP;
+    private static String funcGroupClassPrefix;
+
+    public static void init(Class<?>[] classes, String funcGroupClassPrefix) {
+        Shaded.funcGroupClassPrefix = funcGroupClassPrefix;
+        STRATA_BY_FUNC_GROUP = new Stratum[classes.length];
+
+        for (int i = 0; i < classes.length; i++) {
+            var clazz = classes[i];
+            String resource = "/" + clazz.getName().replace('.', '/') + ".smap";
```
andreaTP, line 255: `replaceAll` ?

andreaTP, line 255: Ideally, the class name should come from the function input parameters and not being re-extracted after the class is generated. Any specific reason for this design?

compiler/src/main/java/com/dylibso/chicory/compiler/internal/Compiler.java
```diff
@@ -575,6 +670,23 @@ private static void emitCallSuper(InstructionAdapter asm) {
                 OBJECT_TYPE.getInternalName(), "<init>", getMethodDescriptor(VOID_TYPE), false);
     }
 
+    private void compileStaticInitializer(InstructionAdapter asm, String internalClassName) {
```
andreaTP, line 673: I'm interested in better understanding the design here.
We are encoding `smap` files in the `class` files generated?
Do you mind to write a little paragraph to get me(and other reviewers) up to speed quickly?

compiler/src/main/java/com/dylibso/chicory/compiler/internal/Shaded.java
```diff
@@ -229,4 +242,63 @@ public static boolean exceptionMatches(WasmException exception, int tag, Instanc
...
+
+    private static void enhanceStackTrace(Throwable e) {
+        if (STRATA_BY_FUNC_GROUP == null) {
+            return;
+        }
+
+        var elements = e.getStackTrace();
+        for (int i = 0; i < elements.length; i++) {
+            var element = elements[i];
+            if (element.getClassName().startsWith(funcGroupClassPrefix)) {
```
andreaTP, line 278: `element.getClassName` should be supported OOTB by Graal etc., we should do a diligence check with a native image build though.

compiler/src/main/java/com/dylibso/chicory/compiler/internal/WasmAnalyzer.java
```diff
@@ -139,13 +140,45 @@ public List<CompilerInstruction> analyze(int funcId) {
...
+                }
+
+                String file = lineMapping.fileName();
+                String path = lineMapping.filePath();
+                debugContext.outputStratum.addLineData(
+                        file, path, lineMapping.line(), lineMapping.count(), outputLineNo, 1);
+
+                labels.add(idx);
+                result.add(new CompilerInstruction(CompilerOpCode.LABEL, idx));
+                result.add(
```
andreaTP, line 175: this is useful only if debug is enabled, correct?

compiler/src/main/java/com/dylibso/chicory/compiler/internal/Compiler.java
andreaTP, line 100: Thinking twice about this change, and, sadly is a breaking change as it changes both the content of a static value and it's semantics (it is now a "template").
Either we Deprecate the old name or do something about it.

docs/docs/experimental/dwarf-debug-symbols.md.off
```diff
@@ -0,0 +1,156 @@
...
+sidebar_position: 6
+sidebar_label: Dwarf Debug Symbols
+title: Dwarf Debug Symbols
+---
+
+# Dwarf Debug Symbols
+
+> Experimental – subject to change in future releases.
+
+Chicory can now read **DWARF** debug symbols embedded in a WebAssembly module and use them to enrich Java stack traces with the *original* source-level file names and line numbers that produced the trap.  This is extremely valuable when you are running modules generated from higher-level languages such as **Rust** or **Go** and need to identify exactly where a failure occurred.
```
andreaTP: `higher-level` ?
I'm a bit confused by this sentence, I think that we should mention that the DWARF information is available for languages compiling directly to WASM, as opposed to languages that are run by VM/interpreters compiled to WASM (e.g. Python or JS).
But maybe I missed something ...

docs/docs/experimental/dwarf-debug-symbols.md.off
```diff
@@ -0,0 +1,156 @@
...
+```
+
+# Getting Started
+
+## 1. Add the Maven dependency
+
+```xml
+<dependency>
+  <groupId>com.dylibso.chicory</groupId>
+  <artifactId>dwarf-rust-experimental</artifactId>
```
andreaTP, line 69: I'm fine at reviewing this PR close to the end keeping the experimental module here(e.g. is it `rust` specific?), but I think that we should grow it up externally to the main repo.
Close to the merge I'm happy to help to tight things together with publishing/repos etc.

docs/docs/experimental/dwarf-debug-symbols.md.off
```diff
@@ -0,0 +1,156 @@
...
+</dependency>
+```
+
+The artefact is purposefully isolated so the feature can evolve independently of the core runtime.  At the moment the implementation uses a Rust library compiled to WASM and then compiled to byte code to extract the dwarf debug symbols.
```
andreaTP: As you tell me often, probably users are not interested at "how the sausage is made"

docs/docs/experimental/dwarf-debug-symbols.md.off
```diff
@@ -0,0 +1,156 @@
...
+
+
+```java title="Interpreter setup"
+import com.dylibso.chicory.dwarf.rust.DebugParser;
+import com.dylibso.chicory.runtime.Instance;
+import com.dylibso.chicory.wasm.WasmModule;
+
+var module = Parser.parse(new File("your.wasm"));
+Instance instance = Instance.builder(module)
+        .withDebugParser(DebugParser::parse)
```
andreaTP, line 99: Here we need to configure differently compiler and interpreter, this is undesirable and unexpected behavior from the API perspective.

E.g.: the `.withDebugParser` is available on Instances created with the compiler, but is (likely?) going to not work.

Have you attempted to move the configuration up one level (i.e. attach the debug info to the `WasmModule`)?
Any blocker/issue?

docs/docs/experimental/dwarf-debug-symbols.md.off
```diff
@@ -0,0 +1,156 @@
...
+            <version>${project.version}</version>
+          </dependency>
+        </dependencies>
+        ...
+      </plugin>
+    </plugins>
+  </build>
+```
+
+# Limitations
```
andreaTP: This section is extremely valuable :+1: I'd move it up to settle expectations

docs/tests/index.test.ts
```diff
@@ -26,6 +26,7 @@ describe("ApprovalTests", () => {
   });
 
   it.each(markdownFiles)('test %s', (f) => {
+    console.log("Testing", f);
```
andreaTP, line 29: leftover?

dwarf-rust-experimental/rust/src/count_vowels.wasm.json
```diff
@@ -0,0 +1,15664 @@
+{
```
andreaTP: Where does this file comes from?
It's huge, is it possible to use a minimal one?

runtime/src/main/java/com/dylibso/chicory/runtime/InterpreterMachine.java
```diff
@@ -2188,6 +2211,105 @@ private static int numberOfValuesToReturn(Instance instance, AnnotatedInstructio
...
+                }
+                continue;
+            }
+
+            // we eat all the interpreter frames.
+            if (wasmDebugInfoAdded) {
+                continue;
+            }
+
+            // The following happens when we reach the first interpreter frame..
```
andreaTP, line 2263: I'm confused, we are in the interpreter here ...

runtime/src/main/java/com/dylibso/chicory/runtime/InterpreterMachine.java
```diff
@@ -2188,6 +2211,105 @@ private static int numberOfValuesToReturn(Instance instance, AnnotatedInstructio
...
+                // keep all the non-interpreter frames...
+                traces.add(trace);
+                if (wasmDebugInfoAdded) {
+                    // don't replace any more interpreter frames..
+                    interpreterFramesFound = true;
+                }
+                continue;
+            }
+
+            // we eat all the interpreter frames.
```
andreaTP, line 2258: There are a lot of rough edges around overall, since the interaction interpreter/compiler seems to cause more work I'd suggest that we start not supporting them.

E.g. we do support only fully compiled code and fully interpreted code.

We can add back support for those cases in a subsequent iteration.

runtime/src/main/java/com/dylibso/chicory/runtime/Stratum.java
```diff
@@ -0,0 +1,155 @@
...
+     * @return a Stratum instance representing the parsed SMAP
+     */
+    static Stratum parseSMapString(String smap) {
+        return SmapParser.parse(smap).getDefaultStratum().optimizeForLookups();
+    }
+
+    /**
+     * Sorts the stratum data for optimized lookups using a binary search.
+     */
+    Stratum optimizeForLookups();
```
andreaTP: This seems a pretty challenging API to be implemented by a user, do we expect ppl to provide their own implementation?

wasm-corpus/Dockerfile
```diff
@@ -63,5 +63,3 @@ ENV PATH="/opt/wabt/bin:${PATH}"
 RUN cargo install wasm-tools
 
 WORKDIR /usr/code
-
-ENTRYPOINT ["./compile.sh"]
```
andreaTP, line 67: ??

wasm-corpus/src/main/resources/rust/count_vowels/src/lib.rs
```diff
@@ -29,4 +29,16 @@ pub extern "C" fn count_vowels(ptr: i32, len: i32) -> i32 {
...
     }
     count
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_count_vowels() {
```
andreaTP, line 39: Are you using this test to extract what pure Rust will be producing?
Please, add at least a comment on how to do it?

andreaTP (changes requested):
Thanks for the great step forward!

I think this is great progress, but we need to refine a few aspects:
- unless there is a strong counter motivation the API to enable debugging should be the same across `Machines`
- There is still a lot of code "around", hard-coded in the compiler/interpreter that does a lot of assumptions on Stratum, lines of code etc.
- `Stratum` still brings behavior + data while, ideally, we should have:
  - a "data-only" API in Chicory
  - injectable behavior completely defined outside
  
To get things easier I propose that we don't support(for now) "mixed" instances (compiled + interpreter) so that we can focus on getting the APIs right with less code around.
Let me know your thoughts @chirino and thanks again!

compiler/src/main/java/com/dylibso/chicory/compiler/internal/Compiler.java
chirino: In theory it might.  But I need to figure out how to setup a project to do it. 
Seems break if it's too big.  Found it compiling the `quickjs-provider.javy-dynamic.wasm` module.

compiler/src/main/java/com/dylibso/chicory/compiler/internal/Compiler.java
chirino: It's already configurable.  If you don't supply a debug parser, the inputStratum is empty.

compiler/src/main/java/com/dylibso/chicory/compiler/internal/Shaded.java
chirino: this replace is a replace all (char arg vs string arg).

chirino: That's what's being done by the caller of this function, in the static init of the CompilerMachine.  I was just trying to reduce the amount of generated code in that static init.

compiler/src/main/java/com/dylibso/chicory/compiler/internal/Compiler.java
chirino: Basically we need to turn the smap strings ( there is one smap string per FuncGroup class) back into stratums.  So we do this work once when the CompilerMachine class is initialized.   The reason we have multiple stratums, is because each FuncGroup class has different source to line mappings.  If we could have fit the whole wasm file into a single class, we would only need 1 stratum.

compiler/src/main/java/com/dylibso/chicory/compiler/internal/WasmAnalyzer.java
chirino: yes, the if branch it's in is only true if debug is enabled.

compiler/src/main/java/com/dylibso/chicory/compiler/internal/Compiler.java
chirino: Can you give me an example of user code that would break with this change?

docs/docs/experimental/dwarf-debug-symbols.md.off
chirino: as opposed to wat / wast, but yeah lets reword.

docs/docs/experimental/dwarf-debug-symbols.md.off
chirino: We could treat them both the same if we manually configure the Interpeter's MachineFactory (not something we typically do right now).

docs/tests/index.test.ts
chirino: yeah

runtime/src/main/java/com/dylibso/chicory/runtime/InterpreterMachine.java
chirino: Only some of the frames are interpreter frames.   The interpreter frames will be sandwiched between either compiler frames or user code frames.

runtime/src/main/java/com/dylibso/chicory/runtime/InterpreterMachine.java
chirino: I already put int the hard work to get this to work. It would be a shame to drop it now.

runtime/src/main/java/com/dylibso/chicory/runtime/Stratum.java
chirino: I don't expect them to provide their own.  Thats why we supply the Stratum api.

wasm-corpus/Dockerfile
chirino: Handy if you want to run bash in the container that has all the build tools to test/debug things.

wasm-corpus/src/main/resources/rust/count_vowels/src/lib.rs
chirino: Yep.. ok

compiler/src/main/java/com/dylibso/chicory/compiler/internal/Compiler.java
andreaTP: How can we validate those changes? The limit seems arbitrary ...
Have you attempted to use a small and self-contained module?

compiler/src/main/java/com/dylibso/chicory/compiler/internal/Compiler.java
andreaTP: > If you don't supply a debug parser

Build time or runtime?
What I meant for "configurable" is:
a module compiled with debug parser that want to disable it at runtime

compiler/src/main/java/com/dylibso/chicory/compiler/internal/Shaded.java
andreaTP: `CompilerUtil.internalClassName`

runtime/src/main/java/com/dylibso/chicory/runtime/InterpreterMachine.java
andreaTP: I propose we get a first iteration that will not work for mixed mode.
It will make it easier to change things around and move forward.

runtime/src/main/java/com/dylibso/chicory/runtime/InterpreterMachine.java
andreaTP: Not a drop, just add it back at a later stage.

runtime/src/main/java/com/dylibso/chicory/runtime/Stratum.java
andreaTP: When we add a public API, it is supposed to be used by users.
Having an API that only matches 1:1 with one specific implementation is a code smell.

wasm-corpus/Dockerfile
andreaTP: ok, fine :+1:

compiler/src/main/java/com/dylibso/chicory/compiler/internal/Compiler.java
andreaTP: This is public API, and this is a subtle change.

compiler/src/main/java/com/dylibso/chicory/compiler/internal/Compiler.java
chirino: Actually I don't think I need the replaceAll there at all.  it's already an internalClassName

compiler/src/main/java/com/dylibso/chicory/compiler/internal/Shaded.java
chirino: Also can't use the utility function since it would need to be shaded.

docs/docs/experimental/dwarf-debug-symbols.md.off
chirino: +1

docs/docs/experimental/dwarf-debug-symbols.md.off
chirino: ok. lets drop that line.

docs/tests/index.test.ts
chirino: Actually this helps the user figure out which page is the one being tested.  Before this change if the docs failed, I did not have a good way to figure out which one failed.

dwarf-rust-experimental/rust/src/count_vowels.wasm.json
chirino: yeah we can likely drop it.

runtime/src/main/java/com/dylibso/chicory/runtime/InterpreterMachine.java
chirino: The sad thing is that dropping that would not simplify this code your commenting on here since we still need to avoid replacing user code frames that could be sandwiched between the interpreter frames.

chirino:
> Thanks for the great step forward!
> 
> I think this is great progress, but we need to refine a few aspects:
> 
> * unless there is a strong counter motivation the API to enable debugging should be the same across `Machines`

I might have a better idea on how to do this.  Maybe allow configuring the withDebugParser on the interpreter machine too.

> * There is still a lot of code "around", hard-coded in the compiler/interpreter that does a lot of assumptions on Stratum, lines of code etc.

Not sure I understand what you mean by this point.  Could you point me to some examples?

> * `Stratum` still brings behavior + data while, ideally, we should have:
>   
>   * a "data-only" API in Chicory
>   * injectable behavior completely defined outside
 
 I tried to make the Stratum interface as narrow as possible.   I'm open to suggestions on how to improve.

> To get things easier I propose that we don't support(for now) "mixed" instances (compiled + interpreter) so that we can focus on getting the APIs right with less code around. Let me know your thoughts @chirino and thanks again!

I could move this to follow up PR.

compiler/src/main/java/com/dylibso/chicory/compiler/internal/Compiler.java
chirino: I found the value by looking at what was throwing the exception:

https://gitlab.ow2.org/asm/asm/-/blob/master/asm/src/main/java/org/objectweb/asm/ByteVector.java#L254

compiler/src/main/java/com/dylibso/chicory/compiler/internal/Compiler.java
chirino: Why enable it at build time if your going to disable it a runtime?  Compilers don't usually work that way.  Either you compiled in debug info and take the runtime hit, or you don't.

andreaTP: The sweet spot would be if we can compile the debugging options in and turn them off by default.
Hard to take a decision without real world data though.

andreaTP:
> I might have a better idea on how to do this. Maybe allow configuring the withDebugParser on the interpreter machine too.

Happy to have a look at a proposal!

> Not sure I understand what you mean by this point. Could you point me to some examples?

Yes, for example `Stratum.optimizeForLookups` is used in various places, but what is supposed to do is completely obscure and delegated to the real implementation. This is clearly a leaky abstraction.

> I tried to make the Stratum interface as narrow as possible. I'm open to suggestions on how to improve.

Honestly, I have the feeling that continuing to go forth and back on this huge implementation is going to be a long and tedious activity for us two, additionally, the number of comments is starting to hurt in readability.
This is a great demonstration of the functionality but it is far from being in a state where it can be merged.

My proposal is to close or convert this PR back to Draft and take out small and consistent pieces one at the time as separate PRs @chirino .

chirino:
How do you feel about using this config to enable the debug parser for the interpreter?

```             
                Instance.builder(loadModule("compiled/count_vowels.rs.wasm"))
                        .withMachineFactory(
                                InterpreterMachine.builder()
                                        .withDebugParser(DebugParser::parse)
                                        .build()
                        )
                        .build();
                        
```

It would look much more like what we are doing for the compiler:

```
                Instance.builder(module)
                        .withMachineFactory(
                                MachineFactoryCompiler.builder(module)
                                        .withDebugParser(DebugParser::parse)
                                        .compile()
                        )
                        .build();
```

chirino:
> My proposal is to close or convert this PR back to Draft and take out small and consistent pieces one at the time as separate PRs @chirino .

The problem is the PR requires many different modules to work together in concert and you cant really test or evaluate the suitability of one part without the rest to get the big picture.

andreaTP:
> The problem is the PR requires many different modules to work together in concert and you cant really test or evaluate the suitability of one part without the rest to get the big picture.

How we did it in the past it to keep a Draft PR open showing the direction, and merging smaller PRs to progressively reduce the diff while we agree on the steps.

chirino:
> How do you feel about using this config to enable the debug parser for the interpreter?
> 
> ```
>                 Instance.builder(loadModule("compiled/count_vowels.rs.wasm"))
>                         .withMachineFactory(
>                                 InterpreterMachine.builder()
>                                         .withDebugParser(DebugParser::parse)
>                                         .build()
>                         )
>                         .build();
>                         
> ```
> 
> It would look much more like what we are doing for the compiler:
> 
> ```
>                 Instance.builder(module)
>                         .withMachineFactory(
>                                 MachineFactoryCompiler.builder(module)
>                                         .withDebugParser(DebugParser::parse)
>                                         .compile()
>                         )
>                         .build();
> ```

Just tried this.. it won't work to the case where your doing runtime compile with some functions interpreted.  As the Compiled code just passes the Instance to create the Interpreter and you don't have debugParser anymore.  Setting it on the instance solved that problem.

chirino:
PR converted to draft.

dwarf-rust-experimental/rust/src/count_vowels.wasm.json
evacchi, line 1: I think this file was checked in by mistake, because it's huge and it accounts for like >75% of this PR's line count 😄

docs/docs/experimental/dwarf-debug-symbols.md.off
```diff
@@ -0,0 +1,157 @@
+---
```
evacchi, line 1: > [‎docs/docs/experimental/dwarf-debug-symbols.md.off](https://github.com/dylibso/chicory/pull/943/files?#diff-4b126a6214918d8ef42e07bd9ca1327db6fe23f4ec1445e9d976591295978b05)

"off" ?

runtime/src/main/java/com/dylibso/chicory/runtime/DebugParser.java
```diff
@@ -0,0 +1,6 @@
+package com.dylibso.chicory.runtime;
+
+import com.dylibso.chicory.wasm.WasmModule;
+import java.util.function.Function;
+
+public interface DebugParser extends Function<WasmModule, Stratum> {}
```
evacchi, line 6: nit: is it useful to extend `Function`? Do we ever use this as a `Function`? otherwise we can just define a plain interface -- in fact, `Function<T,U>` is the pattern the codebase adopts for factories; I would personally prefer properly named factories in the codebase, but we should probably be consistent

evacchi (commented):
I left a few comments (mostly nits). There is just one huge file that makes it look like this PR is bigger than it actually is

compiler/src/main/java/com/dylibso/chicory/compiler/internal/Compiler.java
evacchi: more than this being public, I am not sure why you need the counter, it seems like it's only incremented when `build()` is invoked repeatedly on the same module (?).. do we ever need to do that _and not_ overwrite the output?

docs/docs/experimental/dwarf-debug-symbols.md.off
```diff
@@ -0,0 +1,157 @@
...
+import com.dylibso.chicory.runtime.Instance;
+import com.dylibso.chicory.wasm.WasmModule;
+
+var module = Parser.parse(new File("your.wasm"));
+Instance instance = Instance.builder(module)
+        .withMachineFactory(
+                MachineFactoryCompiler.builder(module)
+                        .withDebugParser(DebugParser::parse)
+                        .compile())
+        .build();
```
evacchi, line 119: shouldn't this be both on the Instance.builder _and_ the machine factory (I've seen this in the test cases)

speaking of which, what if, as a contract, if the Machine factory supports debug symbols, it just implicitly pulls them off the Instance it's being given, allowing us to avoid being explicit here?


```
Instance instance = Instance.builder(module)
   .withDebugParser(DebugParser::parse)
   .withMachineFactory(
                MachineFactoryCompiler.builder(module)
                        // no longer needed: .withDebugParser(DebugParser::parse)
                        .compile())
        .build();
```

dwarf-rust-experimental/rust/LICENSE
```diff
@@ -0,0 +1,29 @@
+// Copyright 2019 The Chromium Authors
```
evacchi, line 1: 🤔

dwarf-rust-experimental/rust/src/main.rs
```diff
@@ -0,0 +1,502 @@
...
+      // }
+      // if let FuncState::Ignored = func_state {
+      //   if row.end_sequence() {
+      //     func_state = FuncState::Start;
+      //   }
+      //   continue;
+      // }
+      // if row.end_sequence() {
+      //   func_state = FuncState::Start;
+      // }
```
evacchi: leftovers?

dwarf-rust-experimental/rust/src/main.rs
```diff
@@ -0,0 +1,502 @@
...
+      };
+
+      let pos = {
+        let mut line = match row.line() {
+          Some(line) => line.checked_sub(1).unwrap().try_into().unwrap(),
+          None => continue, // couldn't attribute instruction to any line
+        };
+
+        // It seems we need to add 1 to the line numbers for Rust.
+        let is_rust = lang == constants::DW_LANG_Rust.0;
```
evacchi, line 208: is this intended to be Rust-specific? are we going to provider parsers for other languages? how does it work for Go, C etc...

runtime/src/main/java/com/dylibso/chicory/runtime/InterpreterMachine.java
```diff
@@ -27,10 +33,26 @@ public class InterpreterMachine implements Machine {
...
     private final Instance instance;
 
+    private final Stratum stratum;
+
     public InterpreterMachine(Instance instance) {
         this.instance = instance;
         stack = new MStack();
         this.callStack = new ArrayDeque<>();
+
+        CustomSection smapSection = instance.module().customSection("SMAP");
```
evacchi, line 47: - maybe let's add a few notes about the convention we are exploiting here, I see that this custom section SMAP is being inserted by us in Generator.java;
- this also means that the module is a "meta-module", that is it won't contain any code; 
- thus, it doesn't even make a lot of sense for it to be interpreted in the first place

build-time-compiler/src/main/java/com/dylibso/chicory/build/time/compiler/Generator.java
```diff
@@ -151,6 +180,14 @@ public void generateMetaWasm(Set<Integer> interpretedFunctions) throws IOExcepti
...
                         }
                         writer.writeSection(SectionId.CODE, out.toByteArray());
+                        if (!outputStratum.isEmpty()) {
+                            var smap = Stratum.toSMapString("interpreted.wasm", outputStratum);
+                            writer.writeSection(
+                                    UnknownCustomSection.builder()
+                                            .withName("SMAP")
+                                            .withBytes(smap.getBytes(StandardCharsets.UTF_8))
+                                            .build());
+                        }
```
evacchi, line 190: maybe let's add a comment describing what's going on here.


also, since this is a text file we could also just save it as a resource and load it as such at run-time

runtime/src/main/java/com/dylibso/chicory/runtime/Stratum.java
```diff
@@ -0,0 +1,365 @@
...
+import java.util.Objects;
+
+/**
+ * Stratum represents a mapping of source code lines to function names and addresses.
+ * <p>
+ * It is used to manage the relationship between input source files and their corresponding output lines,
+ * as well as to provide a way to retrieve function mappings based on line numbers.
+ * </p>
+ */
+public final class Stratum {
```
evacchi, line 18: it seems like the public interface for Stratum is very thin, after all:

```
    boolean isEmpty()
    String getFunctionMapping(int)
    Line getInputLine(int)
    LineMapping getLineMapping(int)
```

maybe we can externalize further the remaining behavior and keep just an interface with the methods above for the core

docs/docs/experimental/dwarf-debug-symbols.md.off
chirino: In progress docs,  didn't want them getting published yet.

runtime/src/main/java/com/dylibso/chicory/runtime/DebugParser.java
chirino: yes so we can do a `ServiceLoader.load(DebugParser.class)` in the ChicoryCompilerGenMojo class.

docs/docs/experimental/dwarf-debug-symbols.md.off
chirino: Kinda..  It's only needed if there are some interpreted functions.  But for completeness yes.

Letting the Machine factory pull the debug symbols from the instance would mean that we have to defer compiling the .class files until the `MachineFactory.appy(Instance )` gets called.  That would be a big change to how that works.  Not sure if we could continue supporting all the runtime compile use cases we currently do.  Like compiling once and using that for multiple instances.

dwarf-rust-experimental/rust/LICENSE
chirino: The sources in the dir originate from https://github.com/bmeurer/wasm-source-map

dwarf-rust-experimental/rust/src/main.rs
chirino: yeah.

dwarf-rust-experimental/rust/src/main.rs
chirino: In theory it is not rust specific.  We just need more tests for other languages to verify.

runtime/src/main/java/com/dylibso/chicory/runtime/InterpreterMachine.java
chirino: I'll add a comment to this section to clarify:

        // The SMAP custom section only gets added to WASM modules that are created by
        // the build time compiler (see Generator.java).
        // The module is meta-module and will only contain code for functions which
        // will be interpreted (typically because they were too large to compile to bytecode).

build-time-compiler/src/main/java/com/dylibso/chicory/build/time/compiler/Generator.java
chirino: So this eventually needs to be loaded by the CompilerInterpreterMachine.  So it would need to know at least the class name to figure out the right resource name.  I was jus trying to avoid changing that interface.

runtime/src/main/java/com/dylibso/chicory/runtime/Stratum.java
chirino: Do we envision needing more than one implementation of it? 

BTW. `LineMapping getLineMapping(int)` is private.

PR #946, opened 2025-06-13
feat: Add deep copy functionality to Instance and related runtime cla…

chirino:
…sses

- Add copy() method to Instance class for creating deep copies of WebAssembly runtime state
- Implement copy() methods in GlobalInstance, TableInstance, and TagInstance classes
- Add copy() interface method to Memory with implementations in ByteArrayMemory and ByteBufferMemory
- Extend ImportValues with copy() method for deep copying import state
- Create comprehensive InstanceCopyTest with test cases for all copy scenarios
- Add global-counter.wat test module with compiled WASM binary for testing global state copying
- Update wasm-corpus Dockerfile to include wabt tools for WebAssembly compilation
- Add copyWithOptions() method that allows you to change the listen or imports for the new instance copy.

This implementation enables creating independent copies of WebAssembly runtime instances with fully isolated state, including memory contents, global variables, tables, tags, and import values.  This is essential for scenarios requiring instance isolation, snapshotting, or parallel execution contexts.

### DIFF
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/Instance.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/Instance.java
@@ -10,6 +10,7 @@ import static com.dylibso.chicory.wasm.types.ExternalType.TAG;
 import static java.util.Objects.requireNonNullElse;
 import static java.util.Objects.requireNonNullElseGet;
 
+import com.dylibso.chicory.runtime.internal.ExperimentalCopier;
 import com.dylibso.chicory.wasm.ChicoryException;
 import com.dylibso.chicory.wasm.InvalidException;
 import com.dylibso.chicory.wasm.UninstantiableException;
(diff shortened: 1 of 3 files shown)


runtime/src/main/java/com/dylibso/chicory/runtime/GlobalInstance.java
```diff
@@ -88,4 +88,13 @@ public void setInstance(Instance instance) {
     public MutabilityType getMutabilityType() {
         return mutabilityType;
     }
+
+    public GlobalInstance copy() {
+        GlobalInstance copy =
+                new GlobalInstance(
+                        this.valueLow, this.valueHigh, this.valType, this.mutabilityType);
+        // The instance field will be set by the calling code
```
andreaTP, line 96: The need of resetting/recomputing fields shows that this is a leaky abstraction.

andreaTP:
For reference, wasmtime experimented with those things:
https://bytecodealliance.org/articles/wasmtime-10-performance

PR #957, opened 2025-06-25
[proposals] Threads

andreaTP:
Opening for visibility, just implemented a good chunk of validation, the opcodes are still missing.

Notes:
- working in the interpreter now
- runs `jq` and `protoc` from `wasilibs`
- no implementation of `atomic.fence` as it's completely untested
- the testsuite is really poor


wasi-tests/src/test/java/com/dylibso/chicory/wasi/WasiPreview1Test.java
```diff
@@ -443,4 +446,53 @@ private static WasiPreview1 wasiWithDirectory(String guest, Path host) {
...
+                                                new ByteArrayMemory(
+                                                        new MemoryLimits(
+                                                                /* 7 */ 4,
+                                                                MemoryLimits.MAX_PAGES,
+                                                                true))))
+                                .build();
+                var module =
+                        Parser.parse(
+                                new File(
+                                        "/home/andreatp/workspace/go-protoc-gen-grpc-java/internal/wasm/protoc-gen-grpc-java.wasm"));
```
andreaTP: My goal was to run this module, coming from:
https://github.com/wasilibs/go-protoc-gen-grpc-java

But something is not working yet :cry:

andreaTP: Working now! Both protoc and jq

andreaTP:
This is ready for review, is the full interpreter implementation.
I'm already working on porting to the compiler :slightly_smiling_face:

chirino:
given atomic fence is not implemented.  What happens right now if a wasm module uses that instruction?  I hope we give the user a decent error message that the wasm module is using a unsupported instruction?

runtime/src/main/java/com/dylibso/chicory/runtime/InterpreterMachine.java
```diff
@@ -1995,6 +2165,321 @@ private static void LOCAL_TEE(MStack stack, Operands operands, StackFrame curren
...
+            throw new InvalidException("unaligned atomic");
+        }
+        synchronized (instance.memory().lock(ptr)) {
+            instance.memory().writeI32(ptr, value);
+        }
+    }
+
+    private static void I64_ATOMIC_STORE8(MStack stack, Instance instance, Operands operands) {
+        var value = (byte) stack.pop();
+        var ptr = (int) (operands.get(1) + (int) stack.pop());
```
chirino: Should this be be `var ptr = readMemPtr(stack, operands);`  instead to be consistent with the others?

runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
```diff
@@ -58,6 +61,133 @@ public ByteArrayMemory(MemoryLimits limits, MemAllocStrategy allocStrategy) {
...
+            }
+        }
+    }
+
+    // Notify all waiters at this address
+    @Override
+    public int notify(int address, int maxThreads) {
+        if (!shared()) {
+            return 0;
+        }
```
chirino, line 180: Should we be doing similar checks or maybe throwing a trap in the other methods just in case they get called in non-shared memory?

runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
```diff
@@ -58,6 +61,133 @@ public ByteArrayMemory(MemoryLimits limits, MemAllocStrategy allocStrategy) {
...
+                        return new AtomicInteger(1);
+                    } else {
+                        v.incrementAndGet();
+                        return v;
+                    }
+                });
+    }
+
+    private int waitOnMonitor(int address, long timeout) {
+        long elapsed = System.nanoTime() + timeout;
```
chirino: this variable name is misleading, I think it should be something like endTime.

runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
```diff
@@ -58,6 +61,133 @@ public ByteArrayMemory(MemoryLimits limits, MemAllocStrategy allocStrategy) {
...
+        try {
+            while (!notifyInProgress.containsKey(address) // prevents spurious wakeup
+                    && System.nanoTime() < elapsed) {
+                var waitTime = elapsed - System.nanoTime();
+                long millis = Math.max(waitTime / 1_000_000L, 0);
+                int nanos = Math.max((int) (waitTime % 1_000_000L), 0);
+                monitors.get(address).wait(millis, nanos);
+            }
```
chirino, line 109: I think this is thread safe because the caller to this method has done a `synchronized (monitor) {` but it's not super obvious, maybe a comment would  help.   Also is the `monitors.get(address)` giving you the same monitor that the caller looked up and synchronized on?  Could we avoid doing another lookup by passing it as function arg?

runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
```diff
@@ -58,6 +61,133 @@ public ByteArrayMemory(MemoryLimits limits, MemAllocStrategy allocStrategy) {
...
+    }
+
+    private void endWaitOn(int address) {
+        if (notifyInProgress.containsKey(address)
+                && notifyInProgress.get(address).decrementAndGet() == 0) {
+            notifyInProgress.remove(address);
+        }
+        if (monitors.containsKey(address) && monitors.get(address).decrementAndGet() == 0) {
+            monitors.remove(address);
+        }
```
chirino, line 129: how about something like this..  should reduce the number of map lookups done.

```
        AtomicInteger notifyCount = notifyInProgress.get(address);
        if (notifyCount != null && notifyCount.decrementAndGet() == 0) {
            notifyInProgress.remove(address);
        }

        AtomicInteger monitor = monitors.get(address);
        if (monitor != null && monitor.decrementAndGet() == 0) {
            monitors.remove(address);
        }
```

runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
```diff
@@ -58,6 +61,133 @@ public ByteArrayMemory(MemoryLimits limits, MemAllocStrategy allocStrategy) {
...
+        try {
+            while (!notifyInProgress.containsKey(address) // prevents spurious wakeup
+                    && System.nanoTime() < elapsed) {
+                var waitTime = elapsed - System.nanoTime();
+                long millis = Math.max(waitTime / 1_000_000L, 0);
+                int nanos = Math.max((int) (waitTime % 1_000_000L), 0);
+                monitors.get(address).wait(millis, nanos);
+            }
+        } catch (InterruptedException ie) {
+            throw new ChicoryInterruptedException("Thread interrupted");
```
chirino, line 112: Maybe added `Thread.currentThread().interrupt(); // Restore interrupt status`

runtime/src/main/java/com/dylibso/chicory/runtime/ByteBufferMemory.java
```diff
@@ -44,6 +47,133 @@ public ByteBufferMemory(MemoryLimits limits, MemAllocStrategy allocStrategy) {
...
+        try {
+            while (!notifyInProgress.containsKey(address) // prevents spurious wakeup
+                    && System.nanoTime() < elapsed) {
+                var waitTime = elapsed - System.nanoTime();
+                long millis = Math.max(waitTime / 1_000_000L, 0);
+                int nanos = Math.max((int) (waitTime % 1_000_000L), 0);
+                monitors.get(address).wait(millis, nanos);
+            }
+        } catch (InterruptedException ie) {
+            throw new ChicoryInterruptedException("Thread interrupted");
```
chirino, line 98: Maybe added `Thread.currentThread().interrupt(); // Restore interrupt status`

andreaTP:
> given atomic fence is not implemented. What happens right now if a wasm module uses that instruction? I hope we give the user a decent error message that the wasm module is using a unsupported instruction?

[This](https://github.com/dylibso/chicory/blob/cd67b24a402c607892752cb3f587bbae0f3c7566/wasm/src/main/java/com/dylibso/chicory/wasm/Validator.java#L1889-L1890) will be the first error that shows up to a user, we can improve the text if you think that is not enough(something like: "open an issue at ...").
So far I haven't seen any usage of it, and the spec is a bit too vague to try to anticipate a possible implementation without examples and tests ... I left it out on purpose.

runtime/src/main/java/com/dylibso/chicory/runtime/InterpreterMachine.java
andreaTP: good catch :+1:

runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
andreaTP: The behavior is fully specified, but there are no tests at all in the testsuite around it, it will probably come up when we implement multi memory support ...

What I can do for now is to go through the spec again and review the behavior at best of my understanding of it.

runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
andreaTP: make sense, good suggestion :+1:

runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
andreaTP: > but it's not super obvious

agree, I'll add a comment

> Could we avoid doing another lookup by passing it as function arg?

Yes :+1:

runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
andreaTP: I'll change it :+1:

runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
andreaTP: Make sense :+1:

andreaTP:
I'll go through the review comments now, I added the implementation in the compiler as a single commit, is just a lot of very repetitive code, but I can take it out for a follow up PR when requested.

runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
andreaTP: I reviewed the spec and the code, the "atomic" opcodes acting on a non-shared memory should, usually, behave as the corresponding non atomic operation.
`wait` instead should trap, I've added the relevant code, but I don't have a great feeling as there are no tests nor examples that can be used to drive the implementation.

I think this is enough as a starting point, and we can iterate when we receive a bug/issue.

andreaTP:
Ready for the next review round!

- addressed all of the comments (thanks @chirino :pray: )
- added the full implementation in the compiler
- added all the relevant tests

compiler/src/main/java/com/dylibso/chicory/compiler/internal/Compiler.java
```diff
@@ -561,7 +562,7 @@ private static void emitFunction(
                         null);
 
         // uncomment if you ever want to troubleshoot invalid bytecode generation
-        // methodWriter = new CheckMethodAdapter(methodWriter);
+        methodWriter = new CheckMethodAdapter(methodWriter);
```
chirino: I assume you uncommented this to aid development.

chirino (approved):
just one small nit but...  LGTM!

compiler/src/main/java/com/dylibso/chicory/compiler/internal/Compiler.java
andreaTP: Yes, that's a day saver, thanks for it!
Good catch!

PR #972, opened 2025-07-07
Test swift

andreaTP:
Opening in draft as the resulting payloads are a bit big to be committed.

I'd file a couple of issues upstream and we move from there.


andreaTP:
Ready for review, I updated the compilation of Swift to use 6.2 snapshot and targeting embedded.
More or less following the examples in: https://github.com/swiftlang/swift-for-wasm-examples

Now the generated payloads are definitely small enough!

evacchi (approved):

PR #977, opened 2025-07-10
Fix flaky threads test on java 11 and win/osx

andreaTP:
I noticed a couple of unrelated failures especially on Win + Java 11.

This is a simpler but apparently more solid setup:
https://github.com/andreaTP/chicory/actions/runs/16200043900


chirino (approved):
lgtm.

PR #993, opened 2025-07-29
add Android-compatible null stream implementations and logger fixes

itsvks19:
- added `IO` utility class for `nullInputStream` and `nullOutputStream` to support Android devices below API 33.

and also a simple fix for this error:
```
java.lang.NoSuchMethodError
Message: No static method getLogger(Ljava/lang/String;)Ljava/lang/System$Logger; in class Ljava/lang/System; or its super classes (declaration of 'java.lang.System' appears in /apex/com.android.art/javalib/core-oj.jar)
```

### DIFF
--- a/log/src/main/java/com/dylibso/chicory/log/SystemLogger.java
+++ b/log/src/main/java/com/dylibso/chicory/log/SystemLogger.java
@@ -1,7 +1,6 @@
 package com.dylibso.chicory.log;
 
 public class SystemLogger implements Logger {
-    // FIX: System.getLogger() is not available on Android.
     private static final System.Logger LOGGER = System.getLogger("chicory");
 
     @Override
--- a/log/src/main/java/com/dylibso/chicory/log/SystemLogger.java
+++ b/log/src/main/java/com/dylibso/chicory/log/SystemLogger.java
@@ -1,6 +1,7 @@
 package com.dylibso.chicory.log;
 
 public class SystemLogger implements Logger {
+    // FIX: System.getLogger() is not available on Android.
     private static final System.Logger LOGGER = System.getLogger("chicory");
 
     @Override
--- a/wasi/src/main/java/com/dylibso/chicory/wasi/IO.java
+++ b/wasi/src/main/java/com/dylibso/chicory/wasi/IO.java
@@ -1,0 +1,145 @@
+package com.dylibso.chicory.wasi;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.Objects;
+
+/**
+ * Provides implementations of null stream utilities that are compatible with Android
+ * devices running API levels below 33. This class contains implementations equivalent to
+ * {@code OutputStream.nullOutputStream()} and {@code InputStream.nullInputStream()}
+ * which were introduced in Android API Level 33.
+ */
+class IO {
+    private IO() {}
+
+    /**
+     * Returns an {@link OutputStream} that discards all bytes written to it.
+     * This implementation provides compatibility for Android devices running below API Level 33,
+     * where {@code OutputStream.nullOutputStream()} is not available.
+     * <p>
+     * This implementation is copied from {@code OutputStream.nullOutputStream()}.
+     *
+     * @return an output stream that discards all bytes written to it
+     */
+    public static OutputStream nullOutputStream() {
+        return new OutputStream() {
+            private volatile boolean closed;
+
+            private void ensureOpen() throws IOException {
+                if (closed) {
+                    throw new IOException("Stream closed");
+                }
+            }
+
+            @Override
+            public void write(int b) throws IOException {
+                ensureOpen();
+            }
+
+            @Override
+            public void write(byte[] b, int off, int len) throws IOException {
+                Objects.checkFromIndexSize(off, len, b.length);
+                ensureOpen();
+            }
+
+            @Override
+            public void close() {
+                closed = true;
+            }
+        };
+    }
+
+    /**
+     * Returns an {@link InputStream} that contains no bytes.
+     * This implementation provides compatibility for Android devices running below API Level 33,
+     * where {@code InputStream.nullInputStream()} is not available.
+     * <p>
+     * This implementation is copied from {@code InputStream.nullInputStream()}.
+     *
+     * @return an input stream that contains no bytes
+     */
+    public static InputStream nullInputStream() {
+        return new InputStream() {
+            private volatile boolean closed;
+
+            private void ensureOpen() throws IOException {
+                if (closed) {
+                    throw new IOException("Stream closed");
+                }
+            }
+
+            @Override
+            public int available() throws IOException {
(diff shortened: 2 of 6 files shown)


andreaTP (commented):
Thanks for the PR @itsvks19 !
I'm on vacation now, will take a look next week if you don't have any specific time requirement.

This is a good candidate for @evacchi , @yigit or @illarionov if they have a cycle to spare on the review.

I don't mind about full test coverage here but in the mid term we should probably plan to integrate wasi tests on Android.

itsvks19:
why every time checks are failing?

log/src/main/java/com/dylibso/chicory/log/SystemLogger.java
```diff
@@ -1,6 +1,7 @@
 package com.dylibso.chicory.log;
 
 public class SystemLogger implements Logger {
+    // FIX: System.getLogger() is not available on Android.
```
andreaTP: Outdated comment?

wasi/src/main/java/com/dylibso/chicory/wasi/IO.java
```diff
@@ -0,0 +1,145 @@
...
+ */
+final class IO {
+    private IO() {}
+
+    /**
+     * Returns an {@link OutputStream} that discards all bytes written to it.
+     * This implementation provides compatibility for Android devices running below API Level 33,
+     * where {@code OutputStream.nullOutputStream()} is not available.
+     * <p>
+     * This implementation is copied from {@code OutputStream.nullOutputStream()}.
```
andreaTP: At least, this comment line should be removed ( [rationale](https://github.com/scala-js/scala-js/blob/d6d6a3778ab3b58b477a619ad341f72c7425d5ee/JAVALIB.md?plain=1#L7-L11) ).

The implementation is straightforward enough to be acceptable anyhow in my opinion.

wasi/src/main/java/com/dylibso/chicory/wasi/IO.java
```diff
@@ -0,0 +1,145 @@
...
+            }
+        };
+    }
+
+    /**
+     * Returns an {@link InputStream} that contains no bytes.
+     * This implementation provides compatibility for Android devices running below API Level 33,
+     * where {@code InputStream.nullInputStream()} is not available.
+     * <p>
+     * This implementation is copied from {@code InputStream.nullInputStream()}.
```
andreaTP: same as above.

wasi/src/main/java/com/dylibso/chicory/wasi/IO.java
```diff
@@ -0,0 +1,145 @@
...
+                return new byte[0];
+            }
+
+            @Override
+            public long skip(long n) throws IOException {
+                ensureOpen();
+                return 0L;
+            }
+
+            /* This method was added in Java 12+ */
```
andreaTP: Any issues in leaving the method uncommented?

wasi/src/main/java/com/dylibso/chicory/wasi/WasiPreview1.java
```diff
@@ -124,8 +124,23 @@ public Builder withOptions(WasiOptions opts) {
             return this;
         }
 
+        private boolean isAndroid() {
```
andreaTP: Also `static` as it doesn't depend on the context.
Should we add an "hidden" specific System Property to enable overriding this behavior as an escape hatch?
e.g.:
if `System.getProperty("chicory.onAndroid")` is defined the value will be honored.

wasi/src/main/java/com/dylibso/chicory/wasi/WasiPreview1.java
```diff
@@ -124,8 +124,23 @@ public Builder withOptions(WasiOptions opts) {
...
+                String runtime = System.getProperty("java.runtime.name");
+                return runtime != null && runtime.toLowerCase(Locale.ENGLISH).contains("android");
+            }
+        }
+
         public WasiPreview1 build() {
-            if (logger == null) {
+            if (logger == null && isAndroid()) {
+                // Android requires explicitly setting a logger since System.Logger is not available
+                throw new WasmRuntimeException(
```
andreaTP: for the future: what would be a sensible default on Android instead of failing?
Is there a leading logging framework that we should integrate with?

wasi/src/main/java/com/dylibso/chicory/wasi/IO.java
itsvks19: Okay

wasi/src/main/java/com/dylibso/chicory/wasi/IO.java
itsvks19: Yeah. It will give an compile time error.

wasi/src/main/java/com/dylibso/chicory/wasi/WasiPreview1.java
itsvks19: You can use `System.out` as android logger can't be used in a Java library (you need to create a android library).

wasi/src/main/java/com/dylibso/chicory/wasi/WasiPreview1.java
itsvks19: I don't know what should we do. You can add a system property if you want. But current implementation is good enough for detecting android.

wasi/src/main/java/com/dylibso/chicory/wasi/IO.java
andreaTP: Is it safe to just remove it?

itsvks19: Yes. You can remove it.

wasi/src/main/java/com/dylibso/chicory/wasi/WasiPreview1.java
andreaTP: Should we provide a `BasicLogger` to automatically do the fallback?

@evacchi what's your take on this?

itsvks19: > Should we provide a `BasicLogger` to automatically do the fallback?

We can provide a `BasicLogger`.

evacchi: yeah I agree, I think we can also use java.util.logging.Logger if we don't want to use sysout directly

andreaTP:
@itsvks19 just to make sure we are aligned,
As a general expectation, when we agree on some changes, I'd wait for you to push the changes.

If, instead, you prefer me to take over this PR just let me know :slightly_smiling_face:

itsvks19:
@andreaTP No, I will push my changes. I was busy.

andreaTP:
Cool, no rush!

itsvks19:
I think everything resolved now. `java.util.logging.Logger` works in android.

PR #1010, opened 2025-08-29
chore: remove proposals remapping from test gen

andreaTP:
I'm starting working on a new proposal, this remapping mechanism was useful when using `wabt` but it's not anymore.
Extracted to reduce the diff of upcoming PRs.


chirino (approved):
lgtm

PR #1027, opened 2025-09-15
sanitize methods and field names in annotation processor

### LINKED ISSUE #1022: Invalid code when an export contains a dot
The annotation processor is emitting invalid code when an export contains a dot `.`.
We should mangle the field and methods names to avoid this issue.

andreaTP:
Fix #1022


chirino (approved):
lgtm

Issue #1012, opened 2025-08-30
Add Java module info


A248:
The chicory dependency is currently lacking module info, e.g. `module-info.java`. This makes it difficult or impossible to use it in a modular application, since the filename-derived module name is unstable. Chicory should ideally provide this information which is easy to add. If Chicory is still targeting Java 8, then at least `Automatic-Module-Name` can be added.

Just to be fully clear: the terminology "module" is overloaded. I'm not referring to WASM modules. Rather, Java has had a module system since 2017.

Modules are a higher-level construct than packages. Check out this guide if you're unfamiliar: https://dev.java/learn/modules/intro/

andreaTP:
Hi @A248 and thanks for taking the time to file this issue!

Makes sense to me but I haven't gone through the process of using module-info myself(or any other automatic inference of it).
At the moment we support Java 11+.

Would you be up in sending over a PR to improve this aspect? It doesn't need to cover the entire codebase as a starting point.
Alternatively do you have links or resources you suggest to look up?

A248:
I completed a brief check of the project structure. It looks to me like everything is compatible with Java modules, since Chigory uses a clean project layout with no overlapping packages between artifacts. I can hopefully get started on a PR over the next couple days. Inshallah

andreaTP:
@A248 polite ping, any update?

andreaTP:
Fixed in #1028

PR #1028, opened 2025-09-15
Add Java module definitions

A248:
Aims to solve #1012

I added Java module definitions where I thought appropriate. Several of the subprojects aren't designed for external consumption, such as test suites and benchmarks.

Packages named "internal" I treated as internal to chicory, i.e. not intended for external consumption.

Additionally, I did my best to check when types from a dependency were exposed in a module's public API, in order to use `requires transitive` appropriately. It's possible I missed a spot, but this will never be breaking and will result in at most a compiler warning (on chicory's side) or the need to write another `requires` (on the library user's side).

**Checkstyle**

Chicory was still using checkstyle 9.3, which is still the default version in the maven-checkstyle-plugin. Because this old checkstyle version was running into problems consuming `module-info.java`, I increased the version to the latest.

This required checking for breaking changes in checkstyle, which I did by looking at [every release since 9.3](https://checkstyle.org/releasenotes.html#Release_9.3). Checkstyle does not follow SemVer so you have to look at every release notes.

**Terminology**

For people reading this PR, the PR itself targets the Java Modules introduced in Java 9. They are a separate concept from WASM modules.


A248:
Theoretically, yes. Some behavior is different on the module path: Java's `ServiceLoader` and `ClassLoader.getResource(AsStream)` are both affected by switching from an unnamed jar to a full module.

in practice, compilation will mostly ensure that the modules are declared properly. But if you want to spin up an integration test, you can always create a subproject that runs Chicory on the module path. This subproject can be created either as a plain Maven module, or using the maven-invoker-plugin.

Theoretically speaking though, there's no way to reach 100% coverage unless you literally fork all test runs and run them in a modular environment. That's why I rely on `rg ServiceLoader` and `rg getResource` when checking a codebase for the ability to use modules (using rg = ripgrep).

andreaTP:
Thanks for the answer, I don't think we need 100% coverage, but an integration test to check that we are not badly messing up will be good...
I like the maven-invoker approach that's already used in this codebase, do you mind to add it to this PR?

andreaTP:
@dmlloyd IIRC you played a bit with the module system, do you want to comment on this PR? :pray:

pom.xml
```diff
@@ -554,6 +554,9 @@
           <excludeGeneratedSources>true</excludeGeneratedSources>
           <checkstyleRules>
             <module name="Checker">
+              <module name="BeforeExecutionExclusionFileFilter">
+                <property name="fileNamePattern" value="module\-info\.java$"/>
+              </module>
```
dmlloyd, line 493: Why exactly is checkstyle failing on `module-info`?

pom.xml
```diff
@@ -636,6 +646,7 @@
           <compilerArgs>
             <arg>-XDcompilePolicy=simple</arg>
             <arg>--should-stop=ifError=FLOW</arg>
+            <arg>-Xlint:-module</arg>
```
dmlloyd: Why are module warnings disabled, specifically?

pom.xml
```diff
@@ -741,6 +752,7 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-surefire-plugin</artifactId>
         <configuration>
+          <useModulePath>false</useModulePath>
```
dmlloyd: I'd be concerned about not having a surefire execution that uses the module path. I think if you're adding module descriptors, you really do need to test them. This could be done by configuring common properties in `pluginManagement` and then setting up two executions in `build`, one for module mode and one for classpath mode.

pom.xml
A248: In order to make "internal" packages stay internal to Chicory, I added qualified exports like `exports ....internal to othermodule;`.

However, this causes a warning message like "module not found". This happens because the downstream module is not present during module compilation. To work around this, we'd need to either disable the lint -- or to restructure Chicory in ways that are very unstandard for a Maven build ([1](https://stackoverflow.com/questions/67795589/how-to-compile-with-javac-modules-with-exports-to-directives-without-any))

Because of `<failOnWarnings>`, the module-related warning message was failing the build, so I disabled itt.

pom.xml
A248: module-info is wholly unsupported by checkstyle, so this is the recommended configuration (per their docs) to skip it.

https://github.com/checkstyle/checkstyle/issues/8240

dmlloyd: Makes sense. I'd put a comment in here to that effect though.

pom.xml
dmlloyd: I'd add a comment here too.

A248:
Good news! Not only did the tests caught a couple bugs, but they showed that the "internal" packages need to be exported after al. This obviates the need for `-Xlint:-module` and makes the PR overall more clean.

All that considered, I don't have any other changes in mind to make to this PR, besides addressing your review concerns, of course.

compiler/pom.xml
```diff
@@ -62,4 +62,23 @@
...
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>default-testCompile</id>
+            <configuration>
+              <compilerArgs>
+                <arg>--add-modules</arg>
+                <arg>java.sql</arg>
```
andreaTP, line 76: this is surprising

wasm-corpus/pom.xml
```diff
@@ -19,6 +19,17 @@
 
   <build>
     <plugins>
+      <plugin>
```
andreaTP, line 22: `wasm-corpus` is not, and should not be, a published artifact, do we need to comply with the module system anyway?

dmlloyd: In module mode, all dependencies also have to be modules. In this case, nothing is published, but the JAR that is built minimally must have a `Automatic-Module-Name` in order to have classes that are imported by tests.

I believe the JAR would already have been built even before this change. Publication would be handled by another mojo.

andreaTP: thanks

andreaTP:
I added a commit to disable checkstyle on Java 11, but now there is an issue with the modules.

Just to make sure I ask it: are those changes fully backward compatible?
E.g.: no changes are expected in user projects/code if they are not using the module system, correct?

andreaTP:
Just to spell it out, I was waiting an iteration from you @A248 on this PR, let me know if you are going to have time to progress on it :slightly_smiling_face: and thanks again :pray:

A248:
Hi there, sorry for the delay. I took a look at the tests which were strangely failing on Java 11.

What's weird is that the approvaltests dependency jar contains `.java` source files alongside the classfiles. Javac is compiling these source files during test compilation. Why this only happens on Java 11 on the module path, but not in later Java versions, is beyond me.

Anyway, I'll hopefully get to the bottom of this and update this PR with the fix.

A248:
Potentially related: https://bugs.openjdk.org/browse/JDK-8338675 (javac modifies jar files when .java files are present inside them!)

andreaTP:
@dmlloyd if you have a spare cycle and are interested in having a look it would be appreciated 🙏

dmlloyd:
The build is failing for me locally, but per that JDK bug I'd try adding `-implicit:none` to the `maven-compiler-plugin` arguments to `javac`.

compiler/pom.xml
andreaTP: needed by approval tests apparently, we should make sure this configuration only apply during tests

andreaTP:
@A248 I want those fixes for the next release that is likely to happen soon, hence I took over this PR, I hope you don't mind.
Thanks a lot for your effort, LGTM!

A248:
All good! Thank you for taking over the PR.

PR #1034, opened 2025-09-17
Clarify the contributing guide about mvnw

A248:
Correct me if I'm wrong, but it seems like you don't want `mvnw` to be a hard requirement. Because of that, I also specified the antrun version, because when I build the project using plain `mvn`, the missing antrun version causes `wasm-tools` to fail to run its Ant task.


andreaTP:
CI failures are fixed in #1032 , sorry for the noise.

andreaTP:
Thanks for the fixes! :+1:

PR #1035, opened 2025-09-17
[WIP] Java 25 support

andreaTP:
This is almost finished, only need to wait for the next release of `maven-dependency-plugin` [reference](https://github.com/apache/maven-dependency-plugin/issues/1499#issuecomment-3303650847).

Tested locally on Zulu 25.

### DIFF
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
@@ -139,7 +139,7 @@ public final class ByteArrayMemory implements Memory {
 
         synchronized (monitor) {
             try {
-                if (((int) INT_ARR_HANDLE.getVolatile(buffer, address)) == expected) {
+                if (readInt(address) == expected) {
                     return waitOnMonitor(
                             address, (timeout < 0) ? Long.MAX_VALUE : timeout, monitor);
                 } else {
@@ -160,7 +160,7 @@ public final class ByteArrayMemory implements Memory {
 
         synchronized (monitor) {
             try {
-                if (((long) LONG_ARR_HANDLE.getVolatile(buffer, address)) == expected) {
+                if (readLong(address) == expected) {
                     return waitOnMonitor(
                             address, (timeout < 0) ? Long.MAX_VALUE : timeout, monitor);
                 } else {
--- a/simd/src/main/java-templates-21/com/dylibso/chicory/simd/VectorOperators.java
+++ b/simd/src/main/java-templates-21/com/dylibso/chicory/simd/VectorOperators.java
@@ -1,0 +1,24 @@
+package com.dylibso.chicory.simd;
+
+final class VectorOperators {
+
+    private VectorOperators() {}
+
+    static final jdk.incubator.vector.VectorOperators.Comparison NE = jdk.incubator.vector.VectorOperators.NE;
+    static final jdk.incubator.vector.VectorOperators.Binary LSHL = jdk.incubator.vector.VectorOperators.LSHL;
+    static final jdk.incubator.vector.VectorOperators.Binary LSHR = jdk.incubator.vector.VectorOperators.LSHR;
+    static final jdk.incubator.vector.VectorOperators.Binary ASHR = jdk.incubator.vector.VectorOperators.ASHR;
+    static final jdk.incubator.vector.VectorOperators.Comparison UNSIGNED_LT = jdk.incubator.vector.VectorOperators.UNSIGNED_LT;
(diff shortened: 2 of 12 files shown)


runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
```diff
@@ -139,7 +139,7 @@ public int waitOn(int address, int expected, long timeout) {
 
         synchronized (monitor) {
             try {
-                if (((int) INT_ARR_HANDLE.getVolatile(buffer, address)) == expected) {
```
andreaTP, line 142: `getVolatile` throws on Zulu 25 as unsupported, but we are in a `synchronized` block so it might be excess of diligence using it here.

wendigo:
You actually don't need a plugin release. You can update plugin dependency directly in a pom (in this case it's ASM)

andreaTP:
@wendigo thanks for the suggestion, have you already done it? do you have a diff? :pray:

wendigo:
@andreaTP around 3 years ago as we were struggling with forward testing new JDK versions before they are released. Here is how we do it: https://github.com/airlift/airbase/blob/master/airbase/pom.xml#L882. We simply force new ASM for the plugins and it works :)

andreaTP:
Nifty trick! Thanks a ton for sharing! :rocket:

andreaTP:
CI is green, ready for review!

simd/src/main/java-templates-21/com/dylibso/chicory/simd/VectorOperators.java
```diff
@@ -0,0 +1,24 @@
...
+
+final class VectorOperators {
+
+    private VectorOperators() {}
+
+    static final jdk.incubator.vector.VectorOperators.Comparison NE = jdk.incubator.vector.VectorOperators.NE;
+    static final jdk.incubator.vector.VectorOperators.Binary LSHL = jdk.incubator.vector.VectorOperators.LSHL;
+    static final jdk.incubator.vector.VectorOperators.Binary LSHR = jdk.incubator.vector.VectorOperators.LSHR;
+    static final jdk.incubator.vector.VectorOperators.Binary ASHR = jdk.incubator.vector.VectorOperators.ASHR;
+    static final jdk.incubator.vector.VectorOperators.Comparison UNSIGNED_LT = jdk.incubator.vector.VectorOperators.UNSIGNED_LT;
```
andreaTP, line 15: `jdk.incubator.vector.VectorOperators` API changed, e.g. `ULT` vs. `UNSIGNED_LT`.

I moved the API behind a class in this codebase to keep things decoupled and implemented it for Java 21 and 25.

simd/src/main/java/com/dylibso/chicory/simd/SimdInterpreterMachine.java
```diff
@@ -17,7 +17,6 @@
 import java.util.function.Function;
 import jdk.incubator.vector.LongVector;
 import jdk.incubator.vector.Vector;
-import jdk.incubator.vector.VectorOperators;
```
evacchi, line 20: ok so if I understand correctly, you are removing this `import` statement because now `VectorOperators` is a generated (templated) class that belongs to the same package.

simd/src/main/java-templates-21/com/dylibso/chicory/simd/VectorOperators.java
```diff
@@ -0,0 +1,24 @@
+package com.dylibso.chicory.simd;
+
+final class VectorOperators {
```
evacchi, line 7:
```suggestion
/**
  * Generated compatibility shim for cross-version compatibility with jdk.incubator.vector.VectorOperators
  *
  */
final class VectorOperators {
```

simd/src/main/java-templates/com/dylibso/chicory/simd/VectorOperators.java
```diff
@@ -0,0 +1,24 @@
+package com.dylibso.chicory.simd;
+
+final class VectorOperators {
```
evacchi, line 7:
```suggestion
/**
  * Generated compatibility shim for cross-version compatibility with jdk.incubator.vector.VectorOperators
  *
  */
final class VectorOperators {
```

evacchi (approved):
nits, otherwise approved

simd/src/main/java/com/dylibso/chicory/simd/SimdInterpreterMachine.java
andreaTP: correct

wendigo:
Nice!

PR #1041, opened 2025-09-22
POC add an optional caching layer for runtime compilation

andreaTP:
This is a quick and dirty example of how a caching mechanism for the `runtime-compiler`.
There are a few questions I'd like to discuss before moving on:

- with this approach the `CompilerResult` becomes public API, I'm unsure this is desirable ...
- the build-time compiler is already serializing things to disk, would it make sense to re-use the same mechanism?

Happy to hear feedback!

### DIFF
--- a/compiler/pom.xml
+++ b/compiler/pom.xml
@@ -51,6 +51,19 @@
       <scope>test</scope>
     </dependency>
     <dependency>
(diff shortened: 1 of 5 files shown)


andreaTP:
Superseded by #1077

PR #1044, opened 2025-09-23
User land compilation caching

andreaTP:
This is an example of how, as of today, is possible to implement caching, fully in user land, on top of the current `compiler`.

I think the API is not extremely intuitive and we should provide a better way of doing it.


andreaTP:
Superseded by #1077

PR #1067, opened 2025-10-10
add: CompiledModule interface to make it easier to work with multiple build time compiled modules.

chirino:


build-time-compiler/src/main/java/com/dylibso/chicory/build/time/compiler/Generator.java
```diff
@@ -244,4 +251,54 @@ private static void generateLoadMethod(
...
+        // Function<Instance, Machine> machineFactory() {
+        //    return ::create;
+        // }
+
+        cu.addImport(Instance.class);
+        cu.addImport(Machine.class);
+        cu.addImport(Function.class);
+
+        var functionType =
+                new ClassOrInterfaceType()
```
andreaTP, line 268: nit - nothing wrong with this approach, but is more "javaparser idiomatic"(as I know it) by using:
`parseClassOrInterfaceType` (all occurrences)

chirino:
> to avoid having people to reload multiple times the `WasmModule` should we cache it in a `static` field? Something like:
> 
> ```java
>     private static WasmModule MODULE = null;
> 
>     public static WasmModule load() {
>         if (MODULE == null) {
>             try (InputStream in = Wast2JsonModule.class.getResourceAsStream("Wast2JsonModule.meta")) {
>                 MODULE = Parser.parse(in);
>             } catch (IOException e) {
>                 throw new UncheckedIOException("Failed to load AOT WASM module", e);
>             }
>         }
> 
>         return MODULE;
>     }
> ```
> 
> wdyt? We can also do it in a follow up iteration, no need to be included here for the merge.

I think that's a good idea.   I'll merge what here so far and do a follow up PR to optimize this.

PR #1068, opened 2025-10-12
Add atomic fence test

andreas-karlsson:
Add litmus test for `atomic.fence` ordering

This adds a threads litmus exercising `atomic.fence` placement:

- Writer loop: `a++ ; fence ; b = a`
- Reader loop: `read b ; fence ; read a` → trap if `a < b`
- Run both loops concurrently for ~200ms.

Rationale
- On V8 and Chicory, removing the reader-side fence reliably triggers `a < b`, indicating reordering/visibility issues; with the fence present, we don’t observe the violation.

Limitations (per Wasm memory model according to AI)
- This is a racy program: `atomic.fence` orders a single thread’s accesses but does not, by itself, establish a cross-thread happens-before. Therefore this is a litmus, not a deterministic conformance test: both pass and fail outcomes are allowed by the spec on compliant runtimes.

Notes / Alternatives
- A spec-deterministic variant would introduce a synchronizing event (e.g., make `b` or a separate flag atomic with seq-cst store/load, or use wait/notify). That guarantees visibility after the event, but then the fence becomes redundant for this pattern.


wasm-corpus/src/main/resources/wat/threads-example.wat
```diff
@@ -100,4 +100,40 @@
...
+    
+    ;; Read 'b' and then 'a', separated by a fence.
+    (local.set $local_b (i32.load (i32.const 4)))
+    (atomic.fence)
+    (local.set $local_a (i32.load (i32.const 0)))
+
+    ;; Check the invariant: a >= b.
+    (if (i32.lt_s (local.get $local_a) (local.get $local_b))
+      (then
+        ;; This can only happen if the fence is broken.
```
andreas-karlsson, line 133: So like I stated in the PR description, I'm not so sure this is true? Maybe a compliant runtime could actually get here because the fence doesn't guarantee visibility across threads. What the spec actually says in these matters is a little above my pay grade.

PR #1072, opened 2025-10-14
Cache the WasmModule in a static field to avoid parsing it multiple times

chirino:
Follow up PR to https://github.com/dylibso/chicory/pull/1067

We use the Initialization-on-Demand Holder idiom so that we get lazy, thread-safe initialization of static value.

### DIFF
--- a/build-time-compiler/src/main/java/com/dylibso/chicory/build/time/compiler/Generator.java
+++ b/build-time-compiler/src/main/java/com/dylibso/chicory/build/time/compiler/Generator.java
@@ -33,7 +33,6 @@ import com.github.javaparser.ast.stmt.CatchClause;
 import com.github.javaparser.ast.stmt.ReturnStmt;
 import com.github.javaparser.ast.stmt.ThrowStmt;
 import com.github.javaparser.ast.stmt.TryStmt;
-import com.github.javaparser.ast.type.ClassOrInterfaceType;
 import com.github.javaparser.utils.SourceRoot;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
@@ -103,7 +102,8 @@ public class Generator {
         type.addConstructor(Modifier.Keyword.PUBLIC).createBody();
 
         generateCreateMethod(cu, type, machineName);
-        generateLoadMethod(cu, type, moduleName, wasmName);
+        generateWasmModuleHolderInnerClass(type, moduleName, wasmName);
+        generateLoadMethod(cu, type);
         generateMachineFactoryMethod(cu, type, moduleName);
         generateWasmModuleMethod(cu, type, moduleName);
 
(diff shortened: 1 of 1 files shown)


chirino:
The new generated code looks like:

```java
    private static class WasmModuleHolder {

        static final WasmModule INSTANCE;

        static {
            try (InputStream in = Wast2JsonModule.class.getResourceAsStream("Wast2JsonModule.meta")) {
                INSTANCE = Parser.parse(in);
            } catch (IOException e) {
                throw new UncheckedIOException("Failed to load AOT WASM module", e);
            }
        }
    }

    public static WasmModule load() {
        try {
            return WasmModuleHolder.INSTANCE;
        } catch (ExceptionInInitializerError e) {
            throw new UncheckedIOException(e.getMessage(), (IOException) e.getCause());
        }
    }
```

andreaTP:
Thanks for the follow up @chirino , do we need the `WasmModuleHolder` at all?
Also, the `try-catch` in the `load` method seems redundant, or am I missing anything?

I'm expecting something like:
```java
    private static final WasmModule INSTANCE;

    static {
        try (InputStream in = Wast2JsonModule.class.getResourceAsStream("Wast2JsonModule.meta")) {
            INSTANCE = Parser.parse(in);
        } catch (IOException e) {
            throw new UncheckedIOException("Failed to load AOT WASM module", e);
        }
    }

    public static WasmModule load() {
        return INSTANCE;
    }
```

build-time-compiler/src/main/java/com/dylibso/chicory/build/time/compiler/Generator.java
```diff
@@ -200,56 +201,137 @@ private static void generateCreateMethod(
...
+                                new MethodCallExpr()
+                                        .setScope(new NameExpr("Parser"))
+                                        .setName("parse")
+                                        .addArgument(new NameExpr("in")),
+                                com.github.javaparser.ast.expr.AssignExpr.Operator.ASSIGN));
+
+        var newException =
+                new ObjectCreationExpr()
+                        .setType(parseClassOrInterfaceType("UncheckedIOException"))
+                        .addArgument(new StringLiteralExpr("Failed to load AOT WASM module"))
```
andreaTP, line 259: Noticed only now :sweat_smile: since you are at is: "build time compiled" instead of AOT

chirino: good catch.

chirino:
> Thanks for the follow up @chirino , do we need the `WasmModuleHolder` at all?
> Also, the `try-catch` in the `load` method seems redundant, or am I missing anything?

This is there just to keep the thrown exception backward compatible.  If we don't catch and throw again, the UncheckedIOException becomes a ExceptionInInitializerError.  Maybe ExceptionInInitializerError is a better runtime exception for this use case.  Not sure, we can simplify if you think users are not trying to explicitly catch the UncheckedIOException.

andreaTP:
> users are not trying to explicitly catch the UncheckedIOException

If a user is doing this is very likely trying to workaround a bug in our integration, it is an unchecked exception so is not a breaking change.... I think is better to simplify :+1:

andreaTP:
@chirino This PR shouldn’t have been merged because:

- There are unresolved comments: https://github.com/dylibso/chicory/pull/1072#issuecomment-3401009315
- It wasn’t approved yet, and our contribution guidelines require at least one positive review: https://github.com/dylibso/chicory/blob/main/CONTRIBUTING.md#code-reviews

I’ve applied the suggested changes myself and opened #1076 to fix this on top.

chirino:
My bad, I thought https://github.com/dylibso/chicory/compare/a74c55c2eab1559e07c1ef3d614e50025f0224ce..02a28e557a74f73bf3b44e8175bb7172c0525362 was addressing the remaining issues.

PR #1076, opened 2025-10-15
Apply review comments from 1072

andreaTP:
Here I apply the suggestions made in: https://github.com/dylibso/chicory/pull/1072#issuecomment-3401009315
Plus a couple of super minor nit in the usage of JavaParser.


chirino:
So we don't use the Holder inner class anymore?,  this means that the WASM module is going to get parsed even the user doesn't call the load() method.  Use case: he could might be loading many CompiledModules into a registry.  But he only might use a subset of those modules based on requests being received.

chirino:
See: https://en.wikipedia.org/wiki/Initialization-on-demand_holder_idiom for more details about the Holder pattern.

andreaTP:
The use case we want to enable using the holder pattern is to register compiled modules in maps without triggering the load of `WasmModule`:

```java
registry.put(“my-compiled-module”, new CompiledModule()) 
```

I'm good with this direction as it enables useful use-cases :+1:

PR #1077, opened 2025-10-15
Directory backed runtime compiler cache

chirino:
This expands on https://github.com/dylibso/chicory/pull/1041 and adds a built in Directory Cache implementation.

When you use it with the runtime compiler:
```
          var cache = new DirectoryCache(Path.of("/cache"));
          var instance =
                Instance.builder(module)
                        .withMachineFactory(
                                MachineFactoryCompiler.builder(module).withCache(cache).compile())
                        .build();
```
It will create the files like:
```
/cache/.tmp/f-9979505440111212447.tmp
```

and then move the `/cache/.tmp/f-9979505440111212447.tmp` file in an atomic operation to
`/cache/sha256/kr/gytkcm43c34ksqta8gmddw4ycfquc2g0qfifcpb-w.jar` (the file path is based on the sha256 of the original wasm file).

It should be safe to use the same cache location by different processes.  If any error occurs while generating the code into the temp dir, the temp file will get deleted.  But if the process is killed before we can clean up, the temp file will not get cleaned up.  User may want manually clean the /cache/.tmp directory when no process is writing to the cache.  

we use 2 character prefix of the sha to avoid hitting directory scaling issues with file systems.  So if you have a VERY large cache all the cache entries will be partitioned over 4096 dirs.


chirino:
Given different version of chicory can generate different runtime byte codes, should the directory path contain some version info?  Or should this just be done by the user which is configuring the cache dir location?

compiler/pom.xml
```diff
@@ -45,6 +45,12 @@
       <artifactId>wasm-corpus</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.google.jimfs</groupId>
+      <artifactId>jimfs</artifactId>
```
andreaTP: I'll have a look tomorrow, meanwhile, quick nit that roastedroot.zerofs is almost a drop in replacement without guava ;-)

chirino: good point

compiler/pom.xml
```diff
@@ -45,6 +45,12 @@
       <artifactId>wasm-corpus</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>io.roastedroot</groupId>
+      <artifactId>zerofs</artifactId>
+      <version>0.1.0</version>
```
andreaTP: can omit the version as it's in the root pom

wasm/src/main/java/com/dylibso/chicory/wasm/WasmModule.java
```diff
@@ -164,6 +171,7 @@ public static final class Builder {
         private final Map<String, CustomSection> customSections = new HashMap<>();
         private final List<Integer> ignoredSections = new ArrayList<>();
         private boolean validate = true;
+        private String messageDigest;
```
andreaTP: nit - just "digest" or "payloadDigest"?

compiler/src/main/java/com/dylibso/chicory/compiler/MachineFactoryCompiler.java
```diff
@@ -102,14 +108,121 @@ public Builder withInterpretedFunctions(Set<Integer> interpretedFunctions) {
...
+
         public Function<Instance, Machine> compile() {
-            var result =
-                    compilerBuilder
-                            .withClassCollectorFactory(ClassLoadingCollector::new)
-                            .build()
-                            .compile();
-            var collector = (ClassLoadingCollector) result.collector();
-            return new MachineFactory(module, collector.machineFactory());
+
```
andreaTP, line 132: The rest of the diff in this file is hard to follow, I understand that it's due to the current API, but I trust it can be better modularized.

compiler/src/main/java/com/dylibso/chicory/compiler/DirectoryCache.java
```diff
@@ -0,0 +1,197 @@
...
+import java.nio.file.attribute.BasicFileAttributes;
+import java.util.Locale;
+import java.util.Objects;
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.regex.Pattern;
+
+/**
+ * Disk-backed sharded directory cache.
+ */
+public class DirectoryCache {
```
andreaTP: I'm a bit reluctant to roll out and, eventually, maintain our own version of a disk based cache.
I'm worried that we will waste time debugging issues with filesystems etc. (like on Windows etc.).

The alternative approach would be to define the interfaces in the core modules, spin a new Maven sub-module that will depend on, e.g., `ehcache` and have a minimal implementation based on it.

wdyt?

compiler/src/main/java/com/dylibso/chicory/compiler/Cache.java
```diff
@@ -0,0 +1,39 @@
...
+     *
+     * @param key    "algo:digest"
+     */
+    Path get(String key) throws IOException;
+
+    /**
+     * Create a unique temporary directory suitable for writing the computation output.
+     * The directory will be on the same filesystem as the final target so that ATOMIC_MOVE works.
+     */
+    TempDir createTempDir() throws IOException;
```
andreaTP: Can it be automatically handled on `put` ?

compiler/src/main/java/com/dylibso/chicory/compiler/Cache.java
```diff
@@ -0,0 +1,39 @@
...
+    interface TempDir extends Closeable {
+        Path path();
+    }
+
+    /**
+     * Return the directory for the given key if it exists else null.
+     *
+     * @param key    "algo:digest"
+     */
+    Path get(String key) throws IOException;
```
andreaTP: I'm a little surprised by this API, I was expecting:

```java
TempDir get(String key)
```

to mirror `put`.

compiler/src/main/java/com/dylibso/chicory/compiler/Cache.java
```diff
@@ -0,0 +1,39 @@
...
+    /**
+     * Atomically publish a completed temp directory into the cache location for the key.
+     * If another thread/process already published for this key the
+     * existing path is returned.
+     *
+     * @param key    "algo:digest"
+     * @param tmpDir a directory containing fully written results (created via createTempDir())
+     * @return the final cache directory path
+     */
+    Path put(String key, TempDir tmpDir) throws IOException;
```
andreaTP: Seems like we are spilling internals to the API layer here, I'm expecting `put` to return `void`.

wasm/src/main/java/com/dylibso/chicory/wasm/WasmModule.java
chirino: will do.

compiler/src/main/java/com/dylibso/chicory/compiler/Cache.java
chirino: The TempDir is temporary and only exists while it holds the compiled data while is being written to disk and until it is atomically inserted into the cache.  This avoids us inserting partially computed data into the cache.  Note that the TempDir is closable and we use it in a try /w resources block so that it's contents are always cleaned up.

Once it's in the cache, it's not temp anymore since it will survive restarts.

compiler/src/main/java/com/dylibso/chicory/compiler/Cache.java
chirino: It could return void. I don't think we use the result.

compiler/src/main/java/com/dylibso/chicory/compiler/Cache.java
chirino: I wanted the cache contents to be serialized to disk in a easy to understand way.  I did not want it to be some serialized format that is hard to understand or maintain.  Basically the compiled classes just become .class files in the cache directory.  So we need a way to atomically write a set of files to a directory.  Writing to a tmp directory and then doing and then moving the directory is well understood way of doing this.  The limitation for this is that the temp directory has to exist in the same volume that your moving to.  So the API help enforce this pattern.  It creates a temp dir in the same volume for you, you write data to the dir, then pass that back to the put to atomically place it into the cache.

compiler/src/main/java/com/dylibso/chicory/compiler/DirectoryCache.java
chirino: It's about 200 loc, not super complex and I'd be happy to add more tests to exercise it on windows.  I've worked on many multi-platform file storage systems in the past and I'm well ware of all the Windows vs Posix issues.

chirino: plus we don't need the in memory caching the ehcache would be doing.

compiler/src/main/java/com/dylibso/chicory/compiler/MachineFactoryCompiler.java
chirino: I think I can simplify.

compiler/src/main/java/com/dylibso/chicory/compiler/DirectoryCache.java
andreaTP: I have no problem rolling it out as an experimental module at first, to evaluate usefulness and usage in the wild.

compiler/src/main/java/com/dylibso/chicory/compiler/Cache.java
```diff
@@ -0,0 +1,37 @@
+package com.dylibso.chicory.compiler;
+
+import java.io.Closeable;
+import java.io.IOException;
+import java.nio.file.Path;
+
+public interface Cache {
```
andreaTP, line 5: Can we push the abstraction a bit higher?
e.g. storing `ClassLoadingCollector`s, or even MachineFactory?

this way the API results in 2 functions:

```java
MachineFactory get(String key) {}
void put (String key, MachineFactory factory) {}
```

If you want to separate adding only the Cache interface in a separate PR I think we can move faster and have more focused discussions :slightly_smiling_face:

compiler/src/main/java/com/dylibso/chicory/compiler/Cache.java
```diff
@@ -0,0 +1,37 @@
...
+
+    /**
+     * The temp dir is deleted once closed.
+     */
+    interface TempDir extends Closeable {
```
andreaTP: This is a "marker interface" I'd prefer to hide the complexity rising the API abstraction other than exposing this.

compiler/pom.xml
chirino: done

compiler/src/main/java/com/dylibso/chicory/compiler/MachineFactoryCompiler.java
chirino: Is it better now?

compiler/src/main/java/com/dylibso/chicory/compiler/DirectoryCache.java
chirino: I'll refactor that out.

compiler/src/main/java/com/dylibso/chicory/compiler/Cache.java
chirino: That would force us to expose public apis to encode/decod MachineFactory instances which I think we should aovid.

compiler/src/main/java/com/dylibso/chicory/compiler/Cache.java
chirino: It's gone now.

compiler/src/main/java/com/dylibso/chicory/compiler/Cache.java
chirino: We now use a simpler Cache interface.  Forces, a copy of the generated byte codes into memory, but makes the cache interface simpler so that it can also be implemented with things like redis etc.

compiler/src/main/java/com/dylibso/chicory/compiler/MachineFactoryCompiler.java
andreaTP: yup, thanks :+1:

compiler/src/main/java/com/dylibso/chicory/compiler/MachineFactoryCompiler.java
```diff
@@ -102,14 +112,115 @@ public Builder withInterpretedFunctions(Set<Integer> interpretedFunctions) {
...
+                    jos.closeEntry();
+                }
+            }
+            return baos.toByteArray();
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
+    }
+
+    private static ClassLoadingCollector loadCollector(byte[] jarData) throws IOException {
```
andreaTP:
```suggestion
    private static ClassLoadingCollector loadClassLoadingCollector(byte[] jarData) throws IOException {
```

nit to make it consistent

dircache/src/main/java/com/dylibso/chicory/dircache/internal/PathUtils.java
```diff
@@ -0,0 +1,36 @@
+package com.dylibso.chicory.compiler.internal;
+
+import java.io.IOException;
+import java.nio.file.FileVisitResult;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.SimpleFileVisitor;
+import java.nio.file.attribute.BasicFileAttributes;
+
+public final class PathUtils {
```
andreaTP, line 10: Move to `dircache` sub-module

compiler/src/test/java/com/dylibso/chicory/compiler/internal/CacheTest.java
```diff
@@ -0,0 +1,91 @@
...
+import com.dylibso.chicory.compiler.Cache;
+import com.dylibso.chicory.compiler.MachineFactoryCompiler;
+import com.dylibso.chicory.runtime.Instance;
+import com.dylibso.chicory.wasm.Parser;
+import java.io.IOException;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicInteger;
+import org.junit.jupiter.api.Test;
+
+public class CacheTest {
```
andreaTP, line 15: Nice :+1:

dircache/src/test/java/com/dylibso/chicory/dircache/DirectoryCacheTest.java
```diff
@@ -0,0 +1,191 @@
...
+
+    @Test
+    public void shouldCacheCompiledResultNativeFS() throws IOException {
+
+        var module =
+                Parser.parse(
+                        DirectoryCacheTest.class.getResourceAsStream(
+                                "/compiled/count_vowels.rs.wasm"));
+
+        var cacheDir = Files.createTempDirectory("test");
```
andreaTP: Use JUnit `TempDir` instead? like: https://github.com/dylibso/chicory/blob/52dd3cff450acc2f9d0dfb7f12ec93ac96a01c42/wasm-tools/src/test/java/com/dylibso/chicory/tools/wasm/WasmToolsTest.java#L19

dircache/src/test/java/com/dylibso/chicory/dircache/DirectoryCacheTest.java
```diff
@@ -0,0 +1,191 @@
...
+
+    @Test
+    public void testConcurrentAccessNativeFS() throws IOException {
+
+        var module =
+                Parser.parse(
+                        DirectoryCacheTest.class.getResourceAsStream(
+                                "/compiled/count_vowels.rs.wasm"));
+
+        var cacheDir = Files.createTempDirectory("test");
```
andreaTP: JUnit `TempDir`

wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java
```diff
@@ -231,8 +235,18 @@ public static WasmModule parse(Path path) {
...
+        MessageDigest messageDigest = null;
         try (InputStream is = inputStreamSupplier.get()) {
-            parse(is, (s) -> onSection(moduleBuilder, s));
+            InputStream maybeDigestedInputStream = is;
+            // wrap the input stream so we can calculate the sha256 hash of the buffer
+            try {
+                messageDigest = MessageDigest.getInstance("SHA-256");
+                maybeDigestedInputStream = new DigestInputStream(is, messageDigest);
+            } catch (NoSuchAlgorithmException ignore) {
+                // if we can't get the algorithm, then we just don't set the MessageDigest
```
andreaTP: This is going to be very very hard to debug ...

Can we just throw the exception? Or make the `MessageDigest` String configurable?

andreaTP (changes requested):
A few minor notes, but this is definitely going in the right direction!

After those small fixes either add a page to the docs in the PR or open a ticket so that we won't forget.
Additionally, add the new dependency to the `bom`, and verify the included/excluded modules in the release profile.

Thanks a lot!

wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java
chirino: Given we only need the algorithm for caching I was thinking it would be better to gracefully fallback to not setting the digest which in turn causes us to not cache.

We can make the digest configurable.  From a system prop like:

```java
    private static final String DIGEST_ALGORITHM = System.getProperty("com.dylibso.chicory.wasm.Parser.DIGEST_ALGORITHM", "SHA-256");
```

andreaTP: > From a system prop?

Seems reasonable, if we make it configurable I think we should loudly fail, otherwise the failure is going to pass unobserved and the fact that caching doesn't work is too long down the line to catch it easily.

andreaTP (approved):
LGTM, thanks @chirino great job!

andreaTP:
Let's do a release in the following days to use those changes in the Bazel integration.
@chirino is a PATCH release enough?
Do you see any possible issue mixing `compiler` and `wasm` modules?

chirino:
The new compiler version depends on the new wasm version (to access the wasm digest).  So a normal release might be safer.

PR #1084, opened 2025-10-24
Work around JVM miscompile of unsigned integer comparison

### LINKED ISSUE #843: JVM issue with `memoryCopy`

dmlloyd:
The HotSpot JVM appears to miscompile the intrinsic for unsigned integer comparision in certain compilation tiers. We can avoid it by manually inlining the comparison.

Credit to @chirino and @andreaTP for chasing this down with some novel detective work.

Fixes #843

### DIFF
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/MemCopyWorkaround.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/MemCopyWorkaround.java
@@ -13,7 +13,8 @@ public final class MemCopyWorkaround {
     }
 
     public static boolean shouldUseMemWorkaround() {
-        return shouldUseMemWorkaround(System.getProperty("java.version"));
+        return false;
+        // return shouldUseMemWorkaround(System.getProperty("java.version"));
     }
 
     public static boolean shouldUseMemWorkaround(String version) {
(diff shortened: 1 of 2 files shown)


runtime/src/main/java/com/dylibso/chicory/runtime/MemCopyWorkaround.java
```diff
@@ -13,7 +13,8 @@ private interface MemoryCopyFunc {
     }
 
     public static boolean shouldUseMemWorkaround() {
-        return shouldUseMemWorkaround(System.getProperty("java.version"));
+        return false;
```
andreaTP, line 16: I'm ok with this, should we deprecate this class entirely so that it can be removed in the next major release?

dmlloyd:
Surprise! It's failing! Interesting that the fix worked locally though 🤔

dmlloyd:
It looks like the intrinsic thing was another red herring that affected C2 in an oblique way. So, back to the drawing board!

PR #1090, opened 2025-10-29
resolve ValTypes after TypeSection is formed

andreaTP:
This is a "step 0" to enable the implementation of Wasm GC.
The plan is to try, as much as possible, to come up with small, self contained PRs implementing Wasm GC to guarantee the rest of the code base consistency and reduce the risk of accidentally breaking the stable API.

In Wasm GC the recursive types refer to themselves, so we need to split the `ValType` computation from the resolution.
In this PR, the resolution of `ValType` happens mutating the already instantiated `ValType` instead of doing it in the Builder.

Thanks to @evacchi for the help in getting this to the bottom of the last detail.
@hypercubestart if you wanna have a look :slightly_smiling_face:

### DIFF
--- a/wasm/src/main/java/com/dylibso/chicory/wasm/types/ValType.java
+++ b/wasm/src/main/java/com/dylibso/chicory/wasm/types/ValType.java
@@ -3,7 +3,6 @@ package com.dylibso.chicory.wasm.types;
 import com.dylibso.chicory.wasm.ChicoryException;
 import com.dylibso.chicory.wasm.InvalidException;
 import java.util.List;
-import java.util.Objects;
 import java.util.function.Function;
 
 /**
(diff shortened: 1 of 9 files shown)


wasm/src/main/java/com/dylibso/chicory/wasm/types/ValType.java
```diff
@@ -58,13 +66,33 @@ private ValType(int opcode, int typeIdx, FunctionType resolvedFunctionType) {
             typeIdx = TypeIdxCode.EXN.code();
             opcode = ID.RefNull;
         } else if ((opcode == ID.RefNull || opcode == ID.Ref) && typeIdx >= 0) {
-            Objects.requireNonNull(resolvedFunctionType);
+            assert resolvedFunctionTypeId >= 0;
```
evacchi, line 69: I don't know if this is intended, but notice this won't throw unless assertions are explicitly enabled (i.e. the `-ea` flag), which happens rarely

evacchi (approved):
one minor nit!

wasm/src/main/java/com/dylibso/chicory/wasm/types/ValType.java
andreaTP: It happens in CI at least :wink: : https://github.com/dylibso/chicory/blob/00983a75a6870fc0faece34f095a5add27282b9a/.github/workflows/ci.yaml#L54

wasm/src/main/java/com/dylibso/chicory/wasm/types/ValType.java
```diff
@@ -58,13 +66,33 @@ private ValType(int opcode, int typeIdx, FunctionType resolvedFunctionType) {
...
         this.resolvedFunctionType = resolvedFunctionType;
 
         this.id = createId(opcode, typeIdx);
     }
 
+    public ValType resolve(TypeSection typeSection) {
+        return resolve(typeSection, -1);
+    }
+
+    public ValType resolve(TypeSection typeSection, int currentIdx) {
```
andreaTP, line 81: I'm going forward and I'm not 100% sure about this new API ...

andreaTP:
superseded by #1097

PR #1097, opened 2025-11-05
Wasm gc step1

andreaTP:
Attempting to validate the changes in #1090 I ended up(again) porting most of the work(and some new) already done to this new branch.

It's not "finished" and the encoding of types might change again, but I'm convinced is a decent starting point, and it enables implementing GC opcodes.

The public API is preserved and the test coverage is increased without regressions.

The current plan is:
- publish 1.6 with everything ready but those changes
- merge this PR
- keep doing incremental changes to support entire WasmGC
- plan a release when the process is done

### DIFF
--- a/wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java
+++ b/wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java
@@ -584,9 +584,9 @@ public final class Parser {
     }
 
     private static SubType parseSubType(int id, ByteBuffer buffer) {
-        if (id == 0x50
-                || // non final typeIdx
-                id == 0x4F) { // final typeIdx
+        if (id == 0x50 // non final typeIdx
+                || id == 0x4F // final typeIdx
+        ) {
             var count = (int) readVarUInt32(buffer);
             var typeIdxs = new int[count];
 
--- a/wasm/src/main/java/com/dylibso/chicory/wasm/Validator.java
+++ b/wasm/src/main/java/com/dylibso/chicory/wasm/Validator.java
@@ -2170,7 +2170,6 @@ final class Validator {
         }
 
         if (!errors.isEmpty()) {
-            errors.stream().forEach(e -> e.printStackTrace());
             throw new InvalidException(
                     errors.stream().map(Throwable::getMessage).collect(joining(" - ")));
         }
--- a/wasm/src/main/java/com/dylibso/chicory/wasm/types/ArrayType.java
+++ b/wasm/src/main/java/com/dylibso/chicory/wasm/types/ArrayType.java
@@ -1,5 +1,7 @@
 package com.dylibso.chicory.wasm.types;
 
+import java.util.Objects;
+
 public final class ArrayType {
     private final FieldType fieldType;
 
@@ -9,6 +11,20 @@ public final class ArrayType {
 
     public FieldType fieldType() {
         return fieldType;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        ArrayType arrayType = (ArrayType) o;
+        return Objects.equals(fieldType, arrayType.fieldType);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hashCode(fieldType);
     }
 
     public static Builder builder() {
--- a/wasm/src/main/java/com/dylibso/chicory/wasm/types/CompType.java
+++ b/wasm/src/main/java/com/dylibso/chicory/wasm/types/CompType.java
@@ -7,7 +7,15 @@ public final class CompType {
     private final StructType structType;
     private final FunctionType funcType;
 
+    private static void requireExactlyOneNonNull(Object a, Object b, Object c) {
+        if ((a == null ? 0 : 1) + (b == null ? 0 : 1) + (c == null ? 0 : 1) != 1) {
+            throw new IllegalArgumentException("Exactly one field must be filled");
+        }
+    }
+
     private CompType(ArrayType arrayType, StructType structType, FunctionType functionType) {
+        requireExactlyOneNonNull(arrayType, structType, functionType);
+
         this.arrayType = arrayType;
         this.structType = structType;
         this.funcType = functionType;
--- a/wasm/src/main/java/com/dylibso/chicory/wasm/types/StorageType.java
+++ b/wasm/src/main/java/com/dylibso/chicory/wasm/types/StorageType.java
@@ -1,10 +1,20 @@
 package com.dylibso.chicory.wasm.types;
 
+import java.util.Objects;
+
 public final class StorageType {
     private final ValType valType;
     private final PackedType packedType;
 
+    private static void requireExactlyOneNonNull(Object a, Object b) {
+        if ((a == null ? 0 : 1) + (b == null ? 0 : 1) != 1) {
+            throw new IllegalArgumentException("Exactly one field must be filled");
+        }
+    }
+
     private StorageType(ValType valType, PackedType packedType) {
+        requireExactlyOneNonNull(valType, packedType);
+
         this.valType = valType;
         this.packedType = packedType;
     }
@@ -15,6 +25,20 @@ public final class StorageType {
 
     public PackedType packedType() {
         return packedType;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        StorageType that = (StorageType) o;
+        return Objects.equals(valType, that.valType) && packedType == that.packedType;
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(valType, packedType);
     }
 
     public static Builder builder() {
--- a/wasm/src/main/java/com/dylibso/chicory/wasm/types/TypeSection.java
+++ b/wasm/src/main/java/com/dylibso/chicory/wasm/types/TypeSection.java
@@ -14,6 +14,10 @@ public final class TypeSection extends Section {
         this.types = List.copyOf(types);
     }
 
+    // https://github.com/WebAssembly/gc/blob/main/proposals/gc/MVP.md#type-definitions
+    // > the number of type section entries is now the number of recursion groups rather than the
+    // number of individual types.
+    // types() returns the flattened list of individual types
     public FunctionType[] types() {
         return types.stream()
                 .filter(RecType::isLegacy)
--- a/wasm/src/main/java/com/dylibso/chicory/wasm/types/ValType.java
+++ b/wasm/src/main/java/com/dylibso/chicory/wasm/types/ValType.java
@@ -394,16 +394,16 @@ public final class ValType {
         public static final int i31 = 0x6C;
         public static final int EqRef = 0x6D;
         public static final int AnyRef = 0x6E;
-        public static final int ExternRef = 0x6f;
+        public static final int ExternRef = 0x6F;
         public static final int FuncRef = 0x70;
         public static final int NoneRef = 0x71;
         public static final int NoExternRef = 0x72;
         public static final int NoFuncRef = 0x73;
-        public static final int V128 = 0x7b;
-        public static final int F64 = 0x7c;
-        public static final int F32 = 0x7d;
-        public static final int I64 = 0x7e;
-        public static final int I32 = 0x7f;
+        public static final int V128 = 0x7B;
+        public static final int F64 = 0x7C;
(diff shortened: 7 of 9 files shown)


wasm/src/main/java/com/dylibso/chicory/wasm/types/CompType.java
```diff
@@ -0,0 +1,74 @@
+package com.dylibso.chicory.wasm.types;
+
+import java.util.Objects;
+
+public final class CompType {
+    private final ArrayType arrayType;
+    private final StructType structType;
+    private final FunctionType funcType;
+
+    private CompType(ArrayType arrayType, StructType structType, FunctionType functionType) {
```
tombentley, line 16: `comptype ::= <functype> | <structtype> | <arraytype>`

Do you want to check the exclusivity as precondition?

wasm/src/main/java/com/dylibso/chicory/wasm/types/StorageType.java
```diff
@@ -0,0 +1,44 @@
+package com.dylibso.chicory.wasm.types;
+
+public final class StorageType {
+    private final ValType valType;
+    private final PackedType packedType;
+
+    private StorageType(ValType valType, PackedType packedType) {
```
tombentley, line 15: Again, maybe check that exactly one is null

wasm/src/main/java/com/dylibso/chicory/wasm/types/FunctionType.java
```diff
@@ -36,12 +34,15 @@ public boolean equals(Object obj) {
     }
 
     public boolean equals(FunctionType other) {
-        return hashCode == other.hashCode && paramsMatch(other) && returnsMatch(other);
+        // For type equivalence, we need structural comparison even when hashCodes differ
```
tombentley, line 37: I think this means `hashCode()` can be inconsistent with `equals()`, which can be pretty unexpected if your objects leak to code that's not written with such inconsistency in mind. So I wonder if it's strictly necessary to couple the Java notion of equality to type equivalence? If not then you could avoid the WTF kind of moments that people have when HashSets don't work.

wasm/src/main/java/com/dylibso/chicory/wasm/types/ArrayType.java
```diff
@@ -0,0 +1,32 @@
+package com.dylibso.chicory.wasm.types;
+
+public final class ArrayType {
```
tombentley, line 5: Other types override equals and hashCode, but not here. Is that intentional?

wasm/src/main/java/com/dylibso/chicory/wasm/types/TypeSection.java
```diff
@@ -1,40 +1,68 @@
...
-    private final List<FunctionType> types;
+    private final List<RecType> types;
 
-    private TypeSection(List<FunctionType> types) {
+    private TypeSection(List<RecType> types) {
         super(SectionId.TYPE);
         this.types = List.copyOf(types);
     }
 
     public FunctionType[] types() {
```
tombentley, line 21: The fact that this returns different types than were passed to the constructor might be worthy of a comment.

wasm/src/main/java/com/dylibso/chicory/wasm/types/ValType.java
```diff
@@ -321,10 +388,17 @@ private ID() {}
...
         public static final int Ref = 0x64;
-        public static final int ExternRef = 0x6f;
-        // From the Exception Handling proposal
         public static final int ExnRef = 0x69; // -0x17
+        public static final int ArrayRef = 0x6A;
+        public static final int StructRef = 0x6B;
+        public static final int i31 = 0x6C;
+        public static final int EqRef = 0x6D;
+        public static final int AnyRef = 0x6E;
+        public static final int ExternRef = 0x6f;
```
tombentley: You're using both lower and upper case letters in your hex.

wasm/src/main/java/com/dylibso/chicory/wasm/types/ValType.java
```diff
@@ -408,22 +482,40 @@ public boolean isReference() {
...
+                    opcode,
+                    typeIdx,
+                    (ValType.isReference(opcode) && !ValType.isAbsHeapType(opcode) && typeIdx >= 0)
+                            ? typeIdx
+                            : -1,
+                    SubType.builder()
+                            .withCompType(
+                                    CompType.builder().withFuncType(resolvedFunctionType).build())
+                            .build()
+                            .hashCode());
```
tombentley, line 502: Here you create a builder and a SubType only to compute a hashCode. Is there a less garbage-generating way?

wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java
```diff
@@ -491,66 +509,163 @@ private static RawSection parseRawSection(ByteBuffer buffer, byte sectionId, lon
...
+        var count = (int) readVarUInt32(buffer);
+        var builder = StructType.builder();
+        for (int i = 0; i < count; i++) {
+            builder.addFieldType(parseFieldType(buffer));
+        }
+        return builder.build();
+    }
+
+    private static FunctionType parseFunctionType(ByteBuffer buffer) {
+        var paramCount = (int) readVarUInt32(buffer);
```
tombentley, line 547: What happens on `int` overflow?

wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java
```diff
@@ -491,66 +509,163 @@ private static RawSection parseRawSection(ByteBuffer buffer, byte sectionId, lon
...
+                                + " was not 0x5E, 0x5f or 0x60");
+        }
+    }
 
-            // Parse function types (form = 0x60)
-            var paramCount = (int) readVarUInt32(buffer);
-            var paramsBuilder = new ValType.Builder[paramCount];
+    private static SubType parseSubType(int id, ByteBuffer buffer) {
+        if (id == 0x50
+                || // non final typeIdx
```
tombentley: Does the comment apply to the previous line?

wasm/src/main/java/com/dylibso/chicory/wasm/Validator.java
```diff
@@ -456,11 +458,27 @@ void validateData() {
...
 
     void validateTypes() {
-        var types = module.typeSection().types();
-        for (var i = 0; i < types.length; i++) {
-            var t = types[i];
-            t.params().forEach(this::validateValueType);
-            t.returns().forEach(this::validateValueType);
+        for (var i = 0; i < module.typeSection().typeCount(); i++) {
+            var t = module.typeSection().getRecType(i);
+            // TODO: fix me!
```
tombentley, line 463: Did you want to merge with this TODO?

wasm/src/main/java/com/dylibso/chicory/wasm/Validator.java
```diff
@@ -456,11 +458,27 @@ void validateData() {
...
+            if (t.isLegacy()) {
+                // The following code fixes the 2 tests, but breaks many things in GC:
+                // FunctionReferencesType-equivalence.test2
+                // FunctionReferencesType-equivalence.test3
+                // TODO: seems "wrong" when using WasmGC
```
tombentley, line 468: Did you want to merge with this TODO?

wasm/src/main/java/com/dylibso/chicory/wasm/Validator.java
```diff
@@ -2152,6 +2170,7 @@ void validateFunction(int funcIdx, FunctionBody body, FunctionType functionType)
         }
 
         if (!errors.isEmpty()) {
+            errors.stream().forEach(e -> e.printStackTrace());
```
tombentley: Was the the printing just for debugging purposes?

andreaTP:
Thanks a ton @tombentley for the review, much, much appreciated :pray: 

I'm about to go on vacation but will address your comments early next year!

wasm/src/main/java/com/dylibso/chicory/wasm/types/CompType.java
andreaTP: yup :+1:

wasm/src/main/java/com/dylibso/chicory/wasm/types/StorageType.java
andreaTP: :+1:

wasm/src/main/java/com/dylibso/chicory/wasm/types/FunctionType.java
andreaTP: Overall this is a very fair comment, historically we are inheriting this approach from the previous work on Function References.
I think you are right and we should, eventually, move away from it, I'd give a try to address this in an upcoming PR when dealing with types canonicalization if looks acceptable.

As a remark, we don't expect people to store those in Maps and Sets.

wasm/src/main/java/com/dylibso/chicory/wasm/types/ArrayType.java
andreaTP: good catch, done :+1:

wasm/src/main/java/com/dylibso/chicory/wasm/types/TypeSection.java
andreaTP: fair, done :+1:

wasm/src/main/java/com/dylibso/chicory/wasm/types/ValType.java
andreaTP: good catch! fixed!

wasm/src/main/java/com/dylibso/chicory/wasm/types/ValType.java
andreaTP: This method is marked as `Deprecated` and no longer used, I'd avoid to optimize over something that is supposed to be removed.

wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java
andreaTP: Currently, we are not checking some of the limits in place for other runtimes.
The "real world" limit should be this:
https://github.com/dylibso/chicory/blob/f6850866ba54f4b9df857cb5ac12207d5028fa6e/wasm/src/main/java/com/dylibso/chicory/wasm/WasmLimits.java#L25

Most(all?) compilers and optimizers are going to respect that limit (alternatively things won't work in the browser).
Instead of fixing this aspect in this PR I'll open a separate issue to introduce more checks.

wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java
andreaTP: (autoformatter) should be good now

wasm/src/main/java/com/dylibso/chicory/wasm/Validator.java
andreaTP: The implementation in this PR barely reaches the Validator, i.e. is not complete.

When completing the Wasm GC implementation those TODO will go away naturally/organically.

wasm/src/main/java/com/dylibso/chicory/wasm/Validator.java
andreaTP: Same as before, this is a work in progress, I prefer to have a few TODOs around to easily get back to the initial thought process for now.

wasm/src/main/java/com/dylibso/chicory/wasm/Validator.java
andreaTP: Good one! Leftover, removed.

wasm/src/main/java/com/dylibso/chicory/wasm/Parser.java
andreaTP: opened #1154

andreaTP:
Thanks again for the review @tombentley !
I think I responded to all the comments and I'm ready for the next round :slightly_smiling_face:

andreaTP:
@andreas-karlsson I realize this may be outside your main area of interest in the project,
but I’m running a bit low on reviews at the moment and thought I’d ask if you could take a look :pray:

andreas-karlsson:
@andreaTP I'm happy to take a look. I probably won't have time until Sunday though.

andreaTP:
@andreas-karlsson not a big deal, this PR is stalled since 3 months now :sweat_smile: , take your time and thanks a lot :pray:

andreaTP:
Superseded by #1204

PR #1102, opened 2025-11-07
reset Android cache in CI

andreaTP:
The Android CI keeps failing, we should consider removing the caching at all I think...


andreaTP:
Let's merge to fix issues on PRs

Issue #1101, opened 2025-11-07
Flaky dircache test


andreaTP:
This test failed:
https://github.com/dylibso/chicory/actions/runs/19165390477/job/54784783313?pr=1082#step:6:4714

need to check if it's flaky or it was just a runner hiccup.

andreaTP:
Fixed in #1103

andreaTP:
Apparently still happening:
https://github.com/dylibso/chicory/actions/runs/19427456838/job/55578213061?pr=1109#step:6:3186

andreaTP:
`OutOfMemory` detected here:
https://github.com/dylibso/chicory/actions/runs/19498412163/job/55806207897#step:6:5053

chirino:
Interesting.. I know test is trying to load a bunch of modules concurrently, so I guess it would use a bit of memory, but is it that much that would hit memory limits?

andreaTP:
No idea, but looks like something worth having a closer look 🙏

andreaTP:
Flaky test:
https://github.com/dylibso/chicory/actions/runs/20015371404/job/57391821151?pr=1129#step:6:3233

andreaTP:
https://github.com/dylibso/chicory/actions/runs/20130598086/job/57770849562#step:6:2902

andreaTP:
Still an OutOfMemory:

```
Error:  com.dylibso.chicory.experimental.dircache.DirectoryCacheTest.testConcurrentAccessNativeFS(Path) -- Time elapsed: 2.027 s <<< ERROR!
java.util.concurrent.CompletionException: java.lang.OutOfMemoryError: Java heap space
	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:315)
	at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:320)
	at java.base/java.util.concurrent.CompletableFuture$AsyncRun.run(CompletableFuture.java:1807)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.lang.Thread.run(Thread.java:840)
Caused by: java.lang.OutOfMemoryError: Java heap space
	at java.base/java.util.Arrays.copyOf(Arrays.java:3537)
[...]

PR #1103, opened 2025-11-07
fix: flaky dircache test #1101

chirino:
If an io error occurs trying to read a cache entry, return null instead of failing.


andreaTP:
Do you mind to rebase on latest main to fix CI?

PR #1106, opened 2025-11-10
Disable caching of Android devices

andreaTP:


andreaTP:
Caching doesn't seem to make much of a difference, this approach should be more stable.

PR #1110, opened 2025-11-17
Fix [dircache]: AccessDeniedExceptions on windows  #1101

chirino:
AccessDeniedExceptions seems to happen during concurrent access on windows on the write path.  Implemented a graceful retry logic to deal with race.


andreaTP:
Thanks for attempting the fix!
The CI failure seems compelling:
https://github.com/dylibso/chicory/actions/runs/19440736268/job/55623141549?pr=1110#step:6:5138

also, we should not insert timed retries at this lower level, it's surprising behavior and likely to fail anyway eventually.
If this is a "systemic" failure(e.g. an OpenJDK bug or similar) we cannot control we can retry the single test or disable it on specific(old and scarcely supported like win 11) platforms.

andreaTP:
I haven't seen this error popping up, closing for now.

PR #1132, opened 2025-12-08
Fix memory wait and notify

### LINKED ISSUE #1131: Broken lock instructions
When trying a wasm, built from Rust with atomics, that simply contends for a mutex to increment a counter, I see both deadlocks and lost increments.

I'm happy to share the simple Rust program, but I've also identified the problem to be the monitor logic in the Memory implementations. Will open a PR with test and fix shortly.

andreas-karlsson:
Fixes #1131 

The test in the first commit shows the previous implementation was broken. The second commit makes extensive changes to the monitor implementation, but I think they are warranted. It uses a single WaitState for all synchronisation (instead of the two maps). 

Note on Cleanup: This implementation does not garbage collect unused monitors to avoid churn on frequently reused addresses. Given typical usage patterns, the memory overhead is acceptable. If cleanup is deemed necessary, I think a WeakRef/GC based approach would be better rather than a direct one. Let me know if you want me to explore that.

### DIFF
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
@@ -100,9 +100,6 @@ public final class ByteArrayMemory implements Memory {
         WaitState state = waitStates.computeIfAbsent(address, k -> new WaitState());
 
         synchronized (state) {
-            assert (0 <= state.pendingWakeups);
-            assert (state.pendingWakeups <= state.waiterCount);
-
             // Check the condition while holding the lock
             // This must be atomic with the decision to wait
             if (!condition.getAsBoolean()) {
@@ -167,9 +164,6 @@ public final class ByteArrayMemory implements Memory {
         }
 
         synchronized (state) {
-            assert (0 <= state.pendingWakeups);
-            assert (state.pendingWakeups <= state.waiterCount);
-
             int actualWaiters = state.waiterCount - state.pendingWakeups;
 
             if (actualWaiters == 0) {
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/ByteBufferMemory.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/ByteBufferMemory.java
@@ -55,8 +55,12 @@ public final class ByteBufferMemory implements Memory {
     }
 
     // Tracks wait state per address: waiter count and pending wakeups
+    // all field access should be guarded by synchronizing on the instance.
+    // Invariants: 0 <= pendingWakeups <= waiterCount
     private static final class WaitState {
+        // The number of threads currently waiting
         int waiterCount;
+        // The number of waiting threads that have been scheduled to wake up
         int pendingWakeups;
     }
 
@@ -105,11 +109,20 @@ public final class ByteBufferMemory implements Memory {
                         throw new ChicoryInterruptedException("Thread interrupted");
                     }
                 }
-                // Consume one pending wakeup
-                state.pendingWakeups--;
                 return 0; // woken
             } finally {
+
+                if (state.pendingWakeups > 0) {
+                    // any thread leaving the try block is correct to consume
+                    // a pending wakeup IF available:
+                    // ret 0 - woken thread correctly consumes a wakeup
+                    // ret 2 - timeout can only occur with pendingWakeups == 0 so will not consume
+                    // throw - isn't part of the wasm runtime but is semantically correct to consume
+                    state.pendingWakeups--;
+                }
                 state.waiterCount--;
+                assert (0 <= state.pendingWakeups);
+                assert (state.pendingWakeups <= state.waiterCount);
             }
         }
     }
@@ -137,12 +150,13 @@ public final class ByteBufferMemory implements Memory {
         }
 
         synchronized (state) {
-            int actualWaiters = state.waiterCount;
+            int actualWaiters = state.waiterCount - state.pendingWakeups;
+
             if (actualWaiters == 0) {
                 return 0;
             }
 
-            // Calculate how many to wake: min(waiters, maxThreads)
+            // Calculate how many to wake: min(actualWaiters, maxThreads)
             int toWake;
             if (maxThreads < 0) {
                 toWake = actualWaiters; // wake all
@@ -152,10 +166,15 @@ public final class ByteBufferMemory implements Memory {
 
             // Add pending wakeups - waiters will consume these
             state.pendingWakeups += toWake;
+            assert (state.pendingWakeups <= state.waiterCount);
 
-            for (int i = 0; i < toWake; i++) {
-                state.notify();
-            }
+            // It's always safe to notify all. In the wait routine we consume pendingWakeups
+            // and go back to waiting if there are none. We could also choose to notify waiters
+            // one by one in a loop. The optimal choice of whether to notify all should be
+            // toWake > C * state.waiterCount, where C is an unknown constant weighing the cost
+            // of looping through notify vs. having threads wake and go back to waiting. Since
+            // the constant is unknown we opt for the simplest choice of just notifying all.
+            state.notifyAll();
 
             return toWake;
         }


runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
```diff
@@ -62,145 +62,117 @@ public ByteArrayMemory(MemoryLimits limits, MemAllocStrategy allocStrategy) {
...
-                }
+                toWake = Math.min(actualWaiters, maxThreads);
             }
+
+            // Add pending wakeups - waiters will consume these
+            state.pendingWakeups += toWake;
+
+            for (int i = 0; i < toWake; i++) {
+                state.notify();
+            }
```
andreas-karlsson: If toWake == actualWaiters this could be optimized as notifyAll().

runtime/src/main/java/com/dylibso/chicory/runtime/ByteBufferMemory.java
```diff
@@ -48,145 +48,117 @@ public ByteBufferMemory(MemoryLimits limits, MemAllocStrategy allocStrategy) {
...
-            notifyInProgress = null;
+            waitStates = null;
         }
     }
 
-    // atomic wait handling
-    private final Map<Integer, AtomicInteger> monitors;
-    private final Map<Integer, AtomicInteger> notifyInProgress;
+    // Tracks wait state per address: waiter count and pending wakeups
+    private static final class WaitState {
```
andreas-karlsson, line 60: All changes to this file are identical to the other memory. I'd be happy to fold this into an AbstractMemory?

runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
```diff
@@ -62,145 +62,117 @@ public ByteArrayMemory(MemoryLimits limits, MemAllocStrategy allocStrategy) {
...
+    private int waitOn(int address, BooleanSupplier condition, long timeout) {
         if (!shared()) {
             throw new ChicoryException("Attempt to wait on a non-shared memory, not supported.");
         }
-        AtomicInteger monitor = nextMonitor(address);
 
-        synchronized (monitor) {
+        long deadline = (timeout < 0) ? Long.MAX_VALUE : System.nanoTime() + timeout;
+
+        WaitState state = waitStates.computeIfAbsent(address, k -> new WaitState());
```
andreaTP, line 100: this line is suspicious, it will basically perform a `lock` even if it was not triggered before.
The question to answer is:
what should happen when there is a wait over an unlocked atomic?

Gut feeling is that it can `return 0` immediatly without going through the wait procedure at all, but it's worth looking back at the spec and other implementations.

runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
```diff
@@ -62,145 +62,117 @@ public ByteArrayMemory(MemoryLimits limits, MemAllocStrategy allocStrategy) {
...
+                        Thread.currentThread().interrupt();
+                        throw new ChicoryInterruptedException("Thread interrupted");
+                    }
                 }
+                // Consume one pending wakeup
+                state.pendingWakeups--;
+                return 0; // woken
             } finally {
-                endWaitOn(address);
+                state.waiterCount--;
```
andreaTP, line 137: I feel like this is the right place to remove the entry from `waitStates` when both `pendingWakeups` and `waiterCount` are equals to 0.
This will keep the map clean when everything is running as expected.

runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
```diff
@@ -62,145 +62,117 @@ public ByteArrayMemory(MemoryLimits limits, MemAllocStrategy allocStrategy) {
...
+                    int nanos = Math.max((int) (remaining % 1_000_000L), 0);
+                    try {
+                        state.wait(millis, nanos);
+                    } catch (InterruptedException ie) {
+                        Thread.currentThread().interrupt();
+                        throw new ChicoryInterruptedException("Thread interrupted");
+                    }
                 }
+                // Consume one pending wakeup
+                state.pendingWakeups--;
```
andreaTP: I'm a little nervous about this unguarded decrement(also the following one), we should check the behavior is correct even after a timeout happened for example.

What would be a condition where one of those variables goes below 0?

runtime/src/main/java/com/dylibso/chicory/runtime/ByteBufferMemory.java
andreaTP: Thanks for the proposal, I think we can resist a little longer the temptation..
You are right is a bit of copy-paste, but ~100 LOC is not a lot and we can refactor after things stabilize a bit more.

runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
```diff
@@ -62,145 +62,117 @@ public ByteArrayMemory(MemoryLimits limits, MemAllocStrategy allocStrategy) {
...
+        if (state == null) {
             return 0;
         }
 
-        synchronized (monitor) {
-            if (maxThreads < 0 || monitor.get() < maxThreads) {
-                notifyInProgress.put(address, new AtomicInteger(monitor.get()));
-                monitor.notifyAll();
+        synchronized (state) {
+            int actualWaiters = state.waiterCount;
```
andreas-karlsson: I think there's still a bug here. actualWaiters should be `state.waiterCount - state.pendingWakeups`. As it is now, repeated calls to notify can stack up so that later arriving waits will consume older pendingWakeups 🤔

andreaTP (commented):
@andreas-karlsson thanks a lot for this valuable contribution!

I added a few comments for discussion, out of curiosity, are you testing this behavior across runtimes? e.g. V8/wasmtime?

runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
andreaTP: Might be unrelated, but I'm a bit surprised by the fact that `nextMonitor` got removed ...

runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
andreas-karlsson: Not sure I get your question. I don't see how we ever could return 0 immediately. There might be an optional fast path to returning 1 if value != expected. Otherwise I think we need to create the waitState (maybe I should have called this monitor) and acquire it.

For clarity. In this impl, the map is only used to make sure all threads see the same waitState for an adress. The presence (or not) in the map isn't used for any decision, all state book keeping happens by synchronizing on the waitState.

runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
andreas-karlsson: I tried that, but the problem is that other threads coming into wait have already read the waitState and are waiting to acquire it. It can be fixed by adding a `reclaimed` to the waitState that is set when it's removed from the map. If another thread acquires a waitState that happens to be reclaimed it can loop around and read from scratch.

I'm happy to add this to the PR if you want a look.

runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
andreas-karlsson: It's guarded by the outer synchronized block on state. And if the implementation is correct they can't go below zero .

andreas-karlsson:
> ...are you testing this behavior across runtimes? e.g. V8/wasmtime?

Yes. Like I mentioned in the issue we discovered these problems while doing a POC for compiling Rust to multi-threaded wasm. The POC is a simple test where N threads increment a variable guarded by a mutex, so basically the same test I pushed in this PR, but sourced from Rust compilation. The POC runs fine in Node and Go/Wazero but fails in Chicory without these modifications. I can probably push the POC somewhere if you want.

I realise this PR is not ideal in completely changing the implementation pattern. I tried a couple of times fixing the problems with the original implementation, but ultimately found this approach easier.

runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
andreaTP: I was thinking at something like:

```java
if (!waitStates.containsKey(address)) { return 0; }
```

Meaning that if you are waiting on an address that hasn't been locked you should always return early.
Does it make sense now?

runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
andreaTP: > other threads coming into wait have already read the waitState and are waiting to acquire it.

If I understand it correctly, we should be able to perform this operation in the `synchronized` block, within it other threads should not be able to acquire the lock.
But maybe I'm overlooking something ...

andreaTP:
> The POC runs fine in Node and Go/Wazero but fails in Chicory without these modifications.

Thanks for confirming, and thanks for taking the challenge to fix it!

> I realise this PR is not ideal in completely changing the implementation pattern. I tried a couple of times fixing the problems with the original implementation, but ultimately found this approach easier.

No problem at all, it's easy to read :+1:

runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
andreaTP: > And if the implementation is correct

That's accurate, can I ask you to leave a couple of assertions around so that is easy to figure out where to look in case of issues(for example a slightly different behavior on certain platforms like windows).

runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
andreas-karlsson: Wow. I was actually so far oblivious to the fact that this lock mechanism is also used to implement all atomic operations. But either way, the spec says to always block until timeout or notify. How could we just return "woken"?

runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
andreas-karlsson: The problem is that there might be other threads blocked on the same synchronized block (or in notify) waiting to get in. If the last thing we do is silently remove it from the map, the next thread will enter the block but now operate on a dangling state. This can be overcome if we mark the state before leaving the block and then other threads will need to re-read from the map if they acquire a dangling state. This potential re-read needs to happen in both wait and notify.

runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
andreaTP: Thanks for the feedback, I went back to the source of truth and you are right:

https://webassembly.github.io/threads/core/exec/instructions.html#xref-exec-runtime-syntax-instr-admin-mathsf-wait-xref-exec-runtime-syntax-loc-mathit-loc-n

> The thread remains suspended.

I'm now convinced that "implicitly locking" is a legit implementation of the behavior, thanks for bearing with me!

runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
andreaTP: > there might be other threads blocked on the same synchronized block

Assuming everything we discussed so far is correct, I'd expect that when the 2 counters are at 0 there should be strictly no other thread blocked(e.g. everything else has been already notified or timed out).

I attempted to implement this behavior using atomic operations during the insert in the map (e.g. the mentioned `nextMonitor` function).

If with "re-read" you mean restoring a similar behavior I think it's the right thing to do.

andreas-karlsson:
Added a commit that fixes an invariant violation, adds some clarifying comments and more assertions.

Note. ByteBufferMemory is left out of sync until we're happy with the impl.

runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
andreas-karlsson: I tried making a deterministic clean-up that removes when waiterCount reaches zero. It leads to deadlocks. A thread dump showed that it's due to the lock() method, which makes sense. lock() hands out waitStates which might be cleaned-up and replaced leading to multiple waitStates for the same adress. I think the orignial impl. had the same problem. My `reclaimed` flag doesn't help here either since we have no control of the states handed out through lock(). The only way I can think to do cleanup with the lock() api is by holding weak references in the map.

andreas-karlsson:
@andreaTP I got a bit [carried away last night](https://github.com/dylibso/chicory/compare/main...andreas-karlsson:chicory:atomics-in-memory) and experimented with moving all atomic primitives into the Memory interface. The prime reason was that I wanted to try implementing atomics with var-handles. That worked and gave a 5-fold performance increase in our Rust locking stress test. It could also potentially remove lock() from the Memory interface and make that an implementation detail. Curious what you think about this!

andreaTP:
@andreas-karlsson I was about to look deeper into this, but I really like where you are going in your branch!
Making the `lock` an implementation detail sounds really appealing.
We should only be mindful about not rolling out an API breaking change and we are good to go :+1: 

How do you prefer to move on?
Do you want to bring this to an end or do you prefer to squash everything together?

FYI: I'm around till the end of this week, than I have extended vacation till next year(~3 weeks) with much less screen time

andreas-karlsson:
@andreaTP Nice! I'd much rather go in this direction as well. I think backwards compatibility can be ensured by just deprecating `lock` but still make all new Memory methods have default implementation based on it (like in the branch). The branch already passes all tests, but `ByteBufferMemory` would fail the new locking stress test in this PR. I was thinking of opening a new PR, but maybe it's better to just keep working in here. I'll push the Memory changes on top of this, then what remains is:
- Align on the correct names of all new methods
- Still fix the lock problems for the sake of `ByteBufferMemory`
- Figure out lock clean-up

Sounds like a plan?

andreas-karlsson:
@andreaTP Actually! I think it's better to separate these things. The PRs compose cleanly. Let's keep this one about fixing the current lock problems and I'll open a separate PR for the atomics in memory.

andreaTP:
As you prefer :slightly_smiling_face: , give me ~a day to carefully review this PR though, I want to try a couple of things to verify that I do understand all the intrisics.

runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
andreaTP: I looked at the code and reached the same conclusions as you.
I'm fine with this :+1: thanks for the explanation.

runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
```diff
@@ -62,145 +62,142 @@ public ByteArrayMemory(MemoryLimits limits, MemAllocStrategy allocStrategy) {
...
         }
-        AtomicInteger monitor = nextMonitor(address);
 
-        synchronized (monitor) {
+        long deadline = (timeout < 0) ? Long.MAX_VALUE : System.nanoTime() + timeout;
+
+        WaitState state = waitStates.computeIfAbsent(address, k -> new WaitState());
+
+        synchronized (state) {
+            assert (0 <= state.pendingWakeups);
```
andreaTP: nitpick: I'm good with the assertions in the finally block, feel free to remove the once at the start of a `synchronized` blocks

andreas-karlsson:
Fixed your last comments and ByteBufferMemory part is now identical.

andreaTP:
Released 1.6.1 containing those fixes, thanks again!

Issue #883, opened 2025-05-07
What is our promise for build time AoT class compatibility with different Runtime versions.


chirino:
End users may build time generate AoT with chicory 1.3, but then later try to run with chicory 1.4 or 1.5.  
What should our position be on supporting this?  Would the reverse scenario make sense (running with an older runtime)?

Should the runtime continue to offer backward binary compatibility with AoT-generated classes? 

Users would likely love to have that flexibility.  But it adds an additional maintenance burden on us, and would likely require use to setup some way to verify new runtime versions continue to be compatible with previously generated classes.

If we only want to support using a matching runtime, we could have the AOT classes validate the runtime version and bail if it's not the expected version.

electrum:
My understanding is that Chicory is supposed to be backwards compatible within the 1.x releases according to semantic versioning rules. AOT is no different than any other application that uses Chicory as a library.

The AOT methods are an internal implementation detail of the AOT code and are thus inlined. You can think of them as part of the generated code. The fact that we write them as normal Java code rather than generating them via code is simply a convenience for development.

Making a separate AOT runtime library would mean we need to consider binary compatibility. There’s only a handful of methods and I can’t imagine any security fixes that wouldn’t also affect the generated code.

andreaTP:
Thanks for the input @electrum !

> Making a separate AOT runtime library would mean we need to consider binary compatibility. 

Isn't the same as the compatibility WRT `runtime`? at the end of the day `AotMethods` is mostly "bridges" to it.

electrum:
I mean compatibility of the API for the AOT methods, which are static methods with a very specific signature that is dependent on what the generated bytecode needs.

andreaTP:
I'm not sure I'm following your reasoning @electrum 🙁 

Let me try to detail, current status:

generated bytecode -> AOT methods in bytecode -> `runtime` library

proposed:

generated bytecode -> `aot-runtime` library

from a compatibility perspective is the very same to me, we are just moving around what we already have.

> Making a separate AOT runtime library would mean we need to consider binary compatibility.

We need to do it, already, for `runtime`, I see no differences, there is just one less intermediate step.

chirino:
Seems like the short answer to the original question is: YES, using newer runtime versions is supported with an older build-time compiled wasm module. 

[...]

PR #1135, opened 2025-12-10
move atomic primitives to Memory

### LINKED ISSUE #1137: Memory grow should be atomic
The memory grow operation should be atomic on shared memories. Right now both `ByteArrayMemory` and `ByteBufferMemory` are broken in this regard.

andreas-karlsson:
This PR achieves the following (roughly corresponding to each commit):
- It moves all atomic memory operations to Memory interface.
- It adds a new Rust based multithreaded benchmark that calculates primes based on sieve technique that stresses atomic operations.
- It makes ByteArrayMemory use var-handles for all atomic operations, greatly increasing performance.
- It fixes #1137 for both memory implementations by switching to allocating memory in page-sized chunks.
- It deprecates MemAllocStrategy since it's not relevant when allocating by page.

All changes should be backwards compatible by leveraging default methods and deprecation in Memory.java. The changes are not forwards compatible however. AOT-code generated by an older version will not work with this newer runtime. This is considered acceptable since the old implementation had bugs making multithreaded builds practically unusable.

Note: The var-handle optimisation of ByteArrayMemory will currently fallback to the locking implementation on Java 25, due to dropped support for var-handle atomics. I'm not sure if this will be addressed, otherwise we'll need to make a new Memory implementation based on MemorySegment which is the future-proof API.

### DIFF
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
@@ -75,6 +75,9 @@ public final class ByteArrayMemory implements Memory {
     // Number of currently allocated pages
     private volatile int nPages;
 
+    // Lock for grow operation (only used when memory is shared)
+    private final Object growLock = new Object();
+
     public ByteArrayMemory(MemoryLimits limits) {
         this.limits = limits;
         int maxPages = min(limits.maximumPages(), RUNTIME_MAX_PAGES);
@@ -242,7 +245,16 @@ public final class ByteArrayMemory implements Memory {
     }
 
     @Override
-    public synchronized int grow(int size) {
+    public int grow(int size) {
+        if (!shared()) {
+            return growImpl(size);
+        }
+        synchronized (growLock) {
+            return growImpl(size);
+        }
+    }
+
+    private int growImpl(int size) {
         int prevPages = nPages;
         int numPages = prevPages + size;
 
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/ByteBufferMemory.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/ByteBufferMemory.java
@@ -46,6 +46,9 @@ public final class ByteBufferMemory implements Memory {
     // Number of currently allocated pages
     private volatile int nPages;
 
+    // Lock for grow operation (only used when memory is shared)
+    private final Object growLock = new Object();
+
     public ByteBufferMemory(MemoryLimits limits) {
         this.limits = limits;
         int maxPages = min(limits.maximumPages(), RUNTIME_MAX_PAGES);
@@ -213,7 +216,16 @@ public final class ByteBufferMemory implements Memory {
     }
 
     @Override
-    public synchronized int grow(int size) {
+    public int grow(int size) {
+        if (!shared()) {
+            return growImpl(size);
+        }
+        synchronized (growLock) {
+            return growImpl(size);
+        }
+    }
+
+    private int growImpl(int size) {
         int prevPages = nPages;
         int numPages = prevPages + size;
 


runtime/src/main/java/com/dylibso/chicory/runtime/Memory.java
```diff
@@ -40,6 +41,272 @@ static int bytes(int pages) {
 
     int notify(int address, int maxThreads);
 
+    default int atomicLoadInt(int addr) {
```
andreas-karlsson: @andreaTP I went for "load" cause I was looking at the spec, but in retrospect I think "read" is more consistent with this repo. Would you prefer `atomic(Read/Write)<type>`?

I also understand the `<type>` as using... ok I think I see the convention now, and I didn't follow it. All write methods can use Java type names and corresponding type i.e. `Byte`, `Short`, `Int` and `Long` it's unambiguous. All read methods should specify `<type>` as `<U|I><width>` and always return `long` with the correct sign extension.

Is this correct? I guess there is one offending example in `writeI32(...)`...

andreas-karlsson: No.. then I discovered the `readInt` etc. I guess these are convenience and return the signed type. So the question is.. should we have only the convenient methods in this interface and sign extension can be covered deeper in the stack?

andreas-karlsson: I went with `atomic(Read|Write)(Byte|Short|Int|Long)`

runtime/src/main/java/com/dylibso/chicory/runtime/InterpreterMachine.java
```diff
@@ -2286,56 +2272,84 @@ private static void I64_ATOMIC_STORE(MStack stack, Instance instance, Operands o
...
+                throw new IllegalStateException("Unexpected atomic op: " + op);
+        }
+        stack.push(oldVal);
     }
 
     private static void I32_ATOMIC_RMW_CMPXCHG(MStack stack, Instance instance, Operands operands) {
-        var replacement = (int) stack.pop(); // c3
-        var expected = (int) stack.pop(); // c2
-        var ptr = readMemPtr(stack, operands); // i
+        var replacement = (int) stack.pop();
```
andreaTP: nit: I think the names in the comments are reflecting the spec, I'd leave it for the future reader.

runtime/src/main/java/com/dylibso/chicory/runtime/Memory.java
andreaTP: Those method names suffered a bit of "organic grow", the best we can do is to keep some level of readability.
I'm good with the proposed naming :+1:

compiler/src/main/java/com/dylibso/chicory/compiler/internal/Shaded.java
```diff
@@ -169,110 +169,86 @@ public static void memoryWriteDouble(int base, double value, int offset, Memory
 
     public static int memoryAtomicIntByteRead(int base, int offset, Memory memory) {
         var ptr = getAddr(base, offset);
-        synchronized (memory.lock(ptr)) {
-            return memory.read(ptr);
-        }
+        return Byte.toUnsignedInt(memory.atomicReadByte(ptr));
```
andreaTP, line 182: Let me reason out loudly:
- assume we release those changes in a 1.7.0 release
- a user compiles using the build time compiler with 1.7.0
- attempting to use the generated classes with 1.6.0 of the runtime is going to break

correct?

andreaTP: We discussed in #883 and we are aiming for forward and not backward compatibility of generated code.

andreaTP (commented):
@andreas-karlsson The changes look good, it’s a nice cleanup, and I agree that `lock` is a bit of a leaky abstraction.

That said, without the “atomic operations based on VarHandle” portion (which I think is a great idea!), the overall value of this PR feels reduced.

Would you be up to opening another PR in parallel to show the direction we’re aiming for? Even with CI broken, it would give us a sense of how far we are from the desired goal.

I honestly assumed the VarHandle API was more stable, but given the incompatibilities you mentioned, I’m now wondering how well this will work with `native-image`. I think we’re missing a CI step to verify that compatibility.

andreas-karlsson:
@andreaTP I did more investigation into the Java 25 issue. It turns out atomic VarHandle operations still work on ByteBuffers but not on regular byte arrays. [This test](https://gist.github.com/andreas-karlsson/e6dcc3dc54e893aa5b64c88b15618020) confirms it.

Since `isAccessModeSupported` returns the correct status, I can gate the implementation granularly. I can add commits enabling atomic optimizations for both memory implementations; `ByteArrayMemory` will simply fall back to locks on Java 25.

It would be great to pair this with a performance test running on all platforms/versions in CI. I've noticed JMH in the repo and will look into adding a benchmark.

That said, I still believe the refactoring warrants inclusion independent of the optimization. I hadn't thought about the forward compatibility issue though! Will need to wrap my head around that..

andreaTP:
> I did more investigation into the Java 25 issue

I'll follow up asking around to relevant people :+1: 

> I can add commits enabling atomic optimizations for both memory implementations

Thanks, look like a reasonable path forward

> JMH in the repo and will look into adding a benchmark.

I was going to propose it, sounds great!

dmlloyd:
The issue with Java 25 most likely stems from this issue (and its fix): https://bugs.openjdk.org/browse/JDK-8318966

Basically, you can't (generally speaking) perform multi-byte atomic operations on `byte[]` because its minimum element alignment is 1 byte, and most CPUs require a minimum alignment that is equal to the size of the atomic type.

The best fix is to detect Java 22 or later, and use `MemorySegment` (or `VarHandle` derived from `MemorySegment`) wrapping a `long[]` instead of wrapping a `byte[]`.

andreas-karlsson:
@dmlloyd Thanks! That's most likely it. I'll take a proper look at implementation switching to `MemorySegment` when available.

@andreaTP I took the liberty to squash and rebase the branch. Then fired off two hopeful commits in quick succession, but both turned very red. I'll put this PR back in draft and workout something more solid.

andreas-karlsson:
I only tried the `ThreadsProposalTest` before pushing, and that worked on both Java 21/25. But it turns out in other cases we do a bunch of operations on ByteBuffers that don't work with direct buffers, and I guess there was a reason `ByteBufferMemory` didn't use direct buffers. I'll focus the solution on getting `ByteArrayMemory` to work first.

andreas-karlsson:
@andreaTP I've just pushed a new benchmark based on parallel sieve primes computation. It uses a lot of atomics and also some locking. I also pushed an optimized ByteArrayMemory, but it's basically same as before. Using MemorySegment as suggested isn't a clean fit in ByteArrayMemory, I think it would rather be a separate Memory impl.
```
      (memoryType)  (numWorkers)   Mode  Cnt   Score   Error  Units
   ByteArrayMemory             4  thrpt    5  86,899 ± 2,367  ops/s
  ByteBufferMemory             4  thrpt    5  24,947 ± 1,094  ops/s
```
These are the scores I'm getting when ByteArrayMemory is optimized and ByteBufferMemory is not. I guess it'd be better to compare just ByteArrayMemory optimized vs unoptimized. But to do that in CI I guess it would need separate PRs?

Note: The reason it works on Java 25 now is cause it falls back to locking.

compiler/src/main/java/com/dylibso/chicory/compiler/internal/Shaded.java
andreas-karlsson: Yes, calling new runtime methods in `Shaded` will break on an older runtime... what's the rational behind this requirement? Setups where one dynamically pushes newly compiled classes into an otherwise static installation? I don't see any way around it. What does that mean? This PR would require a new major?

andreas-karlsson:
This is still very flaky 😞  will need to do more testing in CI

andreaTP:
@andreas-karlsson hope this helps, to fight flaky tests I usually start with something like this: https://github.com/andreaTP/chicory/commit/33ef658c8d4bb04afe21a1aa19cff9f7a61c223f and run the checks on my fork.

Numbers are great!

andreas-karlsson:
@andreaTP Ah! Smart!

andreas-karlsson:
@andreaTP It [looks](https://github.com/andreas-karlsson/chicory/actions/runs/20167560191) like it was the unguarded grow #1137 that caused the flakiness*. On that branch I'm not using any var-handle atomics but guard `grow` behind a rw-lock, and I'm confident it would also work with var-handle atomics.

Fixing the `grow` bug is easy but can't be done in the "old" Memory interface, it would spill the grow lock into the compiler and interpreter. So that's my biggest question right now, what are the prospects of the Memory interface refactor? Will it need a breaking release? If so, when can that be done? I know you're about to go on vacation, but it would be nice to have some guidance on this cause I will have plenty of free time coming weeks 🙂  

*Edit: The 2/100 failures are unrelated.

andreaTP:
@andreas-karlsson thanks a lot for the help here! Much appreciated!

> So that's my biggest question right now, what are the prospects of the Memory interface refactor? Will it need a breaking release? If so, when can that be done?

If we manage to have the right deprecation in place and we verify forward compatibility we should be able to release a `1.7` with the changes :+1: not a lot of modules are really leveraging the threads proposal at the moment and the risk is low.

andreas-karlsson:
> verify forward compatibility 

@andreaTP Does this mean creating like a memory-adapter in the compiler? That could be done, but i think the bugs in the current impl. makes it unusable for anything non trivial, so I don't know if it's worth the effort to salvage forward compatibility in this case?

andreaTP:
Fair, feel free to move on :+1:

andreaTP:
@andreas-karlsson hi and happy new year!
I'm around now and happy to pick up again those improvements, let me know when is a good time for review :pray:

andreas-karlsson:
@andreaTP Happy new year to you as well!

I've just reacquainted myself with the PR and the state is as follows:
- Trying to use a read-lock to guard memory grow was disastrous(!) for performance, so I switched to allocating pages in separate chunks, which I think makes a lot of sense.
- I think the memory benchmark is a good addition, using a "real" calculation compiled from Rust, pushing both atomics and locks.
- The current ByteArrayMemory impl. works really well, except for Java 25 where it falls back to locks, which leads to performance degradation but is still correct.
- The PR doesn't address ByteBufferMemory at all.

I got a bit bogged down before holidays in how to reason about the two memory implementations, and how to introduce a third one(?!) based on the new MemorySegment. There are many options. Should we have a separate implementation of ByteArrayMemory that gets selected by [Multi Release Jar](https://www.baeldung.com/java-multi-release-jar)? But why would a MemorySegment impl. masquerade as a byte-array impl.? And what's actually the rationale for having even two memory implementations? An alternative would be to instead deprecate ByteArrayMemory and focus on the ByteBufferMemory, which (if using direct buffers) works well with var-handle atomics across all Java versions. We could just make direct buffers an option and fall-back to locks for atomics otherwise. This would consolidate things and lessen the maintenance burden.

Another option would be to scale this PR back to just include the performance benchmark and the changes to the Memory interface and its usage in compiler/interpreter. I slightly favour this because it'd be nice to have the benchmark giving comparable numbers when working on changes to the memory implementations.

Sorry for the long post, but I need some guidance 😅

andreaTP:
Thanks a lot for keeping the engagement @andreas-karlsson !


> Trying to use a read-lock to guard memory grow was disastrous(!) for performance, so I switched to allocating pages in separate chunks, which I think makes a lot of sense.

Looks fair!

> I think the memory benchmark is a good addition, using a "real" calculation compiled from Rust, pushing both atomics and locks.

Agree :+1: thanks!

> The current ByteArrayMemory impl. works really well, except for Java 25 where it falls back to locks, which leads to performance degradation but is still correct.

I think is acceptable, IIRC there is an open Issue for Java 25, so the problem will eventually get solved.

> The PR doesn't address ByteBufferMemory at all.

Currently, we are using `ByteBufferMemory` as a safe default as it works on Android too.
I think the changes can be included in a separate PR if we feel the need for it.

> I got a bit bogged down before holidays in how to reason about the two memory implementations, and how to introduce a third one(?!) based on the new MemorySegment. There are many options.

That's correct, at the moment the situation is:
- `ByteArrayMemory` JVM only, high perf
- `ByteBufferMemory` default, safe

I'd be interested in looking at performance numbers to see if another implementation(in the `runtime` module, viable options are, for example: spitting out a separate sub-module, publishing from another repo etc.) is worth it, maybe we can consider changing the internals of `ByteArrayMemory`?

> Should we have a separate implementation of ByteArrayMemory that gets selected by [Multi Release Jar](https://www.baeldung.com/java-multi-release-jar)?

In principle, I'd refrain from adding additional complexity if not well justified.

> An alternative would be to instead deprecate ByteArrayMemory and focus on the ByteBufferMemory

I'm afraid this is not a desirable option, the gut feeling is that we will fall into spending time chasing Android vs. JVM differences and support instead of focusing on more meaningful targets like perf on the JVM.

> Another option would be to scale this PR back to just include the performance benchmark and the changes to the Memory interface and its usage in compiler/interpreter. I slightly favour this because it'd be nice to have the benchmark giving comparable numbers when working on changes to the memory implementations.

This is all good! :slightly_smiling_face: whatever flow works the best for you!

> Sorry for the long post, but I need some guidance 😅

Very welcome, you are doing great job in this project and I'm grateful for the time you are spending.

andreas-karlsson:
@andreaTP Thanks for the clear guidance and sorry for the delay, I was held up with other things.

I've decided to push through this PR basically as is. I added a commit to make the memory interface more consistent: wait and notify now align with the underlying WASM by having the "atomic" prefix, and I moved `atomicFence` to the `Memory` interface for completeness.

I also considered ensuring the compiler and interpreter use the same set of methods from `Memory`. Right now the compiler emits its own conversions to unsigned, whereas the interpreter uses the `Memory.readUXX` helpers. However, keeping these changes backwards compatible becomes a mess of default implementations, so I left it. Maybe we could take the opportunity to make the `Memory` interface more consistent in the next major release? (I'm a perfectionist 😄)

I think this branch is in a good state and I'm ready to open it for review. The only thing left is `ByteBufferMemory` (implementing atomic grow via separate pages, and adding the `atomicXX` methods to simplify removal of default implementations in `Memory`). I can do that in this PR or a follow-up—whichever you prefer!

runtime/src/main/java/com/dylibso/chicory/runtime/OpcodeImpl.java
```diff
@@ -7,20 +7,28 @@
 import com.dylibso.chicory.wasm.types.OpCode;
 import com.dylibso.chicory.wasm.types.PassiveElement;
 import com.dylibso.chicory.wasm.types.ValType;
-import java.lang.reflect.InvocationTargetException;
 
 /**
- * Note: Some opcodes are easy or trivial to implement as compiler intrinsics (local.get, i32.add, etc).
```
andreas-karlsson, line 13: Oops, will remove this unintentional formatting change.

test-gen-lib/src/main/java/com/dylibso/chicory/testgen/TestGen.java
```diff
@@ -59,11 +59,11 @@ public static void execute(
                         excludedUnlinkableWasts);
 
         // Create destination folders
-        if (!compiledWastTargetFolder.mkdirs()) {
+        if (!compiledWastTargetFolder.isDirectory() && !compiledWastTargetFolder.mkdirs()) {
```
andreas-karlsson, line 62: This was just annoying as it makes it impossible to run `mvn install` without clean.

andreaTP: np :+1:

andreaTP:
Sorry for the delay @andreas-karlsson , I was waiting for the PR to get out of Draft :slightly_smiling_face: .

I skimmed through and I think I'm overall good with the changes, when you want to finalize, please, mark the PR as ready for review and we go through the final steps for merging!

andreas-karlsson:
@andreaTP I needed a bit of weekend time to get this through.. and I thought I was finished. I implemented page-chunked memory also in BytaBufferMemory, but it fails on Java 11/17. I pulled my hair for 5 hours trying to figure out why. Turns out it was #843 or at least it works when I run the test with `-XX:CompileCommand=dontinline,com/dylibso/chicory/runtime/OpcodeImpl.I32_GE_U` So not sure what to do now.. maybe just back out of the ByteBufferMemory changes and accept that it's broken? Or is there a fix? I'm too tired to see this straight right now 😅

andreaTP:
Hi @andreas-karlsson and thanks a lot for your effort and dedication!

If, running the tests with `-XX:CompileCommand=dontinline,com/dylibso/chicory/runtime/OpcodeImpl.I32_GE_U` fixes the issue, I think you are really hitting the JVM bug and we should put in place a workaround (I can take care of it).
Let me have a closer look in the next days :+1:

andreas-karlsson:
Thanks @andreaTP I opened the PR for review.

runtime/src/main/java/com/dylibso/chicory/runtime/alloc/ExactMemAllocStrategy.java
```diff
@@ -1,5 +1,12 @@
 package com.dylibso.chicory.runtime.alloc;
 
+/**
+ * Memory allocation strategy that allocates exactly the requested size.
+ *
+ * @deprecated Memory is now allocated by page (64KB each), so custom allocation
```
andreaTP, line 6: I'm sorry, I haven't realized before that page allocation is incompatible with allocation strategies.
A few questions:
- what's the expected performance increase?
- do we think that users would want to tweak this?
- can paging be made compatible with allocation strategies?

runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
```diff
@@ -212,20 +242,20 @@ public int pages() {
     }
 
     @Override
-    public int grow(int size) {
-        var prevPages = nPages;
-        var numPages = prevPages + size;
+    public synchronized int grow(int size) {
```
andreaTP: I see why this is necessary, at the same time we need this synchronization only when the memory is shared.
What do you think about making possible to disable the `synchronized` block when `shared == false`?

runtime/src/test/java/com/dylibso/chicory/runtime/MemoryTest.java
```diff
@@ -0,0 +1,77 @@
...
+                        (Supplier<Memory>)
+                                () -> new ByteArrayMemory(new MemoryLimits(1, 1000, true))),
+                Arguments.of(
+                        "ByteBufferMemory",
+                        (Supplier<Memory>)
+                                () -> new ByteBufferMemory(new MemoryLimits(1, 1000, true))));
+    }
+
+    /**
+     * Test that concurrent grow and write operations don't lose writes.
```
andreaTP, line 27: Nice, thanks! 👍

wasm-corpus/src/main/resources/compile.sh
```diff
@@ -17,6 +17,15 @@ compileRust() {
   (set -x; cd $1 && cargo build --target=$target &&  cp ./target/${target}/debug/*.wasm "../../compiled/$filename.rs.wasm")
 }
 
+compileRustAtomics() {
+  filename=$(basename "$1")
+  target="wasm32-unknown-unknown"
+  (set -x; cd $1 && \
+    RUSTFLAGS='-C target-feature=+atomics,+bulk-memory -C panic=abort -C link-arg=--export=__stack_pointer' \
+    cargo +nightly-2025-09-01 build --target=$target -Z build-std=core,alloc,std,compiler_builtins,panic_abort --release && \
```
andreaTP, line 25: I'm a bit surprised we need a custom configuration to compile the Rust program.

We might need this infrastructure eventually, but, at the moment, adding the `target-feature=+atomics,+bulk-memory` is probably enough even for the other Rust payloads.
Or we really need a nightly build?

runtime/src/main/java/com/dylibso/chicory/runtime/alloc/ExactMemAllocStrategy.java
andreas-karlsson: The `MemAllocStrategy` is actually equally compatible when allocating by page. I considered keeping it, but then I thought it wasn't worth it. In the previous strategy if you wanted to grow the mem by one page. You still needed to allocate and copy _all_ existing memory. So there was an upside to amortizing the cost by using an over-allocating strategy. With the new allocation the cost is proportional to how many pages you need to grow, and there's really no point in amortizing that, so I thought we'd favour simplicity.

runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
andreas-karlsson: Happy to add conditional locking if you think it's important, but I'd lean toward keeping it simple since grow() is infrequent and uncontended locks are very cheap. Thoughts?

wasm-corpus/src/main/resources/compile.sh
andreas-karlsson: It actually might be possible if I'm using the wasi target. Will take a look!

andreas-karlsson: No. Right. Chicory doesn't implement the wasi threads extension. Also I think it's pretty cool to show an example how threads can be used bare-bones. Either way we're stuck on nightly since we need to rebuild std with atomics.

As a side note. It'd be fun and not too difficult to implement wasi-threads in Chicory! It'd lower the barrier to use multi threading. Although the threads extension isn't strictly standard and wasi-p1 in general is already obsolete..

andreas-karlsson: The verdict is still out on this one. It might be possible to do on stable. It's a jungle of dated information out there, but I've now found a link-arg that might fix it. The built wasm at least has the right exports. Will try tomorrow!

andreas-karlsson: Sorry for the back and forth. But no, the wasm isn't correct. `__wasm_init_tls` is empty which I think is because std needs to be rebuilt with atomics. 

I'll take a look at implementing wasi threads though, and then we can port this build to wasi instead.

runtime/src/main/java/com/dylibso/chicory/runtime/alloc/ExactMemAllocStrategy.java
andreaTP: The only downside is that we are effectively dropping a pre-existing feature in this way.
I like the argument of favoring simplicity though, let's agree to move forward with your proposal, if someone complains we re-iterate.

runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
andreaTP: If I reviewed correctly this is the only place where we'd need a conditional lock.
The expectation is that, if a program doesn't uses a shared memory, it should not trigger any lock.

I think, in this case, the toll on readability is minimal. wdyt?

wasm-corpus/src/main/resources/compile.sh
andreaTP: Let's not increase the scope, this can be solved or cleaned up in a follow up PR :+1:

andreaTP:
@andreas-karlsson last open comment and rebase on latest `main` to get the CI happy and we are ready to merge.
Great job!

andreas-karlsson:
@andreaTP Done! Thanks for shepherding the PR!

Note: First I thought the failing test was my flaky fence test, so I went a head and opened #1180 to fix that, but it turned out to be something else. I assume it's unrelated to this PR, do you know?

andreaTP (approved):
Lgtm, thanks for all the amazing work done here @andreas-karlsson !

andreaTP:
> I assume it's unrelated to this PR, do you know?

It's completely unrelated, we are hitting a Jvm bug on Java <= 17 ( [ref](https://bugs.openjdk.org/browse/JDK-8376400) ).

PR #1178, opened 2026-01-26
Extend the usage of the Jvm workaround to I32_GE_U

andreaTP:
`I32_GE_U` is an opcode that frequently has been showing up during investigations.

Here we extend the memory workaround to create the megamorphic call on older Java versions.

I tested this patch on top of #1135 and it looks to be solving the issue.

As a follow up we should really report the bug upstream in OpenJDK, I'm afraid it might eventually sprawl if we don't do it.

cc. @andreas-karlsson


andreaTP:
@andreas-karlsson thanks for the review!

The goal here is to squeeze all the possible performance when running on a non-buggy JVM.
I just checked and `Long.compareUnsigned` should compile to an intrinsic too, but it will take twice the space on the Java stack(args and result).

Let's keep open the option when we can demonstrate that the difference is minimal though.

PR #1180, opened 2026-01-28
fix ThreadsProposalTest#atomicFenceOrder test flakiness

andreas-karlsson:
## Summary

Fix flaky `atomicFenceOrder` test by replacing timing-based termination with iteration-based approach.

## Changes

- `threads-example.wat`: `fenced_read_and_verify` now returns the counter value
- `ThreadsProposalTest`: Reader loops until counter reaches 10000 instead of relying on a 200ms timer
- Added `@Timeout(30)` as safety net to prevent infinite hangs if test is broken

The rationale is that previously the test verified that at least 10000 iterations had been achieved in the 200ms which was a bit optimistic for CI machines. The new scheme ensures both threads compute concurrently and that it runs until sufficient iterations are made. Still ~0.2s on my machine.


andreaTP:
LTGM, this is making the test more robust anyway, thanks!

PR #1186, opened 2026-02-02
[blog] Syntax Highlight in Java

andreaTP:
I was so much pleased hacking around this one that I decided to put together a blog post to share the smooth experience.


andreaTP:
Thanks @maxandersen for the nerd snipe :+1:

nilslice:
love it!

PR #1288, opened 2026-05-05
Compiler: auto-size dispatch chunks to stay under JIT HugeMethodLimit

zklapow:
The JVM's C2 JIT compiler has a `HugeMethodLimit` (default 8KB bytecode).

At the previous chunk size of 1024 entries, the generated dispatch methods exceed 8KB and never get full C2 optimization. This only affects large wasm modules (>1024 functions), but many real-world modules like CPython, SQLite, etc. easily hit this.

Instead of a fixed chunk size, this PR derives it from the bytecode budget: `highestOneBit(hugeMethodLimit / bytesPerEntry)`, which produces dispatch methods that stay under the limit and get full C2 compilation. The limit is configurable via `-Dchicory.hugeMethodLimit=<bytes>` for JVMs with non-default settings for the hugemethod limit.

**Benchmark results** (2000-function module, JMH, JDK 25):

| targetFunc | (1024/chunk) | (128/chunk, auto-sized) | speedup |
|---|---|---|---|
| 0 | 0.033 us/op | 0.004 us/op | ~8x |
| 999 | 0.034 us/op | 0.004 us/op | ~8.5x |
| 1999 | 0.035 us/op | 0.004 us/op | ~8.75x |

In practice in cpython I found this to give about a 15-20% speedup on hot loops.

### DIFF
--- a/compiler/src/main/java/com/dylibso/chicory/compiler/internal/Compiler.java
+++ b/compiler/src/main/java/com/dylibso/chicory/compiler/internal/Compiler.java
@@ -100,7 +100,17 @@ public final class Compiler {
     private static final MethodType MACHINE_CALL_METHOD_TYPE =
             methodType(long[].class, Instance.class, Memory.class, int.class, long[].class);
 
-    private static final int MAX_MACHINE_CALL_METHODS = 1024; // must be power of two
+    // C2 JIT's HugeMethodLimit (default 8KB) — methods exceeding this get degraded optimization.
+    // Dispatch chunks are sized to stay under this limit for full C2 compilation.
+    private static final int HUGE_METHOD_LIMIT =
+            Integer.getInteger("chicory.hugeMethodLimit", 8000);
+    // Estimated upper bound: tableswitch offset (4) + label + invokestatic (~5) + areturn (1) +
+    // overhead
+    private static final int ESTIMATED_BYTES_PER_DISPATCH_ENTRY = 40;
+    private static final int MAX_DISPATCH_METHODS =
+            Integer.highestOneBit(
+                    HUGE_METHOD_LIMIT / ESTIMATED_BYTES_PER_DISPATCH_ENTRY); // must be power of two
+    private static final int MAX_CALL_INDIRECT_METHODS = 1024; // must be power of two
     // 1024*12 was empirically determined to work for the 50K small wasm functions.
     // So lets start there and halve it until we find a size that works.
     // This should give us the biggest class size possible.
@@ -781,11 +791,11 @@ public final class Compiler {
 
         // static implementation for Machine.call()
         Consumer<InstructionAdapter> callMethod;
-        if (functionTypes.size() < MAX_MACHINE_CALL_METHODS) {
+        if (functionTypes.size() < MAX_DISPATCH_METHODS) {
             callMethod = asm -> compileMachineCallInvoke(asm, 0, functionTypes.size());
         } else {
             // Best value that worked with the 50K small wasm functions
-            var maxMachineCallMethods = MAX_MACHINE_CALL_METHODS << 2;
+            var maxMachineCallMethods = MAX_DISPATCH_METHODS << 2;
             maxMachineCallMethods =
                     loadChunkedClass(
                             functionTypes.size(),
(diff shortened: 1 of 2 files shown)


zklapow:
I originally did this by simply having `MAX_MACHINE_CALL_METHODS` be configureable, but that seemed worse than simply deriving from the limit and I imagine most workloads will have left the default 8kb so this produces more optimal results in the base case.

zklapow:
Will take a look at CI failures

compiler/src/main/java/com/dylibso/chicory/compiler/internal/Compiler.java
```diff
@@ -100,7 +100,15 @@ public final class Compiler {
     private static final MethodType MACHINE_CALL_METHOD_TYPE =
             methodType(long[].class, Instance.class, Memory.class, int.class, long[].class);
 
-    private static final int MAX_MACHINE_CALL_METHODS = 1024; // must be power of two
+    // C2 JIT's HugeMethodLimit (default 8KB) — methods exceeding this get degraded optimization.
+    // Dispatch chunks are sized to stay under this limit for full C2 compilation.
+    private static final int HUGE_METHOD_LIMIT =
+            Integer.getInteger("chicory.hugeMethodLimit", 8000);
+    private static final int BYTES_PER_DISPATCH_ENTRY = 40;
```
andreaTP: Is this number estimated?

compiler/src/main/java/com/dylibso/chicory/compiler/internal/Compiler.java
```diff
@@ -100,7 +100,15 @@ public final class Compiler {
...
+    private static final int MAX_DISPATCH_METHODS =
+            Integer.highestOneBit(
+                    HUGE_METHOD_LIMIT / BYTES_PER_DISPATCH_ENTRY); // must be power of two
+    private static final int MAX_CALL_INDIRECT_METHODS = 1024; // must be power of two
```
andreaTP, line 113: Should we apply the same mechanism to indirect calls too?

jmh/src/main/java/com/dylibso/chicory/bench/BenchmarkDispatchChunkSize.java
```diff
@@ -0,0 +1,64 @@
...
+import org.openjdk.jmh.infra.Blackhole;
+
+// Compare: -Dchicory.hugeMethodLimit=1000000 (effectively no limit) vs default (8000)
+@State(Scope.Benchmark)
+@Warmup(iterations = 3, time = 2)
+@Measurement(iterations = 5, time = 3)
+@OutputTimeUnit(TimeUnit.MICROSECONDS)
+@BenchmarkMode(Mode.AverageTime)
+@Fork(1)
+public class BenchmarkDispatchChunkSize {
```
andreaTP, line 45: Do you have numbers?

andreaTP (changes requested):
Great idea! Thanks, a couple of minor comments/questions.

compiler/src/main/java/com/dylibso/chicory/compiler/internal/Compiler.java
zklapow: ya, estimated I can add a comment or rename to ESTIMATED

compiler/src/main/java/com/dylibso/chicory/compiler/internal/Compiler.java
zklapow: probably... indirect is more complex b/c the sizes are not so easily estimateable, ill take quick pass at it tho.

jmh/src/main/java/com/dylibso/chicory/bench/BenchmarkDispatchChunkSize.java
zklapow: Theres some specific results in the description, just from a local run on my macbook tho not suppper rigorous.

compiler/src/main/java/com/dylibso/chicory/compiler/internal/Compiler.java
zklapow: updated to clarify

jmh/src/main/java/com/dylibso/chicory/bench/BenchmarkDispatchChunkSize.java
zklapow: added a comment here with my current results.

andreaTP (approved):
LGTM, thanks!

PR #1315, opened 2026-07-18
Runtime: fix lost wakeup from late-arriving atomicWaitBroken wait notify

### LINKED ISSUE #1314: atomicWait can steal a wakeup meant for an already-parked thread
`atomicNotify` should only be able to wake threads that are already parked in `atomicWait`. Because of how the pending-wakeup permit is tracked, a thread that calls atomicWait after a notify has already fired can consume that permit and return immediately — without ever actually waiting — even though it wasn't parked when the notify happened. This steals the wakeup from the thread that was legitimately parked, which then has no way to know it needs to re-wait and can sleep forever.

Discovered via deadlocks in an otherwise-correct rw-lock implementation built on top of atomicWait/atomicNotify.

andreas-karlsson:
Fixes #1314 

A pending wakeup left behind by atomicNotify could be stolen by an atomicWait call that arrives after the notify, letting it return without ever waiting — while the thread the wakeup was actually meant for gets no signal and can sleep forever (lost wakeup / deadlock).

Fixes ByteArrayMemory/ByteBufferMemory by tagging each WaitState with a generation counter, bumped on every notify. A waiter may only consume a pending wakeup minted since it registered, so late arrivals wait for the next notify instead of stealing the current one.

Split into two commits for review: a regression test that reproduces the steal (fails on main), then the fix. Also adds a stress test that drives a rw-lock built on atomicWait/atomicNotify under contention, to catch this class of bug more broadly.

### DIFF
--- a/runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
+++ b/runtime/src/main/java/com/dylibso/chicory/runtime/ByteArrayMemory.java
@@ -114,6 +114,14 @@ public final class ByteArrayMemory implements Memory {
         int waiterCount;
         // The number of waiting threads that have been scheduled to wake up
         int pendingWakeups;
+        // Bumped by every notify that issues wakeups. A waiter may only consume
+        // a pending wakeup minted while it was waiting (generation moved since
+        // it registered). Without this, a newly-arrived waiter could steal a
+        // wakeup meant for an already-parked thread. That thread would then
+        // re-check, find none, and sleep forever (lost wakeup).
+        // Only ever compared with ==/!=, never ordered, so wraparound is not
+        // a practical concern.
+        long generation;
     }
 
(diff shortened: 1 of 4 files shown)


andreas-karlsson:
Hello @andreaTP , I've seen the exciting new endive repo, but we haven't migrated yet, so I thought I'd contribute here first. Or maybe you want it the other way around?

andreaTP:
Hi @andreas-karlsson I do not have rights to merge PRs here.

andreas-karlsson:
@andreaTP ah, I see. I've already opened issue and PR in endive :) So what will happen to this repo? Frozen?

nilslice:
@andreas-karlsson we are happy to merge PRs here as well, I was only waiting for CI to pass before engaging.

andreas-karlsson:
@nilslice I don't think the failing CI is related to my changes?

